PR #1, opened 2020-01-07
ENH add finance dataset in Lasso benchmark

mathurinm:
- @tomMoral  should we call the dataset finance, or log1p_train as per LIBSVM terminology?

- the whole part with paths is WIP. Where do you think we should store the downloaded zipped, svmlight, and npz/npy file ? The whole processing can take up to 10 minutes from my experience, so I chose to store all 3 in celer (https://github.com/mathurinm/celer/blob/master/celer/datasets/libsvm.py)

- it is very easy to modify this code in order to download other libsvm datasets. Should I do a `download_libsvm.py` helper file somewhere else (where?), and the instantiate `finance.py`, `news20.py`  ?


tomMoral:
>     * @tomMoral  should we call the dataset finance, or log1p_train as per LIBSVM terminology?

I have no opinion I guess. I would say Finance is probably nicer but this is just esthetic. @agramfort  do you have an opinion here?

>     * the whole part with paths is WIP. Where do you think we should store the downloaded zipped, svmlight, and npz/npy file ? The whole processing can take up to 10 minutes from my experience, so I chose to store all 3 in celer ([mathurinm/celer:celer/datasets/libsvm.py@`master`](https://github.com/mathurinm/celer/blob/master/celer/datasets/libsvm.py)

This is a good idea to cache as much as possible. For the datasets, you should put them in `os.path.join(DATA_DIR, 'finance')` with `DATA_DIR = get_global_setting('cache_dir')` and adding in `benchopt/config.py` a default value in `DEFAULT_GLOBAL` which will be `./data`. Do you think this make sense?

>     * it is very easy to modify this code in order to download other libsvm datasets. Should I do a `download_libsvm.py` helper file somewhere else (where?), and the instantiate `finance.py`, `news20.py`  ?


Yes, this is indeed a good idea. You could add a file `benchopt/datasets_util/libsvm.py`? 

Feel free to come up with better naming for everything :)

benchopt/config.py
```diff
@@ -25,7 +26,7 @@
 
 
 def get_global_setting(name):
-    assert name in DEFAULT_GLOBAL, f"Unkonwn config key {name}"
+    assert name in DEFAULT_GLOBAL, f"Unknown config key {name}"
```
tomMoral: Nice!

benchopt/dataset_util.py
```diff
@@ -0,0 +1,41 @@
...
+N_FEATURES = {'finance': 4272227,
+              'news20': 1355191,
+              'rcv1_train': 47236,
+              'kdda_train': 20216830}
+
+
+def download_libsvm(X_path, y_path, name, replace=False):
+    url = ("https://www.csie.ntu.edu.tw/~cjlin/libsvmtools/datasets/" +
+           NAMES[name] + ".bz2")
+    path = download(url, pjoin(DATA_DIR, name), replace=replace)
```
tomMoral: Don't you need to create `pjoin(DATA_DIR, name)` before downloading?

benchopt/dataset_util.py
```diff
@@ -0,0 +1,41 @@
+import numpy as np
+from scipy import sparse
+from bz2 import BZ2Decompressor
+from os.path import join as pjoin
```
tomMoral: I prefer `import os` and using `os.path.join`.
It is more verbose but avoid custom names or having to import each command separately.

benchopt/dataset_util.py
```diff
@@ -0,0 +1,41 @@
...
+              'kdda_train': 20216830}
+
+
+def download_libsvm(X_path, y_path, name, replace=False):
+    url = ("https://www.csie.ntu.edu.tw/~cjlin/libsvmtools/datasets/" +
+           NAMES[name] + ".bz2")
+    path = download(url, pjoin(DATA_DIR, name), replace=replace)
+
+    decompressor = BZ2Decompressor()
+    with open(pjoin(path, NAMES[name].split('/')[-1]), "r+b") as f, \
```
tomMoral: Maybe compute the path for the decompressed file above.

benchopt/dataset_util.py
```diff
@@ -0,0 +1,41 @@
...
+            open(path, "rb") as g:
+        for data in iter(lambda: g.read(100 * 1024), b''):
+            f.write(decompressor.decompress(data))
+
+        X, y = load_svmlight_file(f, N_FEATURES[name])
+        X = sparse.csc_matrix(X)
+        X.sort_indices()
+        sparse.save_npz(X_path, X)
+        np.save(y_path, y)
```
tomMoral: As you save here `X,y`, do you really need to save the decompressed file? If it is not useful, it is only increasing the size on the disk while. Maybe consider decompressing in a [tempfile](https://docs.python.org/3.7/library/tempfile.html#tempfile.TemporaryFile)?

benchopt/dataset_util.py
```diff
@@ -0,0 +1,41 @@
+import numpy as np
+from scipy import sparse
+from bz2 import BZ2Decompressor
+from os.path import join as pjoin
+
+from download import download
```
tomMoral: Bad import

benchopt/dataset_util.py
mathurinm: `download.download` does it for us: from tehir doc `This will download a file and store it in a '~/data/` folder, creating directories if need be`

benchopt/dataset_util.py
mathurinm: addressed

benchopt/dataset_util.py
mathurinm: ok, avoid splitting line too

benchopt/dataset_util.py
mathurinm: or delete manually `decompressed_path`?

benchopt/dataset_util.py
mathurinm: I put it below since it's not a standard library, how should I do ?

tomMoral: No I though it was not the right name. I did not know `download`. I would group it with the one above (`numpy` is not a stdlib import either) but this is nitpicking.

benchopt/dataset_util.py
tomMoral: either way. The nice thing about tempfile is that you just create it (without opening it) and it works like an open file and clean itself up. But either way is fine by me.

benchmarks/lasso/datasets/finance.py
```diff
@@ -1,6 +1,6 @@
 import numpy as np
+import os
```
tomMoral: My TOC are going to be exposed but I would move it a line above... ;)

benchopt/dataset_util.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
+import os
```
tomMoral: same

benchopt/dataset_util.py
```diff
@@ -25,11 +25,11 @@
...
     url = ("https://www.csie.ntu.edu.tw/~cjlin/libsvmtools/datasets/" +
            NAMES[name] + ".bz2")
-    path = download(url, pjoin(DATA_DIR, name), replace=replace)
+    path = download(url, os.path.join(DATA_DIR, name), replace=replace)
 
     decompressor = BZ2Decompressor()
-    with open(pjoin(path, NAMES[name].split('/')[-1]), "r+b") as f, \
-            open(path, "rb") as g:
+    decomp_path = os.path.join(path, NAMES[name].split('/')[-1])
+    with open(decomp_path, "r+b") as f, open(path, "rb") as g:
```
tomMoral: much better!

tomMoral: I tend to prefer naming `g->original_file` and `f->temp_file` to be more explicit.

benchopt/dataset_util.py
tomMoral: Good!!!

benchopt/dataset_util.py
mathurinm: I went with os.remove for consistency since I could not use a TempFile for the downloaded compressed one.

mathurinm:
@tomMoral I don't know how to fix the ssl error (I don't test my download utils in celer)

mathurinm:
@tomMoral should we go for more liblinear regression datasets directly in this PR, renaming the class to LiblinearRegression with a `name` parameter? 
We can also add a few classification ones

tomMoral:
Ok, I think this PR is in a good state.
I would leave the extension to multiple datasets and the testing of the install for another PR.

agramfort:
LGTM

we should add codecov to have an idea of coverage

mathurinm:
Ok for me too

In the next PR maybe we should also handle passing dataset options to benchopt run, ie if the dataset becomes libsvm, how do I do if I want to run in only on finance and not all libsvm datasets.

tomMoral:
It should already work. It the dataset name is `LibSVM(finance)`, putting `-d finance` will run the benchmark on this dataset.

tomMoral:
For instance right now, if you want to run the benchmark only on one of the simulated dataset, you can put `-d n_features=10000` and only one is selected.

tomMoral:
Thanks @mathurinm !!

PR #8, opened 2020-03-04
[MRG] Add numba solver for cd

mathurinm:
Locally it's not converging on Boston for now


mathurinm:
Curves look nice @tomMoral , can you review ?

benchmarks/lasso/solvers/cd.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class Solver(BaseSolver):
+    name = "cd"
+
+    install_cmd = None
+
+    def set_objective(self, X, y, lmbd):
+        self.X, self.y, self.lmbd = X, y, lmbd
+        # self.run(1)
```
mathurinm: @tomMoral should we run 1 iteration to compile numba? Or is it neglectable ?

agramfort: yes do this to cache the compiled code

setup.py
```diff
@@ -28,7 +28,7 @@
     url=URL,
     packages=find_packages(),
     install_requires=['numpy', 'pandas', 'matplotlib', 'pyyaml',
-                      'click', 'joblib', 'scikit-learn'],
+                      'click', 'joblib', 'scikit-learn', 'numba'],
```
agramfort: we should avoid making these requirements global. This is solver specific

PR #11, opened 2020-03-10
[FIX] lasso name in objective class

mathurinm:


agramfort:
I pushed in master :)

PR #17, opened 2020-03-17
ENH add FISTA solver for Lasso benchmark

mathurinm:
I am doing something wrong since use_acceleration is never set to True:
```
(base) ➜  benchOpt git:(add_fista) ✗ benchopt run lasso -s baseline -n 1 -d boston
Boston
|--Lasso regression(reg=0.05)
|----Baseline(use_acceleration=false): done               
|----Baseline(use_acceleration=false): done               
|--Lasso regression(reg=0.1)
|----Baseline(use_acceleration=false): done               
|----Baseline(use_acceleration=false): done               
|--Lasso regression(reg=0.5)
|----Baseline(use_acceleration=false): done               
|----Baseline(use_acceleration=false): done    
```

We can also discuss the naming, I would go for `w` instead of  `z_hat`
 (same as in sklearn),  and I'm not very satisfied with `use_acceleration` either

### DIFF
--- a/benchmarks/lasso/solvers/baseline.py
+++ b/benchmarks/lasso/solvers/baseline.py
@@ -7,6 +7,12 @@ from benchopt.base import BaseSolver
 class Solver(BaseSolver):
     name = 'Baseline'
 
+    parameters = {'use_acceleration': [False, True]}
+
+    def __init__(self, use_acceleration):
+        self.use_acceleration = use_acceleration
+        super().__init__()
+
     def set_objective(self, X, y, lmbd):
         self.X, self.y, self.lmbd = X, y, lmbd
 
@@ -15,14 +21,21 @@ class Solver(BaseSolver):
     def run(self, n_iter):
         n_features = self.X.shape[1]
         z_hat = np.zeros(n_features)
-        for i in range(n_iter):
+        t_new = 1
+        for _ in range(n_iter):
+            if self.use_acceleration:
+                t_old = t_new
+                t_new = (1 + np.sqrt(1 + 4 * t_old ** 2)) / 2
+                z_old = z_hat.copy()
             grad = self.X.T.dot(self.X.dot(z_hat) - self.y)
             z_hat -= 1 / self.L * grad
             z_hat = self.st(z_hat, 1 / self.L * self.lmbd)
+            if self.use_acceleration:
+                z_hat = z_hat + (t_old - 1.) / t_new * (z_hat - z_old)
         self.z_hat = z_hat
 
     def st(self, z, mu):
-        return np.sign(z) * np.maximum(0, abs(z) - mu)
+        return np.sign(z) * np.maximum(0, np.abs(z) - mu)
 
     def get_result(self):
         return self.z_hat
--- a/benchmarks/lasso/solvers/baseline.py
+++ b/benchmarks/lasso/solvers/baseline.py
@@ -9,10 +9,6 @@ class Solver(BaseSolver):
 
     parameters = {'use_acceleration': [False, True]}
 
-    def __init__(self, use_acceleration):
-        self.use_acceleration = use_acceleration
-        super().__init__()
-
     def set_objective(self, X, y, lmbd):
         self.X, self.y, self.lmbd = X, y, lmbd
 


benchmarks/lasso/solvers/baseline.py
```diff
@@ -7,6 +7,12 @@
 class Solver(BaseSolver):
     name = 'Baseline'
 
+    parameters = {'use_acceleration': [False, True]}
+
+    def __init__(self, use_acceleration):
+        self.use_acceleration = use_acceleration
+        super().__init__()
```
tomMoral: Ok here is an interesting point....

I did some "crazy" stuff so you don't need to implement an `__init__` for the solver. But you cannot implement it.
Basically, it makes it easier to ensure the solver is picklable and it helps exposing the solver parameters in his name. So anything that you set in parameters `key` will be accessible as `self.key`.

To me, the `__init__` should not do anything else than stocking parameters and the base class do that for you. But if you think it is too complicated, we can discuss it.

benchmarks/lasso/solvers/baseline.py
```diff
@@ -15,14 +21,21 @@ def set_objective(self, X, y, lmbd):
...
             grad = self.X.T.dot(self.X.dot(z_hat) - self.y)
             z_hat -= 1 / self.L * grad
             z_hat = self.st(z_hat, 1 / self.L * self.lmbd)
+            if self.use_acceleration:
+                z_hat = z_hat + (t_old - 1.) / t_new * (z_hat - z_old)
         self.z_hat = z_hat
 
     def st(self, z, mu):
-        return np.sign(z) * np.maximum(0, abs(z) - mu)
+        return np.sign(z) * np.maximum(0, np.abs(z) - mu)
```
tomMoral: Just out of curiosity, does this change anything? I tend to use `abs` instead of the other but only out of habit and fear for longer lines ^^

benchmarks/lasso/solvers/baseline.py
tomMoral: This also makes sure that you have default parameters (the first of each list), which is why the test is failing.

benchmarks/lasso/solvers/baseline.py
agramfort: it's the same but it's more readable as it makes explicit that it works on array

benchmarks/lasso/solvers/baseline.py
mathurinm: Ok, IMO we need to edit the doc with an example of solver with parameters because I took inspiration from the `simulated` dataset and there, there is an init.

benchmarks/lasso/solvers/baseline.py
mathurinm: AFAIK there is no time difference

mathurinm:
It works fine of simulated,:
image

 but on Boston I get a single curve:
image

Command is `benchopt run lasso -d simulated -s baseline -n 1 --max-samples 3`

benchmarks/lasso/solvers/baseline.py
tomMoral: I agree, even I had forgotten about this, I had to look at the code.

tomMoral:
It looks like the boston curve have not been generated with the given command line no?
`benchopt run lasso -d simulated -s baseline -n 1 --max-samples 3`

I imagine that the dataset is just a typo, but also `-n 1` should not show any variance as there is only one repetition and it look like there is only 2 samples here. Can you repeat the run and make sure it is the right plot?

benchmarks/lasso/solvers/baseline.py
tomMoral: Actually, you can implement `__init__`, it just needs to have default values for all its parameters.

The fact that we don't implement it is to make sure we can report the parameters' value correctly but it might be tricky to enforce (don't feel very pythonic, default small value might be different to run test).
Something that would probably be easier is to use a decorator that calls `super().__init__` and log the parameters. This decorator would be used over the `__init__` function.

mathurinm:
`benchopt run lasso -d simulated -s baseline -n 3 --max-samples 3`

image

`benchopt run lasso -d boston -s baseline -n 3 --max-samples 3`
image



BUT:
`benchopt run lasso -d boston -s baseline -n 1 --max-samples 3`
image


So I thought something was cached, I run it with `-f` (hence my other comment that I have to wait for all other solvers install)
`benchopt run lasso -d boston -f baseline -n 1 --max-samples 3`
image

and then run with `-s` again, still encountering an issue

image

tomMoral:
Could you run `benchopt run lasso -s baseline -f baseline -d boston -n 1 --max-samples 3`

mathurinm:
It works:
image


Remains to have the same in logreg and we can merge ?

tomMoral:
I would say one solver/benchmark per PR so I am in favor of merging this one.

agramfort:
thx @mathurinm !

PR #18, opened 2020-03-18
CI enable codecov reports

agramfort:


continuous_integration/test_script.cmd
```diff
@@ -1,5 +1,6 @@
-set DEFAULT_PYTEST_ARGS=-vlrx --timeout=60
+set DEFAULT_PYTEST_ARGS=-vlrx --timeout=60 --cov=benchopt
 
 call activate %VIRTUALENV%
 
 pytest --junitxml=%JUNITXML% %DEFAULT_PYTEST_ARGS%
+codecov
```
tomMoral: This script is not run as it is for windows CI.

Also, in pipeline, I'd rather add a separate step in the pipeline that only run if the test step was successful.

dev-requirements.txt
```diff
@@ -3,3 +3,4 @@ cython
 scikit-learn
 pytest
 pytest-timeout
+codecov
```
tomMoral: You need to add the pytest extension `pytest-cov` as a dependency I think.

agramfort:
nice !

now let's increase coverage. viz for example

see https://codecov.io/gh/benchopt/benchOpt/tree/230c6d916118bce1803508238ef09b8283a9da17/benchopt

PR #19, opened 2020-03-20
CLN update Lasso:Baseline solver for a better doc

mathurinm:
-  added a comment to mention that parameters defined in `parameters` can be accessed without being set in an init.
-  replaced z by w
- avoided computation of X.T @ X for Lipschitz constant (costly)

### DIFF
--- a/benchmarks/lasso/solvers/baseline.py
+++ b/benchmarks/lasso/solvers/baseline.py
@@ -5,33 +5,34 @@ from benchopt.base import BaseSolver
 
 
 class Solver(BaseSolver):
-    name = 'Baseline'
+    name = 'Baseline'  # proximal gradient, eventually accelerated
 
+    # any parameter defined here is accessible as a class attribute
     parameters = {'use_acceleration': [False, True]}
 
     def set_objective(self, X, y, lmbd):
         self.X, self.y, self.lmbd = X, y, lmbd
 
-        self.L = np.linalg.norm(self.X.dot(self.X.T), ord=2)
+        self.L = np.linalg.norm(self.X, ord=2) ** 2
 
     def run(self, n_iter):
         n_features = self.X.shape[1]
-        z_hat = np.zeros(n_features)
+        w = np.zeros(n_features)
         t_new = 1
         for _ in range(n_iter):
             if self.use_acceleration:
                 t_old = t_new
                 t_new = (1 + np.sqrt(1 + 4 * t_old ** 2)) / 2
-                z_old = z_hat.copy()
-            grad = self.X.T.dot(self.X.dot(z_hat) - self.y)
-            z_hat -= 1 / self.L * grad
-            z_hat = self.st(z_hat, 1 / self.L * self.lmbd)
+                w_old = w.copy()
+            grad = self.X.T.dot(self.X.dot(w) - self.y)
+            w -= 1 / self.L * grad
+            w = self.st(w, 1 / self.L * self.lmbd)
             if self.use_acceleration:
-                z_hat = z_hat + (t_old - 1.) / t_new * (z_hat - z_old)
-        self.z_hat = z_hat
+                w = w + (t_old - 1.) / t_new * (w - w_old)
+        self.w = w
 
-    def st(self, z, mu):
-        return np.sign(z) * np.maximum(0, np.abs(z) - mu)
+    def st(self, w, mu):
+        return np.sign(w) * np.maximum(0, np.abs(w) - mu)
 
     def get_result(self):
-        return self.z_hat
+        return self.w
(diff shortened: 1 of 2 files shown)


benchmarks/lasso/solvers/baseline.py
```diff
@@ -5,33 +5,34 @@
...
-            z_hat -= 1 / self.L * grad
-            z_hat = self.st(z_hat, 1 / self.L * self.lmbd)
+                w_old = w.copy()
+            grad = self.X.T.dot(self.X.dot(w) - self.y)
+            w -= 1 / self.L * grad
+            w = self.st(w, 1 / self.L * self.lmbd)
             if self.use_acceleration:
-                z_hat = z_hat + (t_old - 1.) / t_new * (z_hat - z_old)
-        self.z_hat = z_hat
+                w = w + (t_old - 1.) / t_new * (w - w_old)
```
agramfort: you should use inplace operations as much as possible.

benchmarks/lasso/solvers/baseline.py
```diff
@@ -5,33 +5,34 @@
...
+        self.w = w
 
-    def st(self, z, mu):
-        return np.sign(z) * np.maximum(0, np.abs(z) - mu)
+    def st(self, w, mu):
+        return np.sign(w) * np.maximum(0, np.abs(w) - mu)
```
agramfort: st could operate inplace to avoid unecessary memory allocations

doc/how.rst
```diff
@@ -61,7 +61,7 @@ from the union of the data and the objective parameters.
 
 You can define your custom solver.
 
-Example of hand written solver
+Example of hand written and parametrized solver
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
agramfort: missing ~~~~

benchmarks/lasso/solvers/baseline.py
```diff
@@ -5,33 +5,34 @@
...
+    name = 'Baseline'  # proximal gradient, eventually accelerated
 
+    # any parameter defined here is accessible as a class attribute
     parameters = {'use_acceleration': [False, True]}
 
     def set_objective(self, X, y, lmbd):
         self.X, self.y, self.lmbd = X, y, lmbd
 
-        self.L = np.linalg.norm(self.X.dot(self.X.T), ord=2)
+        self.L = np.linalg.norm(self.X, ord=2) ** 2
```
mathurinm: @agramfort IMO this is not fair that the constant is precomputed, this should be done in `run()`. what's your take?

agramfort:
I agree

>

benchmarks/lasso/solvers/baseline.py
mathurinm: done

benchmarks/lasso/solvers/baseline.py
mathurinm: I thought of using `shrink = np.maximum(0, 1 - mu / np.abs(w))` but this causes division by 0 and working with masks causes creation of other arrays.

In hindsight the gradient should be the larger cost of the iteration, have you observed the contrary in the past?

agramfort: i doub't know how much impact this will have true. But here you make 4 copies of w in one line

mathurinm:
```In [5]: a = np.array([0., 1.])                                                  

In [6]: a *= np.maximum(0, 1 - 0.5 / np.abs(a))                                 
/home/mathurin/miniconda3/bin/ipython:1: RuntimeWarning: divide by zero encountered in true_divide
  #!/home/mathurin/miniconda3/bin/python

In [7]: a                                                                       
Out[7]: array([0. , 0.5])
```
but this raises a warning bc of mu / 0

I think hopefully numpy reuses temp arrays now and only 2 arrays are needed

benchmarks/lasso/solvers/baseline.py
```diff
@@ -5,33 +5,34 @@
...
         self.X, self.y, self.lmbd = X, y, lmbd
 
-        self.L = np.linalg.norm(self.X.dot(self.X.T), ord=2)
-
     def run(self, n_iter):
+        L = np.linalg.norm(self.X, ord=2) ** 2
+
         n_features = self.X.shape[1]
-        z_hat = np.zeros(n_features)
+        w = np.zeros(n_features)
```
tomMoral: If we decide to go that way I would say that this need to be  changed globally in the benchmark. I think I used `z_hat` everywhere but we can do this in a separate PR.

benchmarks/lasso/solvers/baseline.py
tomMoral: Inplace soft thresholding:
```python
def st(w, mu):
    a -= np.clip(w, -mu, mu)
    return a
```

benchmarks/lasso/solvers/baseline.py
mathurinm: +1 to merge this one and do another PR

tomMoral:
Thx!

PR #20, opened 2020-03-20
[WIP] add Gap Safe Rules Lasso solver

mathurinm:
For now this is an exercise for me to check that it's understandable how to add a solver from the code base
and a way to check with @EugeneNdiaye that the gsroptim code works as expected, and to figure out what is the best strategy to use it for single lambda problems. So far I have used a grid of 100 lambdas between lambda_max and the desired lambda: what do you think of this @EugeneNdiaye ?

### DIFF
--- a/benchmarks/lasso/solvers/gapsafe.py
+++ b/benchmarks/lasso/solvers/gapsafe.py
@@ -23,7 +23,7 @@ class Solver(BaseSolver):
     def run(self, n_iter):
         coefs_ = lasso_path(
             self.X, self.y, [self.lmbd], max_iter=n_iter, eps=1e-14,
-            fit_intercept=False)[1]
+            fit_intercept=False, verbose=False)[1]
         self.coef_ = coefs_[:, -1]
 
     def get_result(self):
--- a/benchmarks/lasso/solvers/gapsafe.py
+++ b/benchmarks/lasso/solvers/gapsafe.py
@@ -19,15 +19,10 @@ class Solver(BaseSolver):
 
     def set_objective(self, X, y, lmbd):
         self.X, self.y, self.lmbd = X, y, lmbd
-        self.lmbd_max = np.max(np.abs(X.T @ y))
 
     def run(self, n_iter):
-        # TODO determine optimal strategy to solve single lambda.
-        # Right now, take a geom grid of 100 lambdas from lmb_max
-        # to lmbd
-        lmbds = self.lmbd_max * np.geomspace(1, self.lmbd / self.lmbd_max,
-                                             num=100)
-        coefs_ = lasso_path(self.X, self.y, lmbds, max_iter=n_iter)
+        coefs_ = lasso_path(
+            self.X, self.y, [self.lmbd], max_iter=n_iter, eps=1e-14)
         self.coef_ = coefs_[:, -1]
 
     def get_result(self):
--- a/benchmarks/lasso/solvers/gapsafe.py
+++ b/benchmarks/lasso/solvers/gapsafe.py
@@ -9,7 +9,7 @@ with safe_import():
 
 
 class Solver(BaseSolver):
-    name = 'Gap Safe'
+    name = 'Gapsafe'
     sampling_strategy = 'iteration'
 
     install_cmd = 'pip'
--- a/benchmarks/lasso/solvers/gapsafe.py
+++ b/benchmarks/lasso/solvers/gapsafe.py
@@ -1,0 +1,34 @@
+import numpy as np
+
+from benchopt.base import BaseSolver
+from benchopt.util import safe_import
+
+
+with safe_import():
+    from gsroptim.lasso import lasso_path  # TODO make it available in init?
+
+
+class Solver(BaseSolver):
+    name = 'Gap Safe'
+    sampling_strategy = 'iteration'
+
+    install_cmd = 'pip'
+    requirements = ['gsroptim']
+    requirements_install = [
+        'git+https://github.com/EugeneNdiaye/Gap_Safe_Rules.git']
+
+    def set_objective(self, X, y, lmbd):
+        self.X, self.y, self.lmbd = X, y, lmbd
+        self.lmbd_max = np.max(np.abs(X.T @ y))
+
+    def run(self, n_iter):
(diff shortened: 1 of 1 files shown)


EugeneNdiaye:
- A "best strategy to use it for single lambda problems" is unclear. Just use one single lambda by default.

- One should be able to run the benchmark with a path of lambda right?

mathurinm:
My intuition is that if the path is chosen in a reasonable fashion, and strong active warm start is used, using a path could be much faster than doing the computation for a single lambda (especially if this lambda is small, in a spirit similar to that of working sets). 
I just wanted to check if you had such a heuristic in mind. 

For full Lasso paths, it should be another benchmark which is not developped yet, as the curves we want to plot are not so obvious (one idea coming to mind is "maximum suboptimality along path" as a function of time)

EugeneNdiaye:
- "I just wanted to check if you had such a heuristic in mind."
As far as I know, there is no clear strategy for design a "grid-speedup" ie pre-selecting a grid of hyperparameter such that Time(path to lambda) <= Time(lambda) with some guarantee (Open question). I may have some heuristic to test.

Other possibilities is an aggressive screening that I will push soon (see appendix of the paper sent by email)

For full Lasso paths, yes maximum suboptimality along path may be a good option but it will not highlight the time needed to compute the full path at some precision.  Will think about that

mathurinm:
I cannot run the solver:
```
(base) ➜  benchOpt git:(gapsafe) ✗ benchopt run lasso -s gapsafe -d boston   
Installing Boston in lasso:... done
Boston
|--Lasso regression(reg=0.05)
|--Lasso regression(reg=0.1)
|--Lasso regression(reg=0.5)
  File "/home/mathurin/workspace/benchOpt/.venv/lasso/bin/benchopt", line 11, in <module>
    load_entry_point('benchopt', 'console_scripts', 'benchopt')()
  File "/home/mathurin/workspace/benchOpt/benchopt/cli.py", line 107, in start
    main()
  File "/home/mathurin/workspace/benchOpt/.venv/lasso/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/mathurin/workspace/benchOpt/.venv/lasso/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/mathurin/workspace/benchOpt/.venv/lasso/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mathurin/workspace/benchOpt/.venv/lasso/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mathurin/workspace/benchOpt/.venv/lasso/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/mathurin/workspace/benchOpt/benchopt/cli.py", line 75, in run
    max_samples=max_samples, n_rep=repetition)
  File "/home/mathurin/workspace/benchOpt/benchopt/runner.py", line 214, in run_benchmark
    plot_benchmark(df, benchmark)
  File "/home/mathurin/workspace/benchOpt/benchopt/viz.py", line 11, in plot_benchmark
    datasets = df.data.unique()
  File "/home/mathurin/workspace/benchOpt/.venv/lasso/lib/python3.7/site-packages/pandas/core/generic.py", line 5274, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'DataFrame' object has no attribute 'data'
```

benchmarks/lasso/solvers/gapsafe.py
```diff
@@ -0,0 +1,29 @@
...
+from benchopt.base import BaseSolver
+from benchopt.util import safe_import
+
+
+with safe_import():
+    from gsroptim.lasso import lasso_path  # TODO make it available in init?
+
+
+class Solver(BaseSolver):
+    name = 'Gap Safe'
```
tomMoral: For now, when using `-s gapsafe`, this is checking that `name.lower == 'gapsafe'` so this fails because of the 'space'.

The error you see is simply that the `DataFrame` is empty. I will try to put a better error report and raise an error when no solver is selected.

tomMoral: `-s safe` should work.

mathurinm: Ah, I see. I changed the name to `gapsafe` 
I wasn't able to determine the cause of the error by myself so showing an error when an unknown solver is passed is probably a good idea!

tomMoral: Also having two separate fields : `id` that is matchable with `-s` and `name` or `display_name`? WDYT?

mathurinm:
I'm in favor of merging now and adding the more refined strategy (solving on a grid of lambdas rather than for a single one) later. @EugeneNdiaye is that OK for you?

agramfort:
do you some plot to share? just to see how it looks in a bench

mathurinm:
image

agramfort:
Nice 👍

EugeneNdiaye:
> I'm in favor of merging now and adding the more refined strategy (solving on a grid of lambdas rather than for a single one) later. @EugeneNdiaye is that OK for you?

Yes it seems better. I will later add the second gapsafe_solver with the pathwise strategy.

mathurinm:
In fact it is the pathwise strategy which is used since it is now in master of Gap Safe Rules.

@tomMoral @agramfort ready for review

EugeneNdiaye:
I have tried another bench: just run the algorithms until they achieves a prescribed precision and report the computational times. The results are not consistent with benchopt:

- I run on Leukemia dataset (I failed to add leukemia.py properly so I just put it straight in boston.py)

basiccd = gapsafe(screen_method=None)
path gapase = gapsafe(default)

WDYT about the differences? (I am affraid that the screening is broken when we run it on benchopt, but I do see why it is so).
checktime
checktime_benchopt

mathurinm:
This belongs to benchmark_lasso anyways, and I would not add too many solvers there. We made gsroptim easy to use in a benchmark and that was the most of it.

Issue #22, opened 2020-03-20
API newly coded solver is not run and no error message is shown


mathurinm:
To test #20 I ran the following command and gapsafe solver was not included in the benchmark. Is it the desired behavior ?

```
(base) ➜  benchOpt git:(gapsafe) ✗ benchopt run lasso -s celer -s gapsafe -d simulated
Simulated(n_samples=100,n_features=5000)
|--Lasso regression(reg=0.05)
|----Celer: done                                          
|--Lasso regression(reg=0.1)
|----Celer: done                                          
|--Lasso regression(reg=0.5)
|----Celer: done                                          
Simulated(n_samples=100,n_features=10000)
|--Lasso regression(reg=0.05)
|----Celer: done                                          
|--Lasso regression(reg=0.1)
|----Celer: done                                          
|--Lasso regression(reg=0.5)
|----Celer: done   
```

tomMoral:
Closing as this has been implemented as a by product in #47 (b1bbb19).

PR #23, opened 2020-03-20
ENH raise warning if an unknown solver is passed

mathurinm:
Addresses the behavior I could not understand in #22 

Is it the good place to do it, and is a warning enough?

### DIFF
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -177,9 +177,9 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
     if dataset_names is not None:
         unknown_datasets = set(dataset_names) - \
             set([ds.name.lower() for ds in datasets])
-        if unknwon_datasets:
+        if unknown_datasets:
             raise ValueError(
-                "Unknown dataset(s): %s" % ', '.join(unknwon_datasets))
+                "Unknown dataset(s): %s" % ', '.join(unknown_datasets))
 
     # Load the solvers and filter them to get the one to run
     solver_classes = list_benchmark_solvers(benchmark)
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -174,6 +174,13 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
     objective_class = get_benchmark_objective(benchmark)
     datasets = list_benchmark_datasets(benchmark)
 
+    if dataset_names is not None:
+        unknwon_datasets = set(dataset_names) - \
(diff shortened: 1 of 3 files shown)


benchopt/util.py
```diff
@@ -274,7 +274,9 @@ def filter_solvers(solvers, solver_names=None, forced_solvers=None,
     if len(solver_names) > 0:
         solvers = [s for s in solvers
                    if s.name.lower() in solver_names + forced_solvers]
-
+        not_found = list(set(solver_names + forced_solvers) - set(solvers))
+        if not_found:
+            warnings.warn('Solver(s) not found: %s' % ', '.join(not_found))
```
tomMoral: Arf sorry wrong button..

I was saying, an error is better as the user requested the solver explicitly.

Also, could you add a test for this? This will be easier with an error.

mathurinm:
@tomMoral here's how I would have written the test, but it doesn't pass since the ValueError is raised "after" Popen is called. Do you think it should be tested with `run`, like above ? How does one directly test the CLI ?

benchopt/runner.py
```diff
@@ -174,6 +174,13 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
     objective_class = get_benchmark_objective(benchmark)
     datasets = list_benchmark_datasets(benchmark)
 
+    if dataset_names is not None:
+        unknwon_datasets = set(dataset_names) - \
```
TomDLT:
```suggestion
        unknown_datasets = set(dataset_names) - \
```
here and there

benchopt/tests/test_cli.py
```diff
@@ -12,3 +13,8 @@ def test_invalid_benchmark():
 def test_invalid_dataset():
     with pytest.raises(SystemExit, match=r"2"):
         run(['lasso', '-d', 'invalid_dataset', '-s', 'baseline'], 'benchopt')
+
+
+def test_invalid_solver():
+    np.testing.assert_raises(ValueError, subprocess.Popen,
+                             'benchopt run lasso -s sdjglsdglxcnv'.split())
```
tomMoral, line 20: With `pytest`, you should use `pytest.raises` to check for raised Errors.

For the test, you can check the output of Popen like this

```python
p = subprocess.Popen('benchopt run lasso -s sdjglsdglxcnv'.split(),
					 stderr=subprocess.PIPE)
stdout, stderr = p.communicate()
assert 'ValueError' in stderr.decode()
```

tomMoral:
Closing this PR as this feature has been implemented in #47

Issue #21, opened 2020-03-20
ENH simulated dataset has nearly diagonal covariance


mathurinm:
I understand that's it a toy dataset, but we could introduce a bit of correlation in the data: the solvers run in < 0.01 s and the results are not trustable. I wouldn't mind per se, if it weren't the first figure that people were shown. Additionally I wouldn't accept such a dataset in a paper I review.

We can easily construct a Toeplitz 2d array without scipy (corr_{ij} = rho ** |i - j|) and use it as covariance when generating data. Afterwards, if one wants, rho could be made a parameter of the dataset.

What is your take on this @tomMoral @agramfort ?

agramfort:
+1 to add toeplitz designs in simulations

>

tomMoral:
+1

tomMoral:
Done in #24

PR #24, opened 2020-03-20
ENH helper for non diagonal covariance in simulated

mathurinm:
Addressing #21, but it takes a long time (> 3 mins) to generate the data now (my guess is that an EVD of correlation matrix is performed, the later being 10k x 10k in some cases)

### DIFF
--- a/benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
+++ b/benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
@@ -26,11 +26,9 @@ class Dataset(BaseDataset):
 
     def get_data(self):
         rng = np.random.RandomState(self.random_state)
-        if self.rho == 0:
-            X = rng.randn(self.n_samples, self.n_features)
-        else:
-            X, y, _ = make_correlated_data(self.n_samples, self.n_features,
-                                           rho=self.rho, random_state=rng)
+
+        X, y, _ = make_correlated_data(self.n_samples, self.n_features,
+                                       rho=self.rho, random_state=rng)
 
         data = dict(X=X, y=y)
 
--- a/benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
+++ b/benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
@@ -1,7 +1,7 @@
 import numpy as np
 
 from benchopt.base import BaseDataset
-from benchopt.utils.datasets.simulated import gen_correlated_data
+from benchopt.utils.datasets.simulated import make_correlated_data
 
 
 class Dataset(BaseDataset):
@@ -13,8 +13,8 @@ class Dataset(BaseDataset):
     parameters = {
         'n_samples, n_features': [
             (100, 5000),
-            (100, 10000)],  # slow to simulate big correlated design
-        'rho': [0, 0.6]
+            (100, 10000)
+        ], 'rho': [0, 0.6]
     }
 
     def __init__(self, n_samples=10, n_features=50, random_state=27, rho=0):
@@ -29,8 +29,8 @@ class Dataset(BaseDataset):
         if self.rho == 0:
             X = rng.randn(self.n_samples, self.n_features)
         else:
-            X = gen_correlated_data(self.n_samples, self.n_features,
-                                    rho=self.rho, random_state=rng)
+            X = make_correlated_data(self.n_samples, self.n_features,
+                                     rho=self.rho, random_state=rng)
 
         y = rng.randn(self.n_samples)
 
--- a/benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
+++ b/benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
@@ -10,9 +10,10 @@ class Dataset(BaseDataset):
     # List of parameters to generate the datasets. The benchmark will consider
     # the cross product for each key in the dictionary.
     parameters = {
-        'n_samples, n_features': [
-            (100, 5000),
-            (100, 10000)],
+        'n_samples, n_features, corr': [
+            (100, 5000, 0),
+            (100, 5000, 0.6),
+            (100, 10000, 0)],  # slow to simulate big correlated design
     }
 
     def __init__(self, n_samples=10, n_features=50, random_state=27, corr=0):
--- a/benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
+++ b/benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
@@ -1,6 +1,7 @@
 import numpy as np
 
 from benchopt.base import BaseDataset
+from benchopt.utils.datasets.simulated import gen_correlated_data
 
 
 class Dataset(BaseDataset):
@@ -10,29 +11,26 @@ class Dataset(BaseDataset):
     # List of parameters to generate the datasets. The benchmark will consider
     # the cross product for each key in the dictionary.
     parameters = {
-        'n_samples, n_features, corr': [
-            (100, 5000, 0),
-            (100, 5000, 0.6),
-            (100, 10000, 0)],  # slow to simulate big correlated design
+        'n_samples, n_features': [
+            (100, 5000),
+            (100, 10000)],  # slow to simulate big correlated design
+        'rho': [0, 0.6]
     }
 
-    def __init__(self, n_samples=10, n_features=50, random_state=27, corr=0):
+    def __init__(self, n_samples=10, n_features=50, random_state=27, rho=0):
         # Store the parameters of the dataset
         self.n_samples = n_samples
         self.n_features = n_features
         self.random_state = random_state
-        self.corr = corr
+        self.rho = rho
 
     def get_data(self):
         rng = np.random.RandomState(self.random_state)
-        if self.corr == 0:
+        if self.rho == 0:
             X = rng.randn(self.n_samples, self.n_features)
         else:
-            # use Toeplitz covariance matrix:
-            idx = np.arange(self.n_features)
-            corr_mat = self.corr ** np.abs(idx[:, None] - idx)
-            X = rng.multivariate_normal(
-                np.zeros(self.n_features), corr_mat, self.n_samples)
+            X = gen_correlated_data(self.n_samples, self.n_features,
+                                    rho=self.rho, random_state=rng)
 
         y = rng.randn(self.n_samples)
 
--- a/benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
+++ b/benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
@@ -29,10 +29,8 @@ class Dataset(BaseDataset):
         if self.rho == 0:
             X = rng.randn(self.n_samples, self.n_features)
         else:
-            X = make_correlated_data(self.n_samples, self.n_features,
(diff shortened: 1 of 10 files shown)


benchmarks/lasso/datasets/simulated.py
```diff
@@ -12,22 +12,26 @@ class Dataset(BaseDataset):
     parameters = {
         'n_samples, n_features': [
             (100, 5000),
-            (100, 10000)]
+            (100, 10000)],
     }
 
-    def __init__(self, n_samples=10, n_features=50, random_state=27):
+    def __init__(self, n_samples=10, n_features=50, random_state=27, corr=0.5):
```
agramfort: you should use corr = 0 by default

benchmarks/lasso/datasets/simulated.py
```diff
@@ -12,22 +12,26 @@ class Dataset(BaseDataset):
...
         # Pass parameters that will be used in Dataset name
         super().__init__(n_samples=n_samples, n_features=n_features)
 
     def get_data(self):
-
+        # use Toeplitz covariance matrix:
+        idx = np.arange(self.n_features)
+        corr_mat = self.corr ** np.abs(idx[:, None] - idx)
         rng = np.random.RandomState(self.random_state)
-        X = rng.randn(self.n_samples, self.n_features)
```
agramfort: keep randn if corr == 0

agramfort:
what is the mechanism for testing this?

tomMoral:
This PR needs to be moved in [benchopt/benchmark_lasso](https://github.com/benchopt/benchmark_lasso).

Also, I realized today that a fast way to generate correlated matrices is to use AR processes :)
For an AR(1) of coefficient rho, the correlation matrix is the one we are looking for I guess:

```
rho = .6
res = []
for _ in range(10):
    res.append(X)
    X = rho * X + np.random.randn(p)
D = np.array(res)
D -= D.mean(axis=1, keepdims=True)
D /= D.std(axis=1, keepdims=True)
plt.matshow(D.dot(D.T) / p)
```

gives
image


WDYT?

mathurinm:
That looks nice ! I'll give it a try in the aforementioned repos

Le lun. 28 sept. 2020 à 19:15, Thomas Moreau <notifications@github.com> a
écrit :

tomMoral:
@mathurinm I did the change based on the AR process.
Could you review the code and tell me if you think like it is okay?

benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 
 from benchopt.base import BaseDataset
+from benchopt.utils.datasets.simulated import gen_correlated_data
```
mathurinm, line 4: maybe `make_correlated_data` as in sklearn.datasets.make_regression ?

benchopt/utils/checkers.py
```diff
@@ -65,3 +67,24 @@ def solver_supports_dataset(solver, dataset):
         return False
 
     return True
+
+
+def check_random_state(seed):
```
mathurinm, line 72: honest question, is it better to implement ours than to import it from sklearn.utils ?

benchopt/utils/checkers.py
```diff
@@ -65,3 +67,24 @@ def solver_supports_dataset(solver, dataset):
...
+
+    Parameters
+    ----------
+    seed : None, int or instance of RandomState
+        If seed is None, return the RandomState singleton used by np.random.
+        If seed is an int, return a new RandomState instance seeded with seed.
+        If seed is already a RandomState instance, return it.
+        Otherwise raise ValueError.
+    """
+    if seed is None or seed is np.random:
```
mathurinm, line 83: why `if seed is np.random` ?

benchopt/utils/datasets/simulated.py
```diff
@@ -0,0 +1,57 @@
+"""Helper to generate simulated data in benchopt."""
+import numpy as np
+
+from ..checkers import check_random_state
+
+
+def gen_correlated_data(n_samples=100, n_features=50, rho=0.6,
+                        random_state=None):
+    """Generate correlated design matrix with decaying correlation rho**|i-j|.
```
mathurinm, line 9: you could make it a verbatim string and avoid the double backslashes I think

benchopt/utils/datasets/simulated.py
```diff
@@ -0,0 +1,57 @@
...
+    random_state : int | RandomState instance | None (default)
+        Determines random number generation for centroid initialization and
+        random reassignment. Use an int to make the randomness deterministic.
+
+    Returns
+    -------
+    X: ndarray, shape (n_samples, n_features)
+        A design matrix with Toeplitz covariance.
+    """
+    if rho >= 1 or rho < 0:
```
mathurinm, line 43: `if not 0 <= rho < 1:` works and is easier IMO

benchopt/utils/datasets/simulated.py
```diff
@@ -0,0 +1,57 @@
...
+    n_samples: int
+        Number of samples in the design matrix.
+    n_features: int
+        Number of features in the design matrix.
+    corr: float
+        Correlation $\\rho$ between successive features. The element $C_{i, j}$
+        in the correlation matrix will be $\\rho^{|i-j|}$. This parameter
+        should be selected in $[0, 1[$.
+    random_state : int | RandomState instance | None (default)
+        Determines random number generation for centroid initialization and
```
mathurinm, line 35: maybe centroid is copy paster from somewhere else ? I don't get what it is

benchopt/utils/datasets/simulated.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    if rho >= 1 or rho < 0:
+        raise ValueError("The correlation `rho` should be chosen in [0, 1[.")
+    rng = check_random_state(random_state)
+
+    sigma = np.sqrt(1 - rho * rho)
+    innovation = rng.randn(n_features + 1, n_samples)
+    U = innovation[0]
+
+    res = [U.copy()]
```
mathurinm, line 51: maybe you could preallocate ?
```
X = np.empty([n_samples, n_features])
X[:, 0] = U
for i in range(1, n_features):
    U *= rho
    U += sigma * innovation[i]
    X[:, j] = U
return X
```

mathurinm: and eventually make X fortran (not sure)

benchopt/utils/datasets/simulated.py
```diff
@@ -0,0 +1,57 @@
...
+    -------
+    X: ndarray, shape (n_samples, n_features)
+        A design matrix with Toeplitz covariance.
+    """
+    if rho >= 1 or rho < 0:
+        raise ValueError("The correlation `rho` should be chosen in [0, 1[.")
+    rng = check_random_state(random_state)
+
+    sigma = np.sqrt(1 - rho * rho)
+    innovation = rng.randn(n_features + 1, n_samples)
```
mathurinm, line 48: any reason to generate them all in one go instead of calling rng for every column generation ?

benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
```diff
@@ -12,19 +13,25 @@ class Dataset(BaseDataset):
     parameters = {
         'n_samples, n_features': [
             (100, 5000),
-            (100, 10000)]
+            (100, 10000)],  # slow to simulate big correlated design
```
mathurinm, line 16: why slow ?
```
In [9]: %timeit -n 1 -r 15 gen_correlated_data(n_samples=100, n_features=10_000, rho=0.6)
91.8 ms ± 4.86 ms per loop (mean ± std. dev. of 15 runs, 1 loop each)
```

benchopt/utils/datasets/simulated.py
```diff
@@ -0,0 +1,57 @@
...
+    U = innovation[0]
+
+    res = [U.copy()]
+    for i in range(1, n_features):
+        U *= rho
+        U += sigma * innovation[i]
+        res.append(U.copy())
+    X = np.array(res)
+    return X.T
```
mathurinm, line 57: do we divide each column by sqrt(n_samples) ?

benchopt/utils/datasets/simulated.py
tomMoral: This is faster to have only one call than multiple calls as it reduces the overhead of calling the blass I would say but I did not benchmark.

benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
tomMoral: This is old  comment from your code, I forgot to remove.

benchopt/utils/datasets/simulated.py
tomMoral: We don't in the other generation process so I would say no.
But we can if you want.

benchopt/utils/datasets/simulated.py
tomMoral: It seems that this is not the case?!?
 
image

mathurinm:
One last thought is that we will always need y along with X, so we generate it as well, passing `snr` and `density` default parameters:
```
coef = np.zeros(n_features)
# random Gaussian coefficients:
coef[np.random.choice(n_features, int(density * n_features), replace=False) = np.random.randn(int(density * n_features))
noise = np.random.randn(n_samples)
y = X @ coef
y += noise / norm(noise) * norm(y) / snr
return X, y
```
we may also choose to return coef for inspection

mathurinm:
I fell this is what I have been needing for years. In sklearn make_regression I don't like the `noise` variable as it's not tangible. 

Generating y along with X seems handy, returning w_true may be useful for other metrics benchmark (spoiler)

mathurinm:
maybe density in (0, 1) is a better parameter than nnz. No strong optinion here

benchopt/utils/datasets/simulated.py
```diff
@@ -0,0 +1,79 @@
+"""Helper to generate simulated data in benchopt."""
+import numpy as np
+from numpy.linalg import norm
+
+from ..checkers import check_random_state
+
+
+def make_correlated_data(n_samples=100, n_features=50, rho=0.6, snr=3,
+                         w_true=None, nnz=10, random_state=None):
```
tomMoral, line 9: Maybe change the name to `make_correlated_regression`?
Also, I would rather have a `density` parameter I think, not a big fan of having to move this if we change the number of features.

For the generation, what do you think, should we have a fixed number of coef or something that can vary a bit around the "mean" that is density?

tomMoral (commented):
Just a small comment

mathurinm:
changed to `density`. I think it's easier to explain the generatin process if the number of non zero is always density * n_features. If the user wants something more complicated, he passes his own `w_true`

mathurinm:
Ready on my side @tomMoral

mathurinm:
Simple test if we want to add one:
```
def test_correlated():
    X, y, w_true = make_correlated_data(snr=0)
    np.testing.assert_allclose(y, X @ w_true)

    snr = 5
    w_true = np.ones(50)
    X, y, _ = make_correlated_data(n_features=w_true.shape[0], snr=5)
    np.testing.assert_allclose(snr, norm(X @ w_true), norm(y - X @ w_true))
```

tomMoral:
I did a small pass on tests.
Will try to merge this tomorrow.

tomMoral:
Thx a lot @mathurinm for this marathon PR! :)

Closes #21

PR #25, opened 2020-03-22
ENH add nnls benchmark with scipy/numba solvers

josephsalmon:
Creating the files for testing nnls (non negative least squares).

Can be run for instance with:
` benchopt run -l nnls -s baseline -s cd -s scipy -d boston --max-samples 32 --repetition 10`


benchmarks/nnls/solvers/prox_gradient_descent.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
+
+
+from benchopt.base import BaseSolver
+
+
+class Solver(BaseSolver):
+    name = 'Baseline'  # proximal gradient, eventually accelerated
```
agramfort: it's already in baseline no?

setup.py
```diff
@@ -28,7 +28,7 @@
     url=URL,
     packages=find_packages(),
     install_requires=['numpy', 'pandas', 'matplotlib',
-                      'click', 'joblib'],
+                      'click', 'joblib', 'scipy'],
```
agramfort: don't do this. It should be a dependency of the solver.

josephsalmon: so where shall I put the scipy requirement for the nnls_scipy.py to be able to run?

benchmarks/nnls/solvers/prox_gradient_descent.py
josephsalmon: yep, I have just removed it.

benchmarks/nnls/solvers/nnls_scipy.py
```diff
@@ -0,0 +1,22 @@
+from benchopt.base import BaseSolver
+from benchopt.util import safe_import
+
+
+with safe_import() as solver_import:
+    import scipy as scp
```
agramfort:
```suggestion
    from scipy import optimize
```

benchmarks/nnls/solvers/nnls_scipy.py
```diff
@@ -0,0 +1,22 @@
...
+    name = 'scipy'
+
+    install_cmd = 'pip'
+    requirements = ['scipy']
+
+    def set_objective(self, X, y):
+        self.X, self.y = X, y
+
+    def run(self, n_iter):
+        self.w = scp.optimize.nnls(self.X, self.y, maxiter=n_iter)[0]
```
agramfort:
```suggestion
        self.w = optimize.nnls(self.X, self.y, maxiter=n_iter)[0]
```

benchmarks/nnls/objective.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def set_data(self, X, y):
+        self.X, self.y = X, y
+
+    def __call__(self, beta):
+        # if (beta > 0).all():
+        #     diff = self.y - self.X.dot(beta)
+        #     return .5 * diff.dot(diff)
+        # else:
+        #     return np.inf
```
agramfort: Why not checking feasibility of solution ?

benchmarks/nnls/datasets/boston.py
```diff
@@ -0,0 +1,24 @@
+from benchopt.base import BaseDataset
```
mathurinm: @tomMoral with a growing number of benchmarks, won't it be difficult to maintain several slightly different versions of the same dataset?

benchmarks/nnls/solvers/baseline.py
```diff
@@ -0,0 +1,35 @@
...
+        for _ in range(n_iter):
+            if self.use_acceleration:
+                t_old = t_new
+                t_new = (1 + np.sqrt(1 + 4 * t_old ** 2)) / 2
+                w_old = w.copy()
+            grad = self.X.T.dot(self.X.dot(w) - self.y)
+            w -= grad / L
+            w = np.clip(w, a_min=0, a_max=None)
+            if self.use_acceleration:
+                w += (t_old - 1.) / t_new * (w - w_old)
```
mathurinm: @josephsalmon with this update rule I am not sure that this w is feasible. This is y_t in Beck and Teboulle notation, you need to return the iterate right after the prox has been applied (x_t, defined as w 2 lines above )

benchmarks/nnls/solvers/cd.py
```diff
@@ -0,0 +1,86 @@
...
+    @staticmethod
+    @njit
+    def cd(X, y, L, n_iter):
+        n_features = X.shape[1]
+        R = np.copy(y)
+        w = np.zeros(n_features)
+        for _ in range(n_iter):
+            for j in range(n_features):
+                old = w[j]
+                w[j] = positive_proj(w[j] + X[:, j] @ R / L[j])
```
mathurinm: Maybe you can just use max(0, ...) without defining a function

benchmarks/nnls/solvers/cd.py
```diff
@@ -0,0 +1,86 @@
...
+        w = np.zeros(n_features)
+        R = np.copy(y)
+        for _ in range(n_iter):
+            for j in range(n_features):
+                old = w[j]
+                start, end = X_indptr[j:j+2]
+                scal = 0.
+                for ind in range(start, end):
+                    scal += X_data[ind] * R[X_indices[ind]]
+                w[j] = positive_proj(w[j] + scal / L[j])
```
mathurinm: ditto

mathurinm:
I also observe that the benchmark take a very long time to run for the baseline solver, I had to stop it after 10 minutes and it was stuck at 40 %. `cd` runs in <0.01 s on the contrary

benchmarks/nnls/solvers/cd.py
agramfort: +1

benchmarks/nnls/solvers/baseline.py
agramfort: this could explain why the check of feasibility is commented in objective

benchmarks/nnls/objective.py
josephsalmon: it was debugging mode for a while and then I forgot...
will add it back.

benchmarks/nnls/solvers/cd.py
josephsalmon: done

benchmarks/nnls/solvers/cd.py
josephsalmon: done

benchmarks/nnls/objective.py
mathurinm: nitpick but the testing should be inverted: (beta < 0).any() is faster to evaluate

benchmarks/nnls/solvers/baseline.py
josephsalmon: the 2 issues are solved.

benchmarks/nnls/objective.py
```diff
@@ -18,7 +18,7 @@ def set_data(self, X, y):
         self.X, self.y = X, y
 
     def __call__(self, beta):
-        if (beta > 0).all():
+        if (beta < 0).any():
```
agramfort: I don't think it's correct. I would say:

`if (beta >= 0).all():`

benchmarks/nnls/objective.py
mathurinm: yes, I was not clear, if you use (beta < 0).any() the action associated is return np.inf

benchmarks/nnls/objective.py
mathurinm: yes, I was not clear, if you use (beta < 0).any() the action associated is return np.inf

josephsalmon:
Here's an extrat of what 
benchopt run -l nnls -d boston --max-samples 99 --repetition 10
produces
benchopt run -l nnls -d boston --max-samples 99 --repetition 10

mathurinm:
Given the way `scipy` nnls works, I don't think we can get more than one point on the curve:
if optimize.nnls gets a number of iterations which is too low, ie the KKT conditions are not satisfied when it reaches max_iter, then it does not return a solution but raises an exception:
see MODE returned value of the fortran routine https://jblevins.org/mirror/amiller/nnls.f90
and the scipy corresponding code: https://github.com/scipy/scipy/blob/v1.4.1/scipy/optimize/nnls.py#L79

So  either you get an exact solution, either you get an error. 

Any thoughts on dealing with this @agramfort @tomMoral ?

agramfort:
looks clean !

merge?

>

mathurinm:
@agramfort I don't think we should merge: if `boston` is not normalized, the benchmark does not complete. IMO this calls for a better sampling strategy of the convergence curve in slow solver cases.

I wouldn't push a failing benchmark to master

josephsalmon:
@mathurinm :  a possible light solution would be to decrease
MAX_ITER = int(1e6)
to 
MAX_ITER = int(1e5) or MAX_ITER = int(1e4)
in runner.py
(a million iteration or so in most solver is a lot).

mathurinm:
`MAX_ITER` can also be made problem dependent, based on `n_samples` and `n_features` ?

benchmarks/nnls/solvers/cd.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def njit(f):  # noqa: F811
+        return f
+
+
+class Solver(BaseSolver):
+    name = "cd"
+
+    install_cmd = 'pip'
+    requirements = ['numba']
```
TomDLT: You might need to add "scipy" here

josephsalmon: good catch. I added this also to the cd.py version in Lasso solvers, as this error was due to a copy pasted from that file.

benchmarks/nnls/solvers/nnls_scipy.py
```diff
@@ -0,0 +1,24 @@
...
+    install_cmd = 'pip'
+    requirements = ['scipy']
+
+    def set_objective(self, X, y):
+        self.X, self.y = X, y
+
+    def run(self, n_iter):
+        # XXX HACK scipy: when maxiter < 10 could get a
+        # RuntimeError: too many iterations
+        self.w = optimize.nnls(self.X, self.y, maxiter=n_iter + 10)[0]
```
agramfort: I still get some RuntimeError 

looking at scipy nnls fortran code it seems it runs without taking into account maxiter and if in the end it required more iterations than maxiter to converge you get an error. I would explain why all nnls solutions you get are at floating point accuracy whatever the choice of maxiter.

benchmarks/nnls/objective.py
```diff
@@ -0,0 +1,30 @@
...
+        self.X, self.y = X, y
+
+    def __call__(self, beta):
+        if (beta >= 0).all():
+            diff = self.y - self.X.dot(beta)
+            return .5 * diff.dot(diff)
+        else:
+            return np.inf
+        # diff = self.y - self.X.dot(beta)
+        # return .5 * diff.dot(diff)
```
agramfort: cleanup

benchmarks/nnls/solvers/nnls_scipy.py
josephsalmon: so I will simply drop the maxiter option...

benchmarks/nnls/solvers/nnls_scipy.py
```diff
@@ -0,0 +1,22 @@
...
+class Solver(BaseSolver):
+    name = 'scipy'
+
+    install_cmd = 'pip'
+    requirements = ['scipy']
+
+    def set_objective(self, X, y):
+        self.X, self.y = X, y
+
+    def run(self, n_iter):
```
agramfort: maybe put a comment of why n_iter is not used but still kept in the function signature

josephsalmon: done, adding:
 
# n_iter is not used here:
# optimize.nnls(self.X, self.y, maxiter=n_iter) could produce
# a RuntimeError for some dataset like Boston.

benchmarks/lasso/solvers/cd.py
```diff
@@ -30,7 +30,7 @@ class Solver(BaseSolver):
     name = "cd"
 
     install_cmd = 'pip'
-    requirements = ['numba']
+    requirements = ['numba', 'sparse']
```
TomDLT: I don't think that would work, `sparse` is not the package name, it is the module name.
```suggestion
    requirements = ['numba', 'scipy']
```

josephsalmon: Looks way better thanks @TomDLT  and @agramfort

TomDLT:
Running the benchmark, the nnls code prints a lot of convergence warnings, which looks like that:
```
|----Scipy:   0.0% (0 / 10)
 NNLS quitting on iteration count.
|----Scipy:   0.0% (1 / 10)
 NNLS quitting on iteration count.
|----Scipy:   0.0% (2 / 10)
 NNLS quitting on iteration count.
|----Scipy:   0.0% (3 / 10)
 NNLS quitting on iteration count.
|----Scipy:   0.0% (4 / 10)
 NNLS quitting on iteration count.
|----Scipy:   0.0% (5 / 10)
 NNLS quitting on iteration count.
|----Scipy:   0.0% (6 / 10)
 NNLS quitting on iteration count.
|----Scipy:   0.0% (7 / 10)
 NNLS quitting on iteration count.
|----Scipy:   0.0% (8 / 10)
 NNLS quitting on iteration count.
|----Scipy:   0.0% (9 / 10)
 NNLS quitting on iteration count.
|----Scipy:   3.1% (0 / 10)
 NNLS quitting on iteration count.
|----Scipy:   3.1% (1 / 10)
 NNLS quitting on iteration count.
|----Scipy:   3.1% (2 / 10)
 NNLS quitting on iteration count.
|----Scipy:   3.1% (3 / 10)
 NNLS quitting on iteration count.
|----Scipy:   3.1% (4 / 10)
 NNLS quitting on iteration count.
|----Scipy:   3.1% (5 / 10)
 NNLS quitting on iteration count.
|----Scipy:   3.1% (6 / 10)
 NNLS quitting on iteration count.
|----Scipy:   3.1% (7 / 10)
 NNLS quitting on iteration count.
|----Scipy:   3.1% (8 / 10)
 NNLS quitting on iteration count.
|----Scipy:   3.1% (9 / 10)
 NNLS quitting on iteration count.
|----Scipy:   6.2% (0 / 10)
 NNLS quitting on iteration count.
|----Scipy:   6.2% (1 / 10)
 NNLS quitting on iteration count.
|----Scipy:   6.2% (2 / 10)
 NNLS quitting on iteration count.
|----Scipy:   6.2% (3 / 10)
 NNLS quitting on iteration count.
|----Scipy:   6.2% (4 / 10)
 NNLS quitting on iteration count.
|----Scipy:   6.2% (5 / 10)
 NNLS quitting on iteration count.
|----Scipy:   6.2% (6 / 10)
 NNLS quitting on iteration count.
|----Scipy:   6.2% (7 / 10)
 NNLS quitting on iteration count.
|----Scipy:   6.2% (8 / 10)
 NNLS quitting on iteration count.
|----Scipy:   6.2% (9 / 10)
 NNLS quitting on iteration count.
```
Capturing the stdout is not straightforward since this is fortran code, but something like [this solution](https://stackoverflow.com/a/29834357/6156647) works.

I wonder if this is something worth adding to benchopt, since this is something that might be useful for other solvers.

agramfort:
open an issue for maybe later but this is not a blocker to me here

>

benchmarks/nnls/datasets/simulated.py
```diff
@@ -0,0 +1,35 @@
...
+class Dataset(BaseDataset):
+
+    name = "Simulated"
+
+    # List of parameters to generate the datasets. The benchmark will consider
+    # the cross product for each key in the dictionary.
+    parameters = {
+        'n_samples, n_features': [
+            (100, 5000),
+            (100, 10000)]
```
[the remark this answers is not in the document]
josephsalmon: done

benchmarks/nnls/objective.py
```diff
@@ -0,0 +1,28 @@
...
+    name = "Non Negative Least Squares"
+
+    parameters = {
+        'fit_intercept': [False],
+    }
+
+    def __init__(self, fit_intercept=False):
+        self.fit_intercept = fit_intercept
+
+        super().__init__()
```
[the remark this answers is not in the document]
josephsalmon: done

benchmarks/nnls/solvers/baseline.py
```diff
@@ -0,0 +1,36 @@
+import numpy as np
+
+
+from benchopt.base import BaseSolver
+
+
+class Solver(BaseSolver):
+    name = 'Baseline'  # proximal gradient, eventually accelerated
```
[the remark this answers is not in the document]
josephsalmon: done

benchmarks/nnls/objective.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __call__(self, beta):
+        if (beta >= 0).all():
+            diff = self.y - self.X.dot(beta)
+            return .5 * diff.dot(diff)
+        else:
+            return np.inf
+
+    def to_dict(self):
+        return dict(X=self.X, y=self.y)
```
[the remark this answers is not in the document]
josephsalmon: done

benchmarks/nnls/solvers/baseline.py
```diff
@@ -0,0 +1,36 @@
...
+
+    # any parameter defined here is accessible as a class attribute
+    parameters = {'use_acceleration': [False, True]}
+
+    def set_objective(self, X, y):
+        self.X, self.y = X, y
```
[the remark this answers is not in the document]
josephsalmon: no idea how to do that...so I added fit_interecept.

benchmarks/nnls/solvers/baseline.py
josephsalmon: You were so right. Sorry I copied that from @mathurinm  baseline in Lasso that needs to be corrected to.
Let me know if you are fine with the current version (actually w in the code is y_t, but the output value is correct now).
Might need some refactoring in a specific PR for the two FISTA implementation (nnls and lasso), wdyt?

mathurinm: I don't know if you got my comment, doing it this way @josephsalmon your `w - w_old` is 0

mathurinm: @tomMoral I am not sure I agree with what you wrote. Code is gone now but if you go to lasso.baseline, the code is correct to me. Before you enter the use_acceleration condition, w is x_t. Instead of the current line you could just write:
`w (y_t+1) = w (x_t) + ratio (w (x_t) - w_old (w_{t-1})`
but that would amount to the same, no ?

mathurinm:
@josephsalmon @tomMoral  
I think w is still w_old with the current code.
Does the following sound readable and correct to you, in the Lasso case:
```        
n_features = self.X.shape[1]
w = np.zeros(n_features)
w_acc = np.zeros(n_features)
w_old = np.zeros(n_features)
t_new = 1
for _ in range(n_iter):
    if self.use_acceleration:
        t_old = t_new
        t_new = (1 + np.sqrt(1 + 4 * t_old ** 2)) / 2
        w_old[:] = w  # x in Beck & Teboulle notation
        w[:] = w_acc  # y in Beck & Teboulle notation
    w -= self.X.T.dot(self.X.dot(w) - self.y) / L
    w = self.st(w, self.lmbd / L)
    if self.use_acceleration:
        w_acc[:] = w + (t_old - 1.) / t_new * (w - w_old)
self.w = w
```

josephsalmon:
I adapted @mathurinm 's version proposed above for nnls.  @tomMoral , it should be now consistent with Beck and Teboulle 2009 code. Do you agree?

benchmarks/nnls/solvers/cd.py
```diff
@@ -9,7 +9,10 @@
 with safe_import() as solver_import:
     from scipy import sparse
     from numba import njit
+    import warnings
```
tomMoral, line 12: Import no in the `safe_import` as it is reserved for solver deps.

benchmarks/nnls/solvers/cd.py
```diff
@@ -9,7 +9,10 @@
...
+    from numba.errors import PerformanceWarning
 
+    warnings.filterwarnings("ignore", category=PerformanceWarning)
```
tomMoral, line 15: Put this in `run`, with a context `warnings.catch_warning`.

benchmarks/nnls/solvers/baseline.py
```diff
@@ -5,32 +5,34 @@
...
+                w_old[:] = w  # x in Beck & Teboulle (2009) notation
+                w[:] = w_acc  # y in Beck & Teboulle (2009) notation
+            w -= self.X.T.dot(self.X.dot(w) - self.y) / L
             w = np.maximum(w, 0)
-            if i >= (n_iter - 1):
-                self.w = w.copy()
             if self.use_acceleration:
-                w += (t_old - 1.) / t_new * (w - w_old)
+                w_acc[:] = w + (t_old - 1.) / t_new * (w - w_old)
+        self.w = w
```
tomMoral, line 35: It seems better now.

benchmarks/nnls/datasets/boston.py
tomMoral: I agree that we should find better ways to refactor, but this is for another PR.

tomMoral:
I merged! thanks a lot @josephsalmon for taking the time! sorry for the slow review :disappointed: I will do better next time!

josephsalmon:
Glad it's through ;)
More work soon on that project.

Issue #27, opened 2020-03-23
Better message for install and import failure


TomDLT:
- Running the benchmark, both `cvxpy` and `cyanure` seem to fail at import, but the error is caught by `safe_import`. The error messages are not informing of the import error. I wonder why you use `safe_import`, and if it would not be more informative to raise an ImportError.
- The corresponding installing steps could also report the failure, like
 `Installing cvxpy in lasso:... failed`.

**Traceback:**
```pytb
Creating venv lasso:... done
Installing Boston in lasso:... done
Installing finance in lasso:... done
Installing Blitz in lasso:... done
Installing cd in lasso:... done
Installing Celer in lasso:... done
Installing cvxpy in lasso:... done
Installing Cyanure in lasso:... done
Installing Lightning in lasso:... done
Boston
|--Lasso regression(reg=0.05)
|----Baseline(use_acceleration=false): done               
|----Baseline(use_acceleration=true): done                
|----Blitz: done                                          
/home/tom/work/github/benchOpt/benchmarks/lasso/solvers/cd.py:60: NumbaPerformanceWarning: '@' is faster on contiguous arrays, called on (array(float64, 1d, A), array(float64, 1d, C))
  w[j] = st(w[j] + X[:, j] @ R / L[j], lmbd / L[j])
/home/tom/work/github/benchOpt/.venv/lasso/lib/python3.8/site-packages/numba/typing/npydecl.py:958: NumbaPerformanceWarning: '@' is faster on contiguous arrays, called on (array(float64, 1d, A), array(float64, 1d, C))
  warnings.warn(NumbaPerformanceWarning(msg))
|----Cd: done                                             
|----Celer: done                                          
|----Cvxpy: failed                                        
Traceback (most recent call last):
  File "/home/tom/work/github/benchOpt/benchopt/runner.py", line 131, in run_one_solver
    sample_curve, objective_value = run_one_sample(
  File "/home/tom/work/github/benchOpt/.venv/lasso/lib/python3.8/site-packages/joblib/memory.py", line 568, in __call__
    return self._cached_call(args, kwargs)[0]
  File "/home/tom/work/github/benchOpt/.venv/lasso/lib/python3.8/site-packages/joblib/memory.py", line 534, in _cached_call
    out, metadata = self.call(*args, **kwargs)
  File "/home/tom/work/github/benchOpt/.venv/lasso/lib/python3.8/site-packages/joblib/memory.py", line 734, in call
    output = self.func(*args, **kwargs)
[...]

PR #29, opened 2020-03-23
ENH Add install and import failure message

TomDLT:
Solves #27

Instead of not raising an error in `pip_install_in_env` when `PRINT_INSTALL_ERROR=False`, now the error is always raised, but caught in `DependenciesMixin.install`.

In a second part of this PR, in ` benchopt/runner.py`, it prevents the solver from running if the module caught an ImportError in `safe_import`.

### DIFF
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -9,6 +9,7 @@ from .util import bash_install_in_env
 from .util import pip_uninstall_in_env
 from .util import check_import_solver
 from .class_property import classproperty
+from .config import PRINT_INSTALL_ERROR
 
 
 # Possible sampling strategies
@@ -125,12 +126,18 @@ class DependenciesMixin:
         if force or not cls.is_installed(env_name=env_name):
             print(f"Installing {cls.name} in {env_name}:...",
                   end='', flush=True)
-            if cls.install_cmd == 'pip':
-                pip_install_in_env(*cls.requirements_install,
-                                   env_name=env_name)
-            elif cls.install_cmd == 'bash':
-                bash_install_in_env(cls.install_script, env_name=env_name)
-            print(" done")
+            try:
+                if cls.install_cmd == 'pip':
+                    pip_install_in_env(*cls.requirements_install,
+                                       env_name=env_name)
+                elif cls.install_cmd == 'bash':
+                    bash_install_in_env(cls.install_script, env_name=env_name)
+                print(" done")
+            except Exception as exception:
+                if PRINT_INSTALL_ERROR:
+                    raise exception
+                else:
+                    print(" failed")
         return cls.is_installed(env_name=env_name)
 
     @classmethod
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -123,7 +123,8 @@ class DependenciesMixin:
         if force:
             cls.uninstall(env_name=env_name)
 
-        if force or not cls.is_installed(env_name=env_name):
+        is_installed = cls.is_installed(env_name=env_name)
+        if force or not is_installed:
             print(f"Installing {cls.name} in {env_name}:...",
                   end='', flush=True)
             try:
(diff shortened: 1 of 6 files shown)


benchopt/runner.py
```diff
@@ -48,6 +49,12 @@ def colorify(message, color=BLUE):
 @mem.cache
 def run_repetition(objective, solver_class, solver_parameters, meta, sample):
 
+    # check if the module caught a failed import
+    module = importlib.import_module(solver_class.__module__)
+    if module.solver_import.failed_import:
+        raise ImportError(
+            f"Failure during import in {solver_class.__module__}.")
+
```
tomMoral: Why do you add this? It seems redundant with l.122 but I am maybe missing how you could end up in the situation where `run_repetition` is called while `run_one_solver` is still called.

benchopt/base.py
```diff
@@ -125,12 +126,18 @@ def install(cls, env_name=None, force=False):
...
+            try:
+                if cls.install_cmd == 'pip':
+                    pip_install_in_env(*cls.requirements_install,
+                                       env_name=env_name)
+                elif cls.install_cmd == 'bash':
+                    bash_install_in_env(cls.install_script, env_name=env_name)
+                print(" done")
+            except Exception as exception:
+                if PRINT_INSTALL_ERROR:
+                    raise exception
```
tomMoral: I am not totally sure for this `raise`, my feeling with `print_install_error` was to print them but to not fail the benchmark. In particular, if one call `benchopt run <benchmark>`, this will fail the whole benchmark while the solver was not specifically requested.
But it is true that this is an option that you need to activate so I think it would make sense.

Maybe I would also change `PRINT_INSTALL_ERROR` to `RAISE_INSTALL_ERROR`?

benchopt/runner.py
```diff
@@ -112,6 +119,13 @@ def progress(id_sample, delta):
         return max(id_sample / max_samples,
                    np.log(max(delta, eps)) / np.log(eps))
 
+    # check if the module caught a failed import
+    module = importlib.import_module(solver_class.__module__)
+    if module.solver_import.failed_import:
+        status = colorify("failed import", YELLOW)
```
tomMoral: I think for error I would put `RED` but I am not sure either.

benchopt/runner.py
TomDLT: Yes, I first added it to `run_one_solver` to raise an ImportError, then I went further to only print "failed" in `run_repetition`, but I figured I could leave it in `run_one_solver` in the case that it is called from another function.

benchopt/base.py
TomDLT: Oh yes I forgot to mention this.

When I run the benchmark with `PRINT_INSTALL_ERROR=True`, it definitely raises an error and stops the benchmark. With this behavior, I would indeed rename it `RAISE_INSTALL_ERROR`. If we prefer to just print the traceback and not raise an error, master also has to be fixed.

benchopt/runner.py
tomMoral: Ok it is fine by me.

benchopt/base.py
tomMoral: So I would go for `RAISE_INSTALL_ERROR`! :smile:

benchopt/runner.py
```diff
@@ -48,6 +49,12 @@ def colorify(message, color=BLUE):
 @mem.cache
 def run_repetition(objective, solver_class, solver_parameters, meta, sample):
 
+    # check if the module caught a failed import
+    module = importlib.import_module(solver_class.__module__)
+    if module.solver_import.failed_import:
```
tomMoral: Not sure if we should require to have the `solver_import` or if we should do the suggested change:
```suggestion
    if (hasattr('solver_import', module)
            and module.solver_import.failed_import):
```
Maybe it is better the way you propose as it is more systematic (all solvers have this `safe_import`) but it means that user need to see this even if the have no package to install.

benchopt/util.py
```diff
@@ -318,9 +316,16 @@ def delete_venv(env_name):
...
     for solver in solvers:
         force_install = solver.name.lower() in forced_solvers
-        solver.install(env_name=env_name, force=force_install)
+        success = solver.install(env_name=env_name, force=force_install)
+        successes.append(success)
+
+    if not all(successes):
+        print("Some solvers were not successfully installed, and will thus be "
+              "ignored. Use 'export BENCHO_RAISE_INSTALL_ERROR=true' to "
+              "stop at any installation failure and print the traceback.")
```
tomMoral: Should we use a warning so it is only print once and we can ignore it?

benchopt/runner.py
TomDLT: Probably better to not require `solver_import`, indeed.

benchopt/util.py
TomDLT: This is printed only once when we install all solvers, but I can change it into a warning.
What is the general guideline toward print statement in the project ?

tomMoral: Great! We don't have a policy yet but I would say avoid annoying  print? ^^'

tomMoral:
PErfect!! Thanks a lot @TomDLT !! :smile:

PR #38, opened 2020-05-02
Circle ci

agramfort:


agramfort:
hum ... https://4-227720375-gh.circle-artifacts.com/0/dev/index.html

tomMoral:
If I may, why do we use circleCI instead of github actions?

agramfort:
because i know how to do it fast in circleci :)

tomMoral:
It is a fair point!
There seems to be a strange thing with the artifact no?

agramfort:
doing it not from a fork

yes the artifact has a pb...

PR #40, opened 2020-05-03
ENH add statsmodels solver for lasso

josephsalmon:
I have added a comparison with `statsmodels` (now having lasso, enet, etc.) for versions 0.11.1
Results are in favor of `sklearn`. 

sklearn_vs_statsmodels_boston_lasso

### DIFF
--- a/benchmarks/lasso/solvers/statsmodels.py
+++ b/benchmarks/lasso/solvers/statsmodels.py
@@ -1,0 +1,30 @@
+from benchopt.base import BaseSolver
+from benchopt.util import safe_import
+
+
+with safe_import() as solver_import:
+    import statsmodels.api as sm
+
+
+class Solver(BaseSolver):
+    name = 'statsmodels'
+
+    install_cmd = 'pip'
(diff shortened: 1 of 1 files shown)


benchmarks/lasso/solvers/statsmodels.py
```diff
@@ -0,0 +1,30 @@
...
+with safe_import() as solver_import:
+    import statsmodels.api as sm
+
+
+class Solver(BaseSolver):
+    name = 'statsmodels'
+
+    install_cmd = 'pip'
+    requirements = ['statsmodels']
+    requirements_import = ['statsmodels']
```
tomMoral, line 14: You don't need this line I think. You only need to specify this if it is different from `requirements`, as `scikit-learn, sklearn`.

josephsalmon: I see, I 've removed it thanks.

tomMoral:
PErfect! Thanks @josephsalmon :) You are on a roll these days!

PR #42, opened 2020-05-07
FIX NumbaPerformanceWarning import

TomDLT:


agramfort:
Shouldn’t we fix the code so there are no warning ?

mathurinm:
It is surprising that a warning is raised since we pass X as fortran, so `X[:, j]` is contiguous

agramfort:
can you replicate @mathurinm ?

tomMoral:
Don't we need to do the same in : benchmarks/lasso/solvers/cd.py

tomMoral:
Maybe an helper for this would make sense? Something like `benchopt/helpers/helpers_numba.py` with a `filter_performance_warning` context?

TomDLT:
I fixed the warnings with ordering the input array `X` using Fortran order.
It is consistently faster with the CD solver, which computes the gradient more quickly if the column are contiguous.

Figure_1
Figure_2
Figure_1
Figure_2
Figure_3

josephsalmon:
good catch @TomDLT !

benchmarks/lasso/solvers/cd.py
```diff
@@ -31,7 +29,8 @@ class Solver(BaseSolver):
     requirements = ['numba', 'scipy']
 
     def set_objective(self, X, y, lmbd):
-        self.X, self.y, self.lmbd = X, y, lmbd
+        # use Fortran order to compute gradient on contiguous columns
+        self.X, self.y, self.lmbd = np.ascontiguousarray(X.T).T, y, lmbd
```
agramfort, line 33:
```suggestion
        self.X, self.y, self.lmbd = np.asfortranarray(X), y, lmbd
```

same remark for below

agramfort:
thx @TomDLT !

PR #43, opened 2020-05-07
ENH add sklearn solver for OLS/NNLS

TomDLT:


josephsalmon:
Thx @TomDLT ! Looks good too me.
The only minor issue for me is having only *one point* for this solver (the underlying BLAS function seems to rely on SVD).
Not sure how to proceed with such kind of solvers though.

agramfort:
@TomDLT can you look into the CI issue? thx

agramfort:
diff looks good but I think we should focus our efforts to produce meaningful and convincing figures to advertise what this piece of software enables.

TomDLT:
I fixed the error, adding a check for failed import in `test_solver`.

I also added a scikit-learn solver for NNLS, using `Lasso(alpha=1e-10, positive=True)`, which is hacky but converges to the same solution.

agramfort:
do you have some nice bench plots to share?

>

PR #44, opened 2020-05-08
CLN better computations for L constants in cd

josephsalmon:
adding @mathurinm propositio to remove linalg.norm in cd.py by substituting (x)**2.sum(axis =0) to linalg.norm(x)

### DIFF
--- a/benchmarks/lasso/solvers/cd.py
+++ b/benchmarks/lasso/solvers/cd.py
@@ -1,7 +1,5 @@
 import numpy as np
 
-from numpy.linalg import norm
-
 from benchopt.base import BaseSolver
 from benchopt.util import safe_import
 
@@ -39,13 +37,12 @@ class Solver(BaseSolver):
         self.run(1)
 
     def run(self, n_iter):
+        L = (self.X ** 2).sum(axis =0)
         if sparse.issparse(self.X):
-            L = sparse.linalg.norm(self.X, axis=0) ** 2
             self.w = self.sparse_cd(
                 self.X.data, self.X.indices, self.X.indptr, self.y, self.lmbd,
                 L, n_iter)
         else:
-            L = norm(self.X, axis=0) ** 2
             self.w = self.cd(self.X, self.y, self.lmbd, L, n_iter)
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


benchmarks/lasso/solvers/cd.py
```diff
@@ -39,13 +37,12 @@ def set_objective(self, X, y, lmbd):
         self.run(1)
 
     def run(self, n_iter):
+        L = (self.X ** 2).sum(axis =0)
```
agramfort, line 40:
```suggestion
        L = (self.X ** 2).sum(axis=0)
```

benchmarks/nnls/solvers/cd.py
```diff
@@ -34,14 +32,12 @@ def set_objective(self, X, y, fit_intercept=False):
 
     def run(self, n_iter):
         filterwarnings("ignore", category=PerformanceWarning)
-
+        L = (self.X ** 2).sum(axis =0)
```
agramfort, line 35:
```suggestion
        L = (self.X ** 2).sum(axis=0)
```

benchmarks/ols/solvers/cd.py
```diff
@@ -34,14 +32,12 @@ def set_objective(self, X, y, fit_intercept=False):
 
     def run(self, n_iter):
         filterwarnings("ignore", category=PerformanceWarning)
-
+        L = (self.X ** 2).sum(axis =0)
```
agramfort, line 35:
```suggestion
        L = (self.X ** 2).sum(axis=0)
```

josephsalmon: sorry, pep8 now corrected ;)

tomMoral:
Thx @josephsalmon

TomDLT:
Optim question: why is L computed with a norm here ?
When do you need power iteration instead ?

mathurinm:
Those are the Lipschitz constants for coordinate descent solvers, so you can get them in closed form (they are the squared norms of the predictors)

You need power iteration for the operator norm of X, as I think sparse.linalg.norm does not handle the operator norm.

TomDLT:
I was especially confused since the baseline solver was also using `np.linalg.norm(X, ord=2)`.
Today I learned that the "2-norm" is not the same for matrices and vectors (https://numpy.org/doc/1.18/reference/generated/numpy.linalg.norm.html), and that `np.linalg.norm(X, ord=2)` is actually computing the max over the eigenvalues computed with a SVD.

PR #45, opened 2020-05-11
ENH: add r wrapper for glmnet solver

mathurinm:
@tomMoral is it the correct way to specify the python install command


agramfort:
I guess CIs need to be updated

mathurinm:
@tomMoral one failure: `assert solver_class.install_cmd in [None, 'pip', 'bash']`:
I install glmnet with a python script. Should I change the install command to bash and prefix my command with `python -c` ?

benchmarks/lasso/solvers/glmnet.py
```diff
@@ -0,0 +1,43 @@
...
+with safe_import() as solver_import:
+    import rpy2.robjects.packages as rpackages
+    from rpy2 import robjects
+    from rpy2.robjects import numpy2ri
+
+
+class Solver(BaseSolver):
+    name = "glmnet"
+
+    install_cmd = 'python'
```
tomMoral, line 16: Yep this is `bash` and you need to furnish the name of a bash script file stored in `install_scripts`.
The command that is called is `bash install_scripts/{install_script} {env}` and is run with the correct environment activated I think.

This looks quite sub-optimal I agree :)

benchmarks/lasso/solvers/glmnet.py
```diff
@@ -0,0 +1,43 @@
...
+        rpackages.importr('glmnet')
+        glmnet = robjects.r['glmnet']
+        fit_dict = {"lambda.min.ratio": self.lmbd / self.lmbd_max}
+        glmnet_fit = glmnet(self.X, self.y, intercept=False,
+                            standardize=False, thresh=tol, **fit_dict)
+        results = dict(zip(glmnet_fit.names, list(glmnet_fit)))
+        as_matrix = robjects.r['as']
+        coefs = np.array(as_matrix(results["beta"], "matrix"))
+        self.w = coefs[:, -1]
+        # TODO del coefs?
```
tomMoral, line 40: I don't think this is necessary, they should be garbage collected once `run` returns.

mathurinm:
I need rpy2 to be installed before installation with bash (which is somehow a fake bash given it's really python, but I don't mind having it this way).
I have changed the code so that even if cmd is bash, some python packages can be installed before if need be.

However this approach fails and I can't understant why. According to CI, this script failed while it looks reasonable:
`        script     = 'source ./.venv//benchopt_test_env/bin/activate\npip install -qq rpy2'`

benchmarks/lasso/solvers/glmnet.py
```diff
@@ -0,0 +1,37 @@
...
+    requirements_import = ['rpy2']
+
+    def set_objective(self, X, y, lmbd):
+        self.X, self.y, self.lmbd = X, y, lmbd
+        self.lmbd_max = np.max(np.abs(X.T @ y))
+
+    def run(self, tol):
+        numpy2ri.activate()
+        rpackages.importr('glmnet')
+        glmnet = robjects.r['glmnet']
```
tomMoral, line 27: I would put this in `set_objective` as this is doing imports that can take time while not related to the benchmark in itself.

benchopt/base.py
```diff
@@ -132,6 +132,10 @@ def install(cls, env_name=None, force=False):
                     pip_install_in_env(*cls.requirements_install,
                                        env_name=env_name)
                 elif cls.install_cmd == 'bash':
+                    # eventually install other packages needed
+                    if hasattr(cls, "requirements_import"):
```
tomMoral, line 136: This should be `requirements`, `requirements_import` is a hook for when the name of the import is different from the name of the `requirement`.

install_scripts/install_glmnet.sh
```diff
@@ -0,0 +1,4 @@
+python -c 'import rpy2.robjects.packages as rpackages; \
+    utils = rpackages.importr("utils"); \
+    utils.chooseCRANmirror(ind=1); \
+    utils.install_packages("glmnet", dependencies=True);'
```
tomMoral, line 4: nice

agramfort:
@mathurinm how does a meaningful benchmark look like with this?

mathurinm:
So, how would you do it ? Not by installing r in bash before rpy2 ?
(Actually an issue since the order is r, rpy2, glmnet in r, corresponding
to bash, python, bash)

Le mar. 12 mai 2020 à 17:42, Thomas Moreau <notifications@github.com> a
écrit :

tomMoral:
> So, how would you do it ? Not by installing r in bash before rpy2 ? (Actually an issue since the order is r, rpy2, glmnet in r, corresponding to bash, python, bash)

You could do everything in bash by just calling `pip install r2py` instead of using deps.

However, I am reluctant to installing stuff on the user system and not in an env. I would rather use conda for this, to have isolated env that works on all platforms with binaries.

mathurinm:
From what I see online, `conda install r-base` should install `R`. 

But I am not unsure, are we using conda at the moment ? I see pip everywhere except in CI

benchmarks/lasso/solvers/glmnet.py
```diff
@@ -0,0 +1,35 @@
...
+    install_cmd = 'conda'
+    requirements = ['r-base', 'rpy2', 'r-glmnet']
+
+    def set_objective(self, X, y, lmbd):
+        self.X, self.y, self.lmbd = X, y, lmbd
+        self.lmbd_max = np.max(np.abs(X.T @ y))
+
+    def run(self, tol):
+        numpy2ri.activate()
+        rpackages.importr('glmnet')
```
agramfort, line 24: rpackages is not defined

mathurinm:
Ok so it was surprisingly easy to do now that we have conda.

The tests pass on linux though `rpackages` is not defined; we don't test that the solvers run @tomMoral, is it on purpose bc it would take a lot of time ?

tomMoral:
Yes this is mostly that and also that we need to install all the solvers to make this run.
As before, we were testing the uninstall after the install, it was very inconvenient but maybe we should now try to make the solver run (and not skip them).

benchmarks/lasso/solvers/glmnet.py
```diff
@@ -0,0 +1,37 @@
...
+    name = "glmnet"
+
+    install_cmd = 'conda'
+    requirements = ['r-base', 'rpy2', 'r-glmnet']
+
+    def set_objective(self, X, y, lmbd):
+        self.X, self.y, self.lmbd = X, y, lmbd
+        self.lmbd_max = np.max(np.abs(X.T @ y))
+        numpy2ri.activate()
+        import rpy2.robjects.packages as rpackages
```
tomMoral, line 22: A nitpick but would be good to move this in the `import_ctx`

tomMoral (approved):
LGTM!!

agramfort:
plots do make sense for you?

>

josephsalmon:
> plots do make sense for you?
> […](#)

which plot? could you attach what you get here?

mathurinm:
@josephsalmon 
image

and there is an error for scipy sparse matrices. I am looking into it

mathurinm:
The sampling strategy for `tol` yields values which are too high for glmnet to improve the objective, causing premature stop of the runner because it sees no improvement.

Reproduce with 
```
import numpy as np
from numpy.linalg import norm

from benchmarks.lasso.solvers.glmnet import Solver
from benchmarks.lasso.datasets.boston import Dataset

sol = Solver()
dataset = Dataset()

p, data = dataset.get_data()
X, y = data['X'], data['y']


lmbd_max = np.max(np.abs(X.T @ y))

lmbd = lmbd_max / 2
sol.set_objective(X, y, lmbd)


for tol in np.geomspace(1, 1e-7, num=7):
    sol.run(tol=tol)
    w = sol.get_result()
    print(norm(y - X @ w) ** 2 / 2 + lmbd * norm(w, ord=1))
```

output:
```
118738.32751256222
118738.32751256222
118738.32751256222
118738.32751256222
118734.00209545472
118733.22737053409
118733.10199692092
```

gdurif:
Hi, I have a question regarding the management of the regularization parameter.

glmnet tries a range of values (set by `lambda.min.ratio`) instead of just a single one and returns the corresponding vector of estimates.

I guess it is different from other methods where you can properly choose a lambda. However, I cannot find where the lambda grid is managed in benchOpt?

Thanks

mathurinm:
@gdurif it is managed here: https://github.com/benchopt/benchOpt/pull/45/files#diff-33ad2912da3474749cfba930bae18b09R29   with 
```fit_dict = {"lambda.min.ratio": self.lmbd / self.lmbd_max}```
then we get back only the coefficients corresponding to the last lambda, with 
```        
coefs = np.array(as_matrix(results["beta"], "matrix"))
self.w = coefs[:, -1]
```

**Comment by josephsalmon** [comment]
@mathurinm : don't you think it would be fairer to use: nlambda = 1 (or 2)?
`nlambda is the number of λ values in the sequence. Default is 100.`
so I presume 100 lambdas are computed in the process, right?

see: https://web.stanford.edu/~hastie/glmnet/glmnet_alpha.html

**Comment by gdurif** [comment]
@mathurinm ok thanks, why choosing the last one?

**Comment by mathurinm** [comment]
@gdurif because it's the one we're interested in from an optimization point of view, we want to solve a single problem fit lambda so we solve a path of 100 values from lambda max to lambda, and pick the coefs for the last problem

@josephsalmon actually since warm start is used it should help to use a grid. When a single lambda is fitted with glmnet.fit, I believe this is the policy used by glmnet. (which amounts to working sets in spirit)

**Comment by gdurif** [comment]
@josephsalmon you're right. And glmnet code is optimized to more efficient when using the default lambda vector instead of giving a user-defined lambda value.

**Comment by josephsalmon** [comment]
ok sounds fair then, let's try that way for a start.

my experience is none with glmnet, so my insights from other solvers (costly to compute 100 values even if warm start helps) will not help here. But for large X matrix, there might a big overhead though.

**Comment by gdurif** [comment]
@mathurinm ok I understand.

**Comment by gdurif** [comment]
@mathurinm glmnet can manage sparse matrix from R package `Matrix`:

```R
# sparse matrix
library(glmnet)
library(Matrix)

x = matrix(rnorm(100 * 20), 100, 20)
y = rnorm(100)

sparse_x = Matrix(x * (abs(x) < 0.5), sparse = TRUE)
fit1 = glmnet(sparse_x, y)
print(fit1)
```

There should be a way to convert `scipy` sparse matrix to `Matrix` sparse matrix, I will look into it.

mathurinm:
@gdurif yes, I am the author of the question :smile: 
I tried locally 
```
        self.X, self.y, self.lmbd = X, y, lmbd
        self.lmbd_max = np.max(np.abs(X.T @ y))

        if sparse.issparse(X):
            # py2ri does not handle sparse
            X = X.tocoo()
            r_Matrix = importr("Matrix")
            self.X = r_Matrix.sparseMatrix(
                i=robjects.IntVector(X.row + 1),
                j=robjects.IntVector(X.col + 1),
                x=robjects.FloatVector(X.data),
                dims=robjects.IntVector(X.shape))
```
but I ended up with high memory consumption on my laptop so I decided to address it in a subsequent PR. We have a mechanism to state that a solver does not support sparse X in the meantime.

gdurif:
@mathurinm ok, good question then :sweat_smile: I'll try to dig into this.

Regarding the number of lamdas, I guess we could find a hack based on the default values for `nlamda` (i.e. 100) and `lambda.min.ratio` which is, according to the doc:

> Smallest value for lambda, as a fraction of lambda.max, the (data derived) entry value (i.e. the smallest value for which all coefficients are zero). The default depends on the sample size nobs relative to the number of variables nvars. If nobs > nvars, the default is 0.0001, close to zero. If nobs < nvars, the default is 0.01. A very small value of lambda.min.ratio will lead to a saturated fit in the nobs < nvars case.

We can deduce the number of lambdas that would correspond to the same grid step for a chosen lambda. I'll try it.

agramfort:
We should maybe protect rpy2 imports to avoid a hard dependency on R

tomMoral:
They are protected no ? The helper need to be used in the `import_ctx`.

agramfort:
ok we're good then

>

benchmarks/lasso/solvers/r_ista.R
```diff
@@ -0,0 +1,40 @@
+##' Functions used in ISTA algrithm
```
agramfort, line 1:
```suggestion
##' Functions used in ISTA algorithm
```

benchmarks/lasso/solvers/r_ista.R
```diff
@@ -0,0 +1,40 @@
...
+##'
+##' @title Functions used in ISTA algrithm
```
agramfort, line 3: idem

benchopt/utils/r_helpers.py
```diff
@@ -21,3 +25,11 @@ def import_rpackages(*packages):
...
         except PackageNotInstalledError:
             raise ImportError(f"R package '{pkg}' is not installed")
+
+
+def import_func_from_r_file(filename):
+    import rpy2.robjects as robjects
+    r_source = robjects.r['source']
+    r_source(filename)
+
+    # print(robjects.r['sessionInfo']())
```
agramfort, line 35: cleanup?

benchmarks/lasso/solvers/r_ista.R
```diff
@@ -0,0 +1,40 @@
...
+}
+
+
+# Main algorithm
+ISTA <- function(X, Y, lambda, n_iter = 10000) {
+    # --------- Initialize parameter ---------
+    p <- ncol(X)
+    parameters <- numeric(p)
+    step_size <- norm(X, type="2")
+    # step_size <- 0.0005
```
agramfort, line 30:
```suggestion
```

benchmarks/lasso/solvers/r_proximal_gradient.py
```diff
@@ -0,0 +1,47 @@
...
+
+    install_cmd = 'conda'
+    requirements = ['r-base', 'r-matrix', 'rpy2']
+    stop_strategy = 'iteration'
+    support_sparse = False
+
+    def set_objective(self, X, y, lmbd):
+        self.X, self.y, self.lmbd = X, y, lmbd
+        self.lmbd_max = np.max(np.abs(X.T @ y))
+        self.r_ista = robjects.r['ISTA']
```
agramfort, line 31: ISTA -> proximal_gradient

benchmarks/lasso/solvers/r_proximal_gradient.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def run(self, n_iter):
+
+        # There is an issue in loading Lapack library with rpy2 so
+        # we cannot compute the SVD in R for now. We compute it using
+        # numpy but this should be fixed at some point. See issue #52
+        step_size = 1 / np.linalg.norm(self.X, ord=2) ** 2
+        coefs = self.r_ista(
```
agramfort, line 39:
```suggestion
        coefs = self.r_proximal_gradient(
```

agramfort (commented):
see how to remove ISTA naming from var / function names

benchmarks/lasso/solvers/r_proximal_gradient.R
```diff
@@ -0,0 +1,37 @@
...
+
+        return(result0)
+    }
+    result <- apply(X, 1, s, lambda0 = lambda)
+    return(result)
+}
+
+
+# Main algorithm
+ISTA <- function(X, Y, lambda, step_size, n_iter) {
```
agramfort, line 25: No Ista please

agramfort (approved):

agramfort:
one more down !

PR #46, opened 2020-05-13
[ENH] move from pip to conda

mathurinm:
Although heavier, conda allows us to install more stuff (e.g. R) in envs and not system wide. We needed it for #45

### DIFF
--- a/.azure_pipeline.yml
+++ b/.azure_pipeline.yml
@@ -71,13 +71,11 @@ jobs:
     displayName: 'Override Git bash shell for Windows'
     condition: eq(variables['Agent.OS'], 'Windows_NT')
 
-  - powershell: conda init $SHELL
+  - powershell: conda init powershell
     displayName: Conda init
     condition: eq(variables['Agent.OS'], 'Windows_NT')
 
-  - bash: |
-      conda init bash
-      exec bash
+  - bash: conda init bash
     displayName: Conda init
     condition: ne(variables['Agent.OS'], 'Windows_NT')
 
--- a/.azure_pipeline.yml
+++ b/.azure_pipeline.yml
@@ -76,8 +76,8 @@ jobs:
     condition: eq(variables['Agent.OS'], 'Windows_NT')
 
   - bash: |
-      conda init $SHELL
-      ./env.sh
+      conda init bash
+      exec bash
     displayName: Conda init
     condition: ne(variables['Agent.OS'], 'Windows_NT')
 
--- a/.azure_pipeline.yml
+++ b/.azure_pipeline.yml
@@ -59,7 +59,7 @@ jobs:
 #         PACKAGER: 'pip'
 
   variables:
-    VIRTUALENV: 'testvenv'
+    CONDAENV: 'testcondaenv'
     JUNITXML: 'test-data.xml'
 
   pool:
--- a/.azure_pipeline.yml
+++ b/.azure_pipeline.yml
@@ -71,13 +71,16 @@ jobs:
     displayName: 'Override Git bash shell for Windows'
     condition: eq(variables['Agent.OS'], 'Windows_NT')
 
-  - powershell: conda init powershell
-    displayName: Conda init
+  - powershell: Write-Host "##vso[task.prependpath]$env:CONDA\Scripts"
+    displayName: Add conda to PATH
     condition: eq(variables['Agent.OS'], 'Windows_NT')
 
-  - bash: conda init bash
-    displayName: Conda init
-    condition: ne(variables['Agent.OS'], 'Windows_NT')
+  - bash: |
+      echo "##vso[task.prependpath]$CONDA/bin"
+      sudo chown -R $USER $CONDA
+    displayName: Add conda to PATH
+    condition: and(eq(variables['PACKAGER'], 'conda'),
+                   ne(variables['Agent.OS'], 'Windows_NT'))
 
   - bash: bash continuous_integration/install.sh
     displayName: 'Install benchopt and its dependencies'
--- a/.azure_pipeline.yml
+++ b/.azure_pipeline.yml
@@ -78,8 +78,6 @@ jobs:
   - bash: |
       conda init $SHELL
       ./env.sh
-      sudo ./svc.sh stop
-      sudo ./svc.sh start
     displayName: Conda init
     condition: ne(variables['Agent.OS'], 'Windows_NT')
 
--- a/.azure_pipeline.yml
+++ b/.azure_pipeline.yml
@@ -75,7 +75,11 @@ jobs:
     displayName: Conda init
     condition: eq(variables['Agent.OS'], 'Windows_NT')
 
-  - bash: conda init $SHELL
+  - bash: |
+      conda init $SHELL
+      ./env.sh
+      sudo ./svc.sh stop
+      sudo ./svc.sh start
     displayName: Conda init
     condition: ne(variables['Agent.OS'], 'Windows_NT')
 
--- a/.azure_pipeline.yml
+++ b/.azure_pipeline.yml
@@ -21,25 +21,18 @@ jobs:
   strategy:
     matrix:
 
-      # linux_py36_ubuntu:
-      #   imageName: 'ubuntu-latest'
-      #   PACKAGER: 'ubuntu'
-      #   VERSION_PYTHON: '3.6'
       linux_py36_conda:
         imageName: 'ubuntu-latest'
-        PACKAGER: 'conda'
         VERSION_PYTHON: '3.6'
         COVERAGE: 'true'
       linux_pylatest_conda:
         imageName: 'ubuntu-latest'
-        PACKAGER: 'conda'
         VERSION_PYTHON: '*'
 
       # MacOS builds
       macOS_pylatest_conda:
         imageName: 'macOS-10.14'
         VERSION_PYTHON: '*'
-        PACKAGER: 'conda'
 
 #       win_py38_pip:
 #         imageName: 'vs2017-win2016'
@@ -79,8 +72,7 @@ jobs:
       echo "##vso[task.prependpath]$CONDA/bin"
       sudo chown -R $USER $CONDA
     displayName: Add conda to PATH
-    condition: and(eq(variables['PACKAGER'], 'conda'),
-                   ne(variables['Agent.OS'], 'Windows_NT'))
+    condition: ne(variables['Agent.OS'], 'Windows_NT')
 
   - bash: bash continuous_integration/install.sh
     displayName: 'Install benchopt and its dependencies'
--- a/.azure_pipeline.yml
+++ b/.azure_pipeline.yml
@@ -21,10 +21,10 @@ jobs:
   strategy:
     matrix:
 
-      linux_py36_ubuntu:
-        imageName: 'ubuntu-latest'
-        PACKAGER: 'ubuntu'
-        VERSION_PYTHON: '3.6'
+      # linux_py36_ubuntu:
+      #   imageName: 'ubuntu-latest'
+      #   PACKAGER: 'ubuntu'
+      #   VERSION_PYTHON: '3.6'
       linux_py36_conda:
         imageName: 'ubuntu-latest'
         PACKAGER: 'conda'
@@ -71,16 +71,13 @@ jobs:
     displayName: 'Override Git bash shell for Windows'
     condition: eq(variables['Agent.OS'], 'Windows_NT')
 
-  - powershell: Write-Host "##vso[task.prependpath]$env:CONDA\Scripts"
-    displayName: Add conda to PATH
+  - powershell: conda init $SHELL
+    displayName: Conda init
     condition: eq(variables['Agent.OS'], 'Windows_NT')
 
-  - bash: |
-      echo "##vso[task.prependpath]$CONDA/bin"
-      sudo chown -R $USER $CONDA
-    displayName: Add conda to PATH
-    condition: and(eq(variables['PACKAGER'], 'conda'),
-                   ne(variables['Agent.OS'], 'Windows_NT'))
+  - bash: conda init $SHELL
+    displayName: Conda init
+    condition: ne(variables['Agent.OS'], 'Windows_NT')
 
   - bash: bash continuous_integration/install.sh
     displayName: 'Install benchopt and its dependencies'
--- a/benchmarks/lasso/solvers/celer.py
+++ b/benchmarks/lasso/solvers/celer.py
@@ -14,8 +14,7 @@ class Solver(BaseSolver):
     sampling_strategy = 'iteration'
 
     install_cmd = 'conda'
-    requirements = ['pip:celer']
-    requirements_import = ['celer']
+    requirements = ['celer']
     requirements_install = ['pip:git+https://github.com/mathurinm/celer.git']
 
     def set_objective(self, X, y, lmbd):
--- a/benchmarks/lasso/solvers/celer.py
+++ b/benchmarks/lasso/solvers/celer.py
@@ -14,8 +14,9 @@ class Solver(BaseSolver):
     sampling_strategy = 'iteration'
 
     install_cmd = 'conda'
-    requirements = ['celer']
-    requirements_install = ['git+https://github.com/mathurinm/celer.git']
+    requirements = ['pip:celer']
+    requirements_import = ['celer']
+    requirements_install = ['pip:git+https://github.com/mathurinm/celer.git']
 
     def set_objective(self, X, y, lmbd):
         self.X, self.y, self.lmbd = X, y, lmbd
--- a/benchmarks/lasso/solvers/celer.py
+++ b/benchmarks/lasso/solvers/celer.py
@@ -13,7 +13,7 @@ class Solver(BaseSolver):
     name = 'Celer'
     sampling_strategy = 'iteration'
 
-    install_cmd = 'pip'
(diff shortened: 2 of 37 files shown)


mathurinm:
@tomMoral I missed the place where conda is installed, however given the current tests failures I was thinking of having one `conda init $SHELL` right after conda installation.

Should it replace:
```  - powershell: Write-Host "##vso[task.prependpath]$env:CONDA\Scripts"
    displayName: Add conda to PATH
    condition: eq(variables['Agent.OS'], 'Windows_NT')

  - bash: |
      echo "##vso[task.prependpath]$CONDA/bin"
      sudo chown -R $USER $CONDA
    displayName: Add conda to PATH
    condition: and(eq(variables['PACKAGER'], 'conda'),
                   ne(variables['Agent.OS'], 'Windows_NT'))
```
?

Also, is it ok to remove packager=ubuntu?

mathurinm:
The test fail locally, `conda activate` fail when executed in subprocess @tomMoral 
Locally, error messages look like:
```
E               RuntimeError: Failed to conda install packages ['numpy', 'cython']
E               Error:
E               CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'.
E               To initialize your shell, run
E               
E                   $ conda init <SHELL_NAME>
E               
E               Currently supported shells are:
E                 - bash
E                 - fish
E                 - tcsh
E                 - xonsh
E                 - zsh
E                 - powershell
E               
E               See 'conda init --help' for more information and options.
E               
E               IMPORTANT: You may need to close and restart your shell after running 'conda init'.

benchopt/util.py:73: RuntimeError
```

benchopt/config.py
```diff
@@ -29,15 +29,15 @@ def get_global_setting(name):
     assert name in DEFAULT_GLOBAL, f"Unknown config key {name}"
 
     # TODO: get the correct type from DEFAULT_GLOBAL
-    venv_name = f"BENCHO_{name.upper()}"
+    condaenv_name = f"BENCHO_{name.upper()}"
```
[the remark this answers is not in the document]
mathurinm: what is var?

mathurinm:
In the end I managed to make it work locally, with the exception that celer, lightning etc are installed from pip, which conda does not handle, so we still need pip for those.

For the install, we can filter with: `install_with_pip = [req for req in cls.requirement_install if req.startswith("git+")]`, and pip install those (the install happens in the conda env)

For the uninstall, it's a bit heavier since names needed for the uninstall like sklearn-contrib-lightning do not start with git+.

What do you think ?

agramfort:
what do you mean conda does not support pip install?

>

mathurinm:
sorry, meant install from github such as `pip install git+https://github.com/tommoral/blitzl1.git@FIX_setup_deps`; you can't do `conda install git+https://github.com/tommoral/blitzl1.git@FIX_setup_deps` so you need to use pip (inside the conda env) for these.

agramfort:
yes use pip inside the conda env for this

>

mathurinm:
Tests are almost green :)
From my experience with this PR I now wonder if we should not go more often with a `.sh` install script for packages, even if it mostly contains conda commands.
- an exemple is conda vs pip for packages installed from github. I am not sure the syntax I used with `pip:` is easy to understand for a beginner. With a bash script and a single conda env, the bash script would simply be `conda activate benchopt_env; pip install git+https...`
- cvxpy needs to be downloaded from conda-forge, so we need to `conda config --append channels conda-forge` before, and remove it after. This is a bit complicated. With a bash script this can be done explicitely and simply

It would feel more flexible to have a bash script for every package, as soon as installation is not all vanilla conda.
WDYT @tomMoral ?

benchopt/util.py
```diff
@@ -103,42 +92,50 @@ def _run_bash_in_env(script, env_name=None, raise_on_error=None,
...
 
 def bash_install_in_env(script, env_name=None):
     """Run a bash install script in the given environment"""
     if env_name is None and not ALLOW_INSTALL:
-        raise ValueError("Trying to install solver not in a virtualenv. "
+        raise ValueError("Trying to install solver not in a conda env. "
                          "To allow this, set BENCHO_ALLOW_INSTALL=True.")
-    env = "$VIRTUAL_ENV" if env_name is not None else "$HOME/.local/"
+    env = env_name if env_name is not None else "base"
+    # TODO correct idea to use base?
```
mathurinm, line 138: @tomMoral I am not 100 % sure here

continuous_integration/install.sh
```diff
@@ -20,23 +20,27 @@ fi
...
+make_conda $TO_INSTALL
 
-elif [[ "$PACKAGER" == "ubuntu" ]]; then
-    sudo apt-get update
-    sudo apt-get install python3-venv python3-virtualenv python3-numpy cython
-    python3 -m virtualenv --system-site-packages --python=python3 $VIRTUALENV
-    source $VIRTUALENV/bin/activate
-    python -m pip install -r dev-requirements.txt
-fi
+# elif [[ "$PACKAGER" == "ubuntu" ]]; then
```
mathurinm, line 37: @tomMoral if we go with conda, we still have to keep this ?

mathurinm:
@agramfort ready for review

.azure_pipeline.yml
```diff
@@ -21,10 +21,10 @@ jobs:
...
     matrix:
 
-      linux_py36_ubuntu:
-        imageName: 'ubuntu-latest'
-        PACKAGER: 'ubuntu'
-        VERSION_PYTHON: '3.6'
+      # linux_py36_ubuntu:
+      #   imageName: 'ubuntu-latest'
+      #   PACKAGER: 'ubuntu'
+      #   VERSION_PYTHON: '3.6'
```
agramfort, line 27: to cleanup @tomMoral ?

benchmarks/logreg/solvers/lightning.py
```diff
@@ -9,11 +9,11 @@
...
     name = 'Lightning'
 
-    install_cmd = 'pip'
-    requirements = ['sklearn-contrib-lightning']
+    install_cmd = 'conda'
+    requirements = ['pip:sklearn-contrib-lightning']
     requirements_import = ['lightning']
     requirements_install = [
-        'git+https://github.com/scikit-learn-contrib/lightning.git'
+        'pip:git+https://github.com/scikit-learn-contrib/lightning.git'
```
agramfort, line 16: to be honest I don't really understand why you need 3 entries for one package. Why a single line is not enough? As if you would pass an env yml or req .txt file.

benchopt/config.py
```diff
@@ -10,10 +10,10 @@
 
 
 DEFAULT_GLOBAL = {
-    'debug': False,
+    'debug': True,
     'allow_install': False,
     'raise_install_error': False,
-    'venv_dir': './.venv/',
+    # 'venv_dir': './.venv/',
```
agramfort, line 16: to cleanup

benchopt/tests/test_benchmarks.py
```diff
@@ -29,19 +29,19 @@ def class_ids(parameter):
...
+# Setup and clean a test env to install all the solvers and check
 # that they are correctly configured
 
 TEST_ENV_NAME = "benchopt_test_env"
 
 
 def setup_module(module):
     print("create env")
-    create_venv(TEST_ENV_NAME, recreate=True)
+    create_condaenv(TEST_ENV_NAME, recreate=True)
```
agramfort, line 40: create_condaenv -> create_conda_env

benchopt/tests/test_benchmarks.py
```diff
@@ -7,7 +7,7 @@
 from benchopt.util import get_benchmark_objective
 from benchopt.util import list_benchmark_solvers
 from benchopt.util import list_benchmark_datasets
-from benchopt.util import create_venv, delete_venv
+from benchopt.util import create_condaenv, delete_condaenv
```
agramfort, line 10:
```suggestion
from benchopt.util import create_conda_env, delete_conda_env
```

benchopt/tests/test_benchmarks.py
```diff
@@ -201,3 +199,14 @@ def test_solver(benchmark_name, solver_class):
...
+
+if __name__ == "__main__":
+    for idx in range(8):
+        if idx == 5:  # cyanure, skip
+            continue
+        bench, solver_class = BENCH_AND_SOLVERS[idx]
+        print(bench, solver_class)
+        create_condaenv(TEST_ENV_NAME, recreate=True)
+        assert solver_class.install(env_name=TEST_ENV_NAME, force=True)
+        assert solver_class.is_installed(env_name=TEST_ENV_NAME)
```
agramfort, line 212: cleanup?

continuous_integration/install.sh
```diff
@@ -20,23 +20,24 @@ fi
...
+#     sudo apt-get update
+#     sudo apt-get install  python3-numpy cython
+#     python3 -m virtualenv --system-site-packages --python=python3 $CONDAENV
+#     conda activate $CONDAENV
+#     python -m pip install -r dev-requirements.txt
+# fi
```
agramfort, line 40: to cleanup

continuous_integration/test_script.sh
```diff
@@ -2,11 +2,12 @@
...
-    source activate $VIRTUALENV
-elif [[ "$PACKAGER" == "ubuntu" ]]; then
-    source $VIRTUALENV/bin/activate
-fi
+# if [[ "$PACKAGER" == "conda" ]]; then
+# . $CONDA_PREFIX/etc/profile.d/conda.sh
+. activate $CONDAENV
+# elif [[ "$PACKAGER" == "ubuntu" ]]; then
+    # source $CONDAENV/bin/activate
+# fi
```
agramfort, line 10: cleanup

benchopt/base.py
```diff
@@ -103,59 +102,42 @@ def is_installed(cls, env_name=None):
...
-        if force:
-            cls.uninstall(env_name=env_name)
+        print(f"Installing {cls.name} in {env_name}:...",
+              end='', flush=True)
+        try:
+            if cls.install_cmd == 'conda':
+                conda_install_in_env(*cls.requirements_install,
+                                     env_name=env_name, force=force)
+            elif cls.install_cmd == 'bash':
+                bash_install_in_env(cls.install_script, env_name=env_name)
```
agramfort, line 127: is it really bash you need here? or just a shell? any chance windows can supported for at least python solvers?

benchmarks/logreg/solvers/lightning.py
tomMoral: Part of it is legacy and part is because package import name is not always the same as the name on Pypi. We need the import name to do `python -c "import import_name"` to check if the install ran. But maybe a solution would be to rely on `safe_import` to check install with `python -c "from benchmarks.benchname.solvers import solvername; solver_import = getattr(solvername, 'solver_import', None); assert solver_import is None or not solver_import.failed_import"`.

benchmarks/lasso/solvers/celer.py
```diff
@@ -13,9 +13,10 @@ class Solver(BaseSolver):
     name = 'Celer'
     sampling_strategy = 'iteration'
 
-    install_cmd = 'pip'
-    requirements = ['celer']
-    requirements_install = ['git+https://github.com/mathurinm/celer.git']
+    install_cmd = 'conda'
+    requirements = ['pip:celer']
```
tomMoral, line 17: Do you really need to put `pip:` twice and not just in `requirements_install`?

benchopt/base.py
tomMoral: We just need a shell but most script here will require compilers I guess so not sure if there is much hope to make this work for windows. That is why using conda for all would be a nice solution.

benchopt/base.py
```diff
@@ -103,59 +102,42 @@ def is_installed(cls, env_name=None):
...
         -------
         is_installed: bool
             True if the class is correctly installed in the environment.
         """
-        # uninstall the class that requires a force reinstall
-        if force:
-            cls.uninstall(env_name=env_name)
+        print(f"Installing {cls.name} in {env_name}:...",
+              end='', flush=True)
+        try:
```
tomMoral, line 122: You remove the skip when install. This would avoid the long dependency resolution + update when the package is already installed so I would rather continue to skip if it is already installed no?

benchopt/config.py
tomMoral: yes

benchopt/config.py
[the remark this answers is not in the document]
tomMoral: This is an environment variable. This allow you to overwritte settings by putting `BENCHO_RAISE_INSTALL_ERROR=1` before you script.

benchopt/tests/test_benchmarks.py
tomMoral: +1

benchopt/util.py
```diff
@@ -103,42 +92,50 @@ def _run_bash_in_env(script, env_name=None, raise_on_error=None,
         Exit code of the script
     """
     if env_name is not None:
-        env_dir = f"{VENV_DIR}/{env_name}"
-
-        script = f"source {env_dir}/bin/activate\n{script}"
+        script = (f". activate {env_name} && {script}")
```
tomMoral, line 95: Don't use the `&&` syntax here. The script can be multi-lines so it won't prevent for running the next lines.
To protect this, we set L.48 in `_run_in_bash` the `set -e` option that kills the script with the first error and return the exit code.

benchopt/util.py
tomMoral: Can't we use the current `env` in this case?

benchopt/util.py
```diff
@@ -290,28 +287,28 @@ def is_included(name, include_patterns=None):
     return False
 
 
-def create_venv(env_name, recreate=False):
-    """Create a virtual env with name env_name and install basic utilities"""
+def create_condaenv(env_name, recreate=False):
+    """Create a conda env with name env_name and install basic utilities"""
 
-    env_dir = f"{VENV_DIR}/{env_name}"
+    force = " -y" * recreate
```
tomMoral, line 293: Not sure if I like this syntax. Maybe putting `" y" if recreate else ""` is more explicit/maintanable.

benchopt/util.py
```diff
@@ -290,28 +287,28 @@ def is_included(name, include_patterns=None):
...
-        print(f"Creating venv {env_name}:...", end='', flush=True)
-        venv.create(env_dir, with_pip=True)
-        # Install benchopt as well as packages used as utilities to install
-        # other packages. The install of benchopt is done with the -e flag
-        # to ease development process
-        # XXX: add a flag to enable/disable develop install?
-        pip_install_in_env("numpy", "cython", "-e .", env_name=env_name)
-        print(" done")
+    print(f"Creating conda env {env_name}:...", end='', flush=True)
+    subprocess.run(f"conda create{force} -n {env_name}", shell=True)
```
tomMoral, line 296: As we don't care about extra spaces, I would rather have `conda create {force} -n ...`, it is less error prone.

tomMoral: You should catch the exit code and raise a proper error if the command fail.

continuous_integration/install.sh
tomMoral: As we rely on conda for everything, it is true it makes less sense to split so I am in favor of removing this

benchopt/base.py
mathurinm: I did so because conda or pip will skip install if requirement is already satisfied. Do you think it is more explicit to keep the skip ? It does not cost much indeed.

benchmarks/logreg/solvers/lightning.py
agramfort: yes I think we could rely on the safe_import in solver file. It would simplify things

benchopt/util.py
mathurinm: Everytime we run subprocess.run() ? Isn't this too heavy ?

benchmarks/logreg/solvers/lightning.py
mathurinm: Now we have only two fields, requirements and requirements_install

benchopt/util.py
```diff
@@ -290,10 +290,10 @@ def is_included(name, include_patterns=None):
 def create_conda_env(env_name, recreate=False):
     """Create a conda env with name env_name and install basic utilities"""
 
-    force = " -y" * recreate
+    force = " -y" if recreate else ""
```
tomMoral, line 293: As you added the space in l.296, you can remove this one :)

benchopt/config.py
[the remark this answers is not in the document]
mathurinm: @tomMoral  Do you mean that the variable name should be `env_var_name` instead of `condaenv_name`?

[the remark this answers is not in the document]
tomMoral: exactly!

benchopt/base.py
tomMoral: As it is probably just shell, I renamed this option `shell`.

Note that I also install `conda` compilers so maybe ther is a way to compile with conda cross platform but not sure how this would work (we should maybe ask jeremie about this).

tomMoral:
But red... :( 
https://dev.azure.com/benchopt/benchopt/_build/results?buildId=337&view=results

tomMoral:
Enfer! I though I was done but `cyanure` is tripping the PR... 
image

I guess I will skip the test for this one and add an issue as it is not central to this PR.

tomMoral:
Its green!!!

PR #47, opened 2020-07-09
ENH add Logreg-L2 benchmark

agramfort:
Screenshot 2020-07-09 at 19 24 41

for front page :)


agramfort:
good to go from my end. Next is to integrate this to the the front page.

tomMoral:
There is something weird going on with LBFGS, it seems it does not converge?
Any idea why?

PR #48, opened 2020-07-10
MRG add bench to frontpage

agramfort:


agramfort:
good to go from my end

https://130-227720375-gh.circle-artifacts.com/0/dev/index.html

PR #50, opened 2020-07-30
POC julia solver

agramfort:
ping @josephsalmon @tomMoral @ngazagna @gowerrobert

just a quick hack to show the way.

please take over and let me know if you see any obvious bug


benchmarks/lasso/solvers/julia_pgd.py
```diff
@@ -0,0 +1,37 @@
+import numpy as np
+import pandas as pd
+from benchopt.base import CommandLineSolver
+from benchopt.util import safe_import_context, import_shell_cmd
+from scipy.io import savemat, loadmat
+
+with safe_import_context() as import_ctx:
+    import h5py
+    train_cmd = import_shell_cmd('julia')
```
gowerrobert, line 9: What about using this PyJulia package:

https://github.com/JuliaPy/pyjulia

then you could call this lasso function directly. Something like

python3 -m pip install julia    # install PyJulia
$ python3
>>> import julia
>>> julia.install()               # install PyCall.jl etc.
>>> from lasso_pdg import lasso        # short demo
>>> lasso(data_fname, model_fname, lambda, n_iter)

agramfort:
I just tried:

from lasso_pgd import lasso
ModuleNotFoundError: No module named 'lasso_pgd'

can you send me folder with a script that works for you?

benchmarks/lasso/solvers/julia_pgd.py
tomMoral: I think this is a good idea, we can even avoid the need to pass the data through files as pyjulia support passing numpy arrays.
The issue in this snippet was that you need to do something a bit different to get the lasso function:
```python
    import julia
    julia.install()
    # configure the julia runtime
    jl = julia.Julia(
        compiled_modules=False
    )
	solve_lasso = jl.include("benchmarks/lasso/solvers/lasso_pgd.jl")
	solve_lasso(X, y, lmbd, n_iter)
```

agramfort:
thx @tomMoral 

that's a big step forward !

PR #54, opened 2020-09-04
FIX execute conda activate from script

TomDLT:
For some reason I could not activate the conda environment from a script.
This fixes the issue, cf https://github.com/conda/conda/issues/7980.

I also added a Makefile recipe to clean conda environment.


tomMoral (approved):
This has been bothering me for a long time! thanks @TomDLT  for the solution !!

This looks way cleaner and should solve several problems.

PR #61, opened 2020-09-11
FIX APGD implementation

### LINKED ISSUE #60: Shouldn't you have two sequences of variables for the accelerated version?
For instance this line only has variables _w_

https://github.com/benchopt/benchOpt/blob/01acb156bd76708b01a30c3de8918609f1cde3ff/benchmarks/lasso/solvers/python_pgd.py#L31

but I think you need two sequences. Something like this (note the additional variables in x)

       L = np.linalg.norm(self.X, ord=2) ** 2
       n_features = self.X.shape[1]
       w = np.zeros(n_features)
       x = np.zeros(n_features)
       t_new = 1

        for _ in range(n_iter):
            if self.use_acceleration:
                t_old = t_new
                t_new = (1 + np.sqrt(1 + 4 * t_old ** 2)) / 2
            grad = self.X.T.dot(self.X.dot(w) - self.y)
            x_old = x.copy()
            x -= grad / L
            x = self.st(x, self.lmbd / L)  # Not sure what this step does
            if self.use_acceleration:
                w = x+ (t_old - 1.) / t_new * (x -x_old)
        self.w = w

mathurinm:
Comparison of the 2 implems:
image

closes #60

### DIFF
--- a/benchmarks/lasso/solvers/python_pgd.py
+++ b/benchmarks/lasso/solvers/python_pgd.py
@@ -18,17 +18,20 @@ class Solver(BaseSolver):
 
         n_features = self.X.shape[1]
         w = np.zeros(n_features)
+        z = np.zeros(n_features)
         t_new = 1
         for _ in range(n_iter):
             if self.use_acceleration:
                 t_old = t_new
                 t_new = (1 + np.sqrt(1 + 4 * t_old ** 2)) / 2
                 w_old = w.copy()
-            grad = self.X.T.dot(self.X.dot(w) - self.y)
-            w -= grad / L
+            else:
+                z = w
+            w -= self.X.T @ (self.X @ z - self.y) / L
             w = self.st(w, self.lmbd / L)
+
             if self.use_acceleration:
-                w += (t_old - 1.) / t_new * (w - w_old)
+                z = w + (t_old - 1.) / t_new * (w - w_old)
         self.w = w
 
     def st(self, w, mu):
--- a/benchmarks/lasso/solvers/python_pgd.py
+++ b/benchmarks/lasso/solvers/python_pgd.py
@@ -18,20 +18,22 @@ class Solver(BaseSolver):
 
         n_features = self.X.shape[1]
         w = np.zeros(n_features)
-        z = np.zeros(n_features)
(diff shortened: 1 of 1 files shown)


agramfort (approved):

benchmarks/lasso/solvers/python_pgd.py
```diff
@@ -18,17 +18,20 @@ def run(self, n_iter):
...
         for _ in range(n_iter):
             if self.use_acceleration:
                 t_old = t_new
                 t_new = (1 + np.sqrt(1 + 4 * t_old ** 2)) / 2
                 w_old = w.copy()
-            grad = self.X.T.dot(self.X.dot(w) - self.y)
-            w -= grad / L
+            else:
+                z = w
+            w -= self.X.T @ (self.X @ z - self.y) / L
```
TomDLT, line 30:
```suggestion
            w = z - self.X.T @ (self.X @ z - self.y) / L
```

TomDLT (changes requested):
That explains the poor performances of the old APGD on the Boston dataset:
Figure_1
However, your fix diverges:
Figure_1b
This is because the gradient step should be taken from the auxiliary variable. With my proposed fix:
Figure_2

mathurinm:
Thanks for the thorough review !
It is still weird that pgd outperforms apgd on Boston, don't you think ?

TomDLT:
Rerunning the benchmark with 100 repetitions.
Figure_1
Figure_2
Figure_3

agramfort:
I also get now some:

|----Python-PGD[use_acceleration=False]: done (did not converge)
|----Python-PGD[use_acceleration=True]: done (did not converge)

on simulated or boston.

it seems we have a regression here.

agramfort (approved):
@TomDLT I changed a bit more to avoid an extra copy for PDG without accel.

agramfort:
and forget about my brain fart about regression above. I was using -n instead of -r ...

all good on my side

TomDLT (approved):

tomMoral (approved):

tomMoral:
Thanks everyone!

PR #66, opened 2020-09-22
[DEV] Add julia gd solver for logreg l2

ngazagna:
I added gradient descent for logreg_l2 problem, see files:
- `benchmarks/logreg_l2/solvers/julia_gd.jl`
- `benchmarks/logreg_l2/solvers/julia_gd.py`

And I did check that the loss and the gradient were correctly computed by running
`test_julia/test_logreg_l2_grad.jl`

PS: I will need to reuse functions in `benchmarks/logreg_l2/solvers/logistic.jl` for all the other stochastic methods (SGD, SAG, SAGA, etc), yet I was unable to import this file `julia_gd.jl` using canonical 
```julia
include("./benchmarks/logreg_l2/solvers/logistic.jl")
```
Anyone knows how to do ?


benchmarks/logreg_l2/solvers/logistic.jl
```diff
@@ -0,0 +1,33 @@
+
+# Loss evaluation
+function logistic_loss(X, y, w::Array{Float64})
+    # return sum( log.(1. .+  exp.(-(y .* (X*w)))) ) # other implementation
```
agramfort, line 4: why did you choose one version vs the other? One is faster? more stable?

benchmarks/logreg_l2/solvers/logistic.jl
```diff
@@ -0,0 +1,33 @@
...
+end
+
+# Define function and gradient with mini-batching and l2 regularization
+# f(w) = \frac{1}{n} \sum_{i=1}^n f_i (w)
+#      = \frac{\lambda}{2} \norm{w}_2^2 + \sum_{i=1}^n \left( \phi_i (x_i^{\top} w) \right)
+# where \phi_i (z) = \log(1+\exp(-y_i z))
+
+# B stands for a mini-batch of indices
+# logreg_l2_loss(w, B) = (n_samples / length(B)) * logistic_loss(X[B, :], y[B], w) + (0.5 * lambda * norm(w)^2);
+# logreg_l2_grad(w, B) = (n_samples / length(B)) * logistic_grad(X[B, :], y[B], w) .+ (lambda .* w);
```
agramfort, line 33: missing empty line at the end of file

agramfort (commented):
I think the only julia file we need here is julia_gd.jl 

benchopt has tests (it will test it reaches good objective value on test data)

benchmarks/logreg_l2/solvers/logistic.jl
ngazagna: Thanks for the advice

benchmarks/logreg_l2/solvers/julia_gd.jl
```diff
@@ -0,0 +1,68 @@
...
+
+# Loss evaluation
+function logistic_loss(X, y, w::Array{Float64}) # correct
+    # return sum( log.(1. .+  exp.(-(y .* (X*w)))) ) # other implementation
+    return -sum( log.(logistic_diff_phi(y .* (X*w))) )
+end
+
+# Gradient evaluation
+function logistic_diff_phi(z::Array{Float64}) # correct
+    # Derivative of phi_i is ``more or less" the sigmoid
```
mathurinm, line 17: just wondering, can you clarify what "more or less" means ? it seems to me that you compute exactly that, in 2 different ways depending on the sign of the argument, to avoid precision errors I believe ?

benchmarks/logreg_l2/solvers/logistic.jl
ngazagna: The uncommented one was used by Robert in StochOpt.jl.
The commented one is mine, according to `@btime` command it allocates less memory than the second and seems to be a bit faster. Let's keep the commented one then.

benchmarks/logreg_l2/solvers/julia_gd.jl
ngazagna: I'm sorry, the function is *exactly* the sigmoid function. The "more or less" was referring to the name of the function itsfelf. Let 
 
be the i-th loss, then its derivative is 
 
To make everything clearer I'll rename this function as "sigmoid".

And you are correct, the sigmoid is computed differently based on the sign of the input to avoid numerical overflow as explained in this [post](https://medium.com/@shaktiwadekar/how-to-avoid-numerical-overflow-in-sigmoid-function-numerically-stable-sigmoid-function-5298b14720f6).

ngazagna:
I incorporated the required changes (comments on the sigmoid and its link to the loss derivative, deletion of useless files). I think the `julia_gd.py` file is also needed.

ngazagna:
I've seen several changes of the structure of the repo. Should I git merge master into this DEV branch ?

agramfort:
no you should open a PR on this repo instead
https://github.com/benchopt/benchmark_logreg_l2

>

tomMoral:
Hello @ngazagna 

Sorry for the complex changes, we decided to move all benchmark in separated repo to avoid maintenance burden.
As @agramfort  said, you should open a new PR on [benchopt/benchmark_logreg_l2](https://github.com/benchopt/benchmark_logreg_l2) instead.


For the error, this should be fixed on master so if you open this new PR (which simply corresponds to cloning the repo and copying your solver in `solvers`), this should be fixed.
Thanks a lot

ngazagna:
Thanks for the answer. Perfect, I'll do that !

PR #67, opened 2020-09-22
[DEV] Add julia sgd solver for logreg l2

ngazagna:


agramfort (commented):
looks good !

@ngazagna can you share here a figure you obtain on a existing benchmark?

ngazagna:
> looks good !
> 
> @ngazagna can you share here a figure you obtain on a existing benchmark?

Sure, here is are two examples. Obviously, the other methods do not initialize the weights to zero, which might explain the initial loss gap.

[julia_sgd_1.pdf](https://github.com/benchopt/benchOpt/files/5266629/julia_sgd_1.pdf)

[julia_sgd_2.pdf](https://github.com/benchopt/benchOpt/files/5266627/julia_sgd_2.pdf)

agramfort:
this really tells me that we need a dedicated benchmark for datasets where
stochastic methods make sense.

>

PR #78, opened 2020-10-01
ENH install hook / JuliaSover / FIX cache

tomMoral:
Add pre/post install hook and use them to make sure julia is correctly installed.

This is to avoid having to clone manually `.julia/registries/General` in every repo, as the failures in benchopt/benchmark_logreg_l2#1 .
This requires changing a bit the API for Julia solver as it should now derive from `JuliaSolver` and the proper preamble is now:

```python
from benchopt.utils.julia_helpers import JuliaSolver
from benchopt.utils.julia_helpers import get_jl_interpreter
from benchopt.utils.julia_helpers import assert_julia_installed

with safe_import_context() as import_ctx:
    assert_julia_installed()
```

### DIFF
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -163,6 +163,7 @@ class DependenciesMixin:
             print(f"Installing {cls.name} in {env_name}:...",
                   end='', flush=True)
             try:
+                cls._pre_install_hook(env_name=env_name)
                 if cls.install_cmd == 'conda':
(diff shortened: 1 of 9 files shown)


agramfort (approved):
LGTM !

agramfort:
just CIs complain :(

tomMoral:
Yes this is because of the fix on the cache.
I have the solution but weird bug. Will push asap.

PR #84, opened 2020-11-08
ENH allow multi output objective

tomMoral:
Objective can now return a dictionary with multiple values that will be recorded.
this allow to monitor different metrics to display in the results.


agramfort:
Is this backward compatible?

tomMoral:
yep!

agramfort (approved):
Amen !

we'll bother about the doc later ;)

tomMoral:
I think we just miss a bit more doc on this one before merging.

PR #86, opened 2020-12-02
ENH add option to launch a debugger on error

tomMoral:
add a `--pdb` option to the `run` command so a debugger (`ipdb` if present else `pdb`) is open on error for the solver.
This should prove helpful to develop new solvers.


agramfort (approved):
nice !

PR #87, opened 2020-12-04
[CLN] replace : in filename

tanglef:
Replace forbidden character in name file for Windows.
The timestamp is now Year-month-day_HourhMinutemSeconds.


tomMoral (approved):
LGTM

tomMoral:
Thx @tanglef

PR #92, opened 2020-12-10
add some doc and split test-doc

tanglef:
Add some explanations about the multi output (especially the NaNs in the dataframe).

### DIFF
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -360,9 +360,13 @@ class BaseObjective(ParametrizedNameMixin):
     - `compute(beta)`: computes the value of the objective function for an
       given estimate beta. Beta is given as a flat 1D vector of size
       corresponding to the `scale` value returned by `Dataset.get_data`. The
-      output should be a float or a dictionary of floats. If a dictioanry is
-      returned, it should at least contain a key `objective_value` associated
-      to a scalar value which will be used to detect convergence.
+      output should be a float or a dictionary of floats.
+      If a dictionary is returned, it should at least contain a key
+      `objective_value` associated to a scalar value which will be used to detect
+      convergence.
+      With a dictionary, mutliple metric values can be stored at once instead
+      of runnning each separately.
+    
     """
 
     _base_class_name = 'Objective'
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -363,7 +363,7 @@ class BaseObjective(ParametrizedNameMixin):
       output should be a float or a dictionary of floats.
       If a dictionary is returned, it should at least contain a key
       `objective_value` associated to a scalar value which will be used to detect
-      convergence. With a dictionary, mutliple metric values can be stored at once instead
+      convergence. With a dictionary, multiple metric values can be stored at once instead
       of runnning each separately.
     
     """
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -363,8 +363,7 @@ class BaseObjective(ParametrizedNameMixin):
       output should be a float or a dictionary of floats.
       If a dictionary is returned, it should at least contain a key
       `objective_value` associated to a scalar value which will be used to detect
-      convergence.
-      With a dictionary, mutliple metric values can be stored at once instead
+      convergence. With a dictionary, mutliple metric values can be stored at once instead
       of runnning each separately.
     
     """
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -314,7 +314,10 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
     Returns
     -------
     df : instance of pandas.DataFrame
-        The benchmark results.
+        The benchmark results. If multiple metrics were computed, each
+        one is stored in separate column. If the number of metrics computed
+        by the objective is not the same for all parameters, the missing data
+        is set to `NaN`.
     """
     # Load the objective class for this benchmark and the datasets
     objective_class = get_benchmark_objective(benchmark)
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -315,7 +315,7 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
     -------
     df : instance of pandas.DataFrame
         The benchmark results. If multiple metrics were computed, each
-        one is stored in separate column. If the number of metrics computed
+        one is stored in a separate column. If the number of metrics computed
         by the objective is not the same for all parameters, the missing data
         is set to `NaN`.
     """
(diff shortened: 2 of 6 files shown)


benchopt/base.py
```diff
@@ -360,9 +360,13 @@ class BaseObjective(ParametrizedNameMixin):
...
       given estimate beta. Beta is given as a flat 1D vector of size
       corresponding to the `scale` value returned by `Dataset.get_data`. The
-      output should be a float or a dictionary of floats. If a dictioanry is
-      returned, it should at least contain a key `objective_value` associated
-      to a scalar value which will be used to detect convergence.
+      output should be a float or a dictionary of floats.
+      If a dictionary is returned, it should at least contain a key
+      `objective_value` associated to a scalar value which will be used to detect
+      convergence.
+      With a dictionary, mutliple metric values can be stored at once instead
```
mathurinm, line 367: typo

benchopt/runner.py
```diff
@@ -314,7 +314,10 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
     Returns
     -------
     df : instance of pandas.DataFrame
-        The benchmark results.
+        The benchmark results. If multiple metrics were computed, each
+        one is stored in separate column. If the number of metrics computed
```
mathurinm, line 318: in a separate

benchopt/viz/plot_convergence_curve.py
```diff
@@ -14,7 +14,7 @@ def plot_convergence_curve(df, benchmark):
     Returns
     -------
     fig : instance of matplotlib.figure.Figure
-        The matplotlib figure.
+        The matplotlib figure of the objective values.
```
mathurinm, line 17: Is it necessary to mention it ? It could be something else

doc/how.rst
```diff
@@ -29,10 +29,14 @@ A objective should define a `set_data` method that allows
 to specify the data and a `compute` method that allows
 to evaluate the objective for a given value of the iterate.
 
+Multiple values can be computed in one objective as long as they are
+stored in a dictionary with a key being `objective_value`. This allows
+to compare different metrics in the same run.
```
mathurinm, line 34: to compute different metrics in a single run ?

benchopt/viz/plot_convergence_curve.py
tanglef: For now, the only column plotted is explicitly set as being the 'objective_values'

doc/how.rst
tanglef: With multiple outputs, we can compute multiple metrics when the function compute from the objective is called.
So in theory run the command once and compute each value for the different metrics needed.

benchopt/base.py
```diff
@@ -360,9 +360,12 @@ class BaseObjective(ParametrizedNameMixin):
     - `compute(beta)`: computes the value of the objective function for an
       given estimate beta. Beta is given as a flat 1D vector of size
       corresponding to the `scale` value returned by `Dataset.get_data`. The
-      output should be a float or a dictionary of floats. If a dictioanry is
-      returned, it should at least contain a key `objective_value` associated
-      to a scalar value which will be used to detect convergence.
+      output should be a float or a dictionary of floats.
+      If a dictionary is returned, it should at least contain a key
+      `objective_value` associated to a scalar value which will be used to detect
+      convergence. With a dictionary, mutliple metric values can be stored at once instead
```
tomMoral, line 366:
```suggestion
      convergence. With a dictionary, multiple metric values can be stored at once instead
```

benchopt/viz/plot_convergence_curve.py
mathurinm: with the stochastic benchmark it will no longer be. for me it does not bring a lot to be more specific here and it will need to be changed soon. "The matplotlib figure with the results" ?

benchopt/runner.py
```diff
@@ -314,7 +314,10 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
     Returns
     -------
     df : instance of pandas.DataFrame
-        The benchmark results.
+        The benchmark results. If multiple metrics were computed, each
+        one is stored in a separate column. If the number of metrics computed
+        by the objective is not the same for all parameters, the missing data
+        is set to `NaN`.
```
tomMoral, line 320: Not sure if we want to add this. The computation should not output NaN in principle as the metric is always computed the same no?

doc/how.rst
mathurinm: my beef is with "compare" (a comparison should be between solvers, and then multiple runs are needed) but that's cosmetic

tomMoral, line 34:
```suggestion
to compute different metrics at once.
```

benchopt/tests/test_benchmarks/dummy_benchmark/doc_objective.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class Objective(BaseObjective):
+    name = "Lasso Regression"
+
+    parameters = {
+        'reg': [0.05, .1, .5]
+    }
+
+    def __init__(self, reg=.1, fit_intercept=False):
```
tomMoral, line 11: I would remove the `fit_intercept` as it is not used anywhere.

benchopt/runner.py
tanglef: The computation does not make NaNs. The conversion to csv by pandas does.
It's like in the test situation. It is possible to compare 1 metric for parameter a and 2 metrics for parameter b.
In that case, the final csv file contains the columns of the dic (one for each metric). But the first parameter doesn't have any value to put in the second metric column. So this part of the column is filled with NaNs by pandas.

doc/how.rst
tanglef: I can use the words "to monitor" used in the original PR if you prefer.

benchopt/runner.py
tomMoral: Ok

benchopt/tests/test_benchmarks/dummy_benchmark/doc_objective.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def compute(self, beta):
+        diff = self.y - self.X.dot(beta)
+        return .5 * diff.dot(diff) + self.lmbd * abs(beta).sum()
+
+    def _get_lambda_max(self):
+        return abs(self.X.T.dot(self.y)).max()
+
+    def to_dict(self):
+        return dict(X=self.X, y=self.y, lmbd=self.lmbd)
```
tomMoral, line 26: As this version is for the doc, I think we could add a couple of comments here no?

tomMoral (approved):
LGTM overall. Just some extra comments if you have the time but else ready to merge.

benchopt/tests/test_benchmarks/dummy_benchmark/doc_objective.py
tanglef: Comments added for compute, to_dict and parameters

tomMoral (approved):
LGTM!
Thx @tanglef

PR #96, opened 2020-12-10
Add relative suboptimality, suboptimality and optimality plots

ngazagna:
instead of `convergence_curve`, I created 3 types of plots without modifying the saved df
- `suboptimality_curve` : $F(x) - F(x*)$, the default one
- `relative_suboptimality_curve` : $(F(x) - F(x*)) / (F(x^0) - F(x*))$
- `optimality_curve` : $F(x)$

### DIFF
--- a/benchopt/viz/plot_suboptimality_curve.py
+++ b/benchopt/viz/plot_suboptimality_curve.py
@@ -1,0 +1,80 @@
+import matplotlib.pyplot as plt
+
+
+def plot_suboptimality_curve(df, benchmark, relative=False):
+    """Plot suboptimality curve for a given benchmark and dataset.
+
+    Plot suboptimality if relative == False, that is
+    F(x) - F(x*) as a function of x
+
+    Plot relative suboptimality if relative == True, that is
+    (F(x) - F(x*)) / (F(x^0) - F(x*)) as a function of x
+
+    with F(x*) is the smallest value reached in df.
+
+    Parameters
+    ----------
+    df : instance of pandas.DataFrame
+        The benchmark results.
+    benchmark : str
+        The path to the benchmark folder.
+
+    Returns
+    -------
+    fig : instance of matplotlib.figure.Figure
+        The matplotlib figure.
+    """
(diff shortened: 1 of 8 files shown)


benchopt/viz/plot_suboptimality_curve.py
```diff
@@ -0,0 +1,82 @@
...
+        The benchmark results.
+    benchmark : str
+        The path to the benchmark folder.
+
+    Returns
+    -------
+    fig : instance of matplotlib.figure.Figure
+        The matplotlib figure.
+    """
+    plot_suboptimality_curve(df, benchmark, relative=True)
```
agramfort, line 82:
```suggestion
    return plot_suboptimality_curve(df, benchmark, relative=True)
```

agramfort:
@ngazagna I pushed here so we can see the images it produces on circleci

agramfort:
see https://397-227720375-gh.circle-artifacts.com/0/dev/auto_examples/index.html
https://397-227720375-gh.circle-artifacts.com/0/dev/auto_examples/plot_run_benchmark.html#sphx-glr-auto-examples-plot-run-benchmark-py
https://397-227720375-gh.circle-artifacts.com/0/dev/auto_examples/plot_run_benchmark_python_R_julia.html#sphx-glr-auto-examples-plot-run-benchmark-python-r-julia-py

does it look good to you?

josephsalmon:
image

"Did not converge" seems to long.
and also:
what about changing `use_acceleration` to `acc` everywhere (would avoid too long strings)?

josephsalmon:
so:
`acceleration`
instead of
`use_acceleration=True`?

josephsalmon:
my bad, `accelerated`instead

agramfort:
@ngazagna you have a conflict to fix

ngazagna:
ready to merge I think

agramfort:
thanks @ngazagna 

let's now rename viz to plotting and get back to the stochastic log reg benchmarks.

let's keep the ball rolling !

PR #97, opened 2020-12-10
Fix PGD with R (LAPACK import)

tanglef:
Allow to use LAPACK library for the SVD with Anaconda rpy2 installation or other if Anaconda didn't replace it

### DIFF
--- a/benchopt/utils/r_helpers.py
+++ b/benchopt/utils/r_helpers.py
@@ -1,27 +1,30 @@
 import os
 
-# Make sure that R_HOME is loaded from the current interpreter to avoid
-# using the parent interpreter R_HOME in the sub-interpreter.
+# # Make sure that R_HOME is loaded from the current interpreter to avoid
+# # using the parent interpreter R_HOME in the sub-interpreter.
 if os.environ.get('R_HOME', None) is not None:
-    del os.environ['R_HOME']
+    os.environ.pop('R_HOME')
 
 
 import rpy2  # noqa: E402
 import rpy2.robjects.packages as rpackages  # noqa: E402
 try:
     from rpy2.robjects.packages import PackageNotInstalledError
+    import rpy2.situation
 except ImportError:
     # Backward compat for rpy2 version < 3.3
     try:
         from rpy2.rinterface_lib.embedded import \
             RRuntimeError as PackageNotInstalledError
+        import rpy2.situation
     except ImportError:
         # Backward compat for rpy2 version < 3
         from rpy2.rinterface import RRuntimeError as PackageNotInstalledError
 
 # Hide the R warnings
 rpy2.robjects.r['options'](warn=-1)
-
+# Set the R_HOME directory to the one of the R RHOME ouput
+os.environ['R_HOME'] = rpy2.situation.r_home_from_subprocess()
 
 def import_rpackages(*packages):
     """Helper to import R packages in the import_ctx"""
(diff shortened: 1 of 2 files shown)


benchopt/utils/r_helpers.py
```diff
@@ -1,27 +1,30 @@
 import os
 
-# Make sure that R_HOME is loaded from the current interpreter to avoid
-# using the parent interpreter R_HOME in the sub-interpreter.
+# # Make sure that R_HOME is loaded from the current interpreter to avoid
+# # using the parent interpreter R_HOME in the sub-interpreter.
```
agramfort, line 4:
```suggestion
# Make sure that R_HOME is loaded from the current interpreter to avoid
# using the parent interpreter R_HOME in the sub-interpreter.
```

benchopt/utils/r_helpers.py
```diff
@@ -3,25 +3,28 @@
...
 import rpy2.robjects.packages as rpackages  # noqa: E402
 try:
     from rpy2.robjects.packages import PackageNotInstalledError
+    import rpy2.situation
 except ImportError:
     # Backward compat for rpy2 version < 3.3
     try:
         from rpy2.rinterface_lib.embedded import \
             RRuntimeError as PackageNotInstalledError
+        import rpy2.situation
```
tomMoral, line 19: If it is always present (not compat for <3.3), this should be imported only once before l.11

benchopt/utils/r_helpers.py
```diff
@@ -3,25 +3,28 @@
...
     except ImportError:
         # Backward compat for rpy2 version < 3
         from rpy2.rinterface import RRuntimeError as PackageNotInstalledError
 
 # Hide the R warnings
 rpy2.robjects.r['options'](warn=-1)
-
+# Set the R_HOME directory to the one of the R RHOME ouput
+os.environ['R_HOME'] = rpy2.situation.r_home_from_subprocess()
```
tomMoral, line 27: Nice Fix!

benchopt/utils/r_helpers.py
tanglef: It was introduced in 2.9 so I didn't know how far you wanted the compatibility.

benchopt/utils/r_helpers.py
```diff
@@ -3,25 +3,28 @@
 # Make sure that R_HOME is loaded from the current interpreter to avoid
 # using the parent interpreter R_HOME in the sub-interpreter.
 if os.environ.get('R_HOME', None) is not None:
-    del os.environ['R_HOME']
+    os.environ.pop('R_HOME')
```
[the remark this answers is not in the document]
tanglef: Just an habit to avoid `del` from my side...

benchopt/utils/r_helpers.py
tomMoral: You call twice the same code so if this fails in l.13, this will also fail here (it is not a compat call here).
It should be imported in its own `try...except` to be compatible pior to 2.9 or simply called outside of the `try...except`.
I am in favor of the later but as you want.

benchopt/utils/r_helpers.py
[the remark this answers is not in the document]
tomMoral: I'd rather avoid unnecessary changes and I prefer `del` if you don't use the result.

tomMoral (commented):
Couple of comments.

tanglef:
Yes, you're right. I put the import outside the try except.

tomMoral (approved):
I did a small modification as I realized that the linting was not running on the PR and fixed a couple of flake8 error.
Good to merge for me once the test pass!
Thx @tanglef

tomMoral:
Thx a lot @tanglef !!
I had kind of given up on this one! :)

PR #99, opened 2020-12-10
RFC remove file util.py and add Benchmark

tomMoral:
Large refactoring to get rid of `util.py`, move `datasets,helpers` to proper sub modules and create a class `Benchmark` for easy benchmark manipulation.

This should make the code easier to maintain.


benchopt/utils/dependencies_mixing.py
```diff
@@ -0,0 +1,125 @@
+
```
agramfort, line 1: mixin not mixing

agramfort (approved):
just a nitpick

agramfort (approved):

PR #101, opened 2020-12-13
fix given bug found in nnls benchmark

agramfort:


tomMoral (commented):
This is already done in https://github.com/benchopt/benchOpt/blob/master/benchopt/base.py#L392

The bug was with `nnls` benchmark which was implementing `__call__` and not `compute`. This has been [fixed on `nnls` master](https://github.com/benchopt/benchmark_nnls/commit/ee1be2fc05cddaf32191ae363bf7c50409cfbdaa).

PR #105, opened 2020-12-15
ENH add benchopt publish command

tomMoral:
Add a `benchopt publish` command to upload the benchmark results to [`benchopt/results`](https://github.com/benchopt/results).

The command is based on `pygithub` and upload the file when given a token. The token can be set either directly with `-t TOKEN`, with an environment variable `BENCHO_GITHUB_TOKEN` or in the config file.

### DIFF
--- a/benchopt/cli.py
+++ b/benchopt/cli.py
@@ -7,7 +7,8 @@ from benchopt import run_benchmark
 from benchopt.viz import plot_benchmark
 from benchopt.benchmark import Benchmark
 
-from benchopt.utils.files import _get_output_folder
+from benchopt.config import get_global_setting
+from benchopt.utils.github import publish_result_file
 from benchopt.utils.dynamic_modules import _load_class_from_module
 from benchopt.utils.shell_cmd import _run_shell_in_conda_env, create_conda_env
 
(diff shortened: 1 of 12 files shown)


benchopt/cli.py
```diff
@@ -160,29 +161,47 @@ def run(benchmark, solver_names, forced_solvers, dataset_names,
...
-                f"Could not find result file {filename}. Available result "
-                f"files are:\n- {all_csv_files}"
-            )
-    else:
-        result_filename = all_csv_files[-1]
+    # Get the result file
+    benchmark = Benchmark(benchmark)
+    result_filename = benchmark.get_result_file(filename)
 
+    # Plot he results.
```
tomMoral, line 168:
```suggestion
    # Plot the results.
```

tomMoral (commented):
@agramfort do you want to review this one?

agramfort (commented):
works like a charm !

shall we push some doc in this PR?

tomMoral:
Merging this as it fixes some errors for plotting. Let's do some Doc in a follow up PR.

zaccharieramzi:
Is it possible that this PR has introduced a behaviour where plots are now presented in separate figures (and saved as such)?
If so is this behaviour intended?

PR #111, opened 2020-12-17
[ENH] adding markers to plots

josephsalmon:
This is to solve issues on the numbers of colors used for the plotting.
Adding markers will help to discriminate, and also increases the number of possible of possibility.

One issue though, it might requires some tweaking to have sames order of markers in plotly and matplotlib.

Examples are attached.
Screenshot from 2020-12-17 14-36-26
Screenshot from 2020-12-17 14-36-41

### DIFF
--- a/benchopt/plotting/helpers_compat.py
+++ b/benchopt/plotting/helpers_compat.py
@@ -4,17 +4,14 @@ try:
 except ImportError:
     go = None
 
-TRANSLATE_MARKERS = {v: i for i, v in enumerate(plt.Line2D.markers)}
 
 def fill_between_x(fig, x, q1, q9, y, color, marker, label, plotly=False):
-
     if not plotly:
         plt.loglog(x, y, color=color, marker=marker, label=label, linewidth=3)
         plt.fill_betweenx(y, q1, q9, color=color, alpha=.3)
         return fig
 
     color = f'rgba{color}'
-    marker = TRANSLATE_MARKER[marker]
     fig.add_trace(go.Scatter(
         x=x, y=y,
         line_color=color, marker_symbol=marker, mode='lines+markers',
--- a/benchopt/plotting/helpers_compat.py
+++ b/benchopt/plotting/helpers_compat.py
@@ -4,6 +4,7 @@ try:
 except ImportError:
     go = None
 
+TRANSLATE_MARKERS = {v: i for i, v in enumerate(plt.Line2D.markers)}
 
 def fill_between_x(fig, x, q1, q9, y, color, marker, label, plotly=False):
 
@@ -13,6 +14,7 @@ def fill_between_x(fig, x, q1, q9, y, color, marker, label, plotly=False):
         return fig
 
     color = f'rgba{color}'
+    marker = TRANSLATE_MARKER[marker]
     fig.add_trace(go.Scatter(
         x=x, y=y,
         line_color=color, marker_symbol=marker, mode='lines+markers',
--- a/benchopt/plotting/helpers_compat.py
+++ b/benchopt/plotting/helpers_compat.py
@@ -5,17 +5,17 @@ except ImportError:
     go = None
 
 
-def fill_between_x(fig, x, q1, q9, y, color, label, plotly=False):
+def fill_between_x(fig, x, q1, q9, y, color, marker, label, plotly=False):
 
     if not plotly:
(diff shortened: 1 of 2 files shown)


benchopt/plotting/helpers_compat.py
```diff
@@ -5,17 +5,17 @@
     go = None
```
tomMoral, line 6: A gross translation like this migh be okay. We don't need exact equivalence, more a stable one no?

```suggestion

TRANSLATE_MARKERS = {v: i for i, v in enumerate(plt.Line2D.markers)}
```

tomMoral (commented):
A few comments on the code organisation.

josephsalmon:
Updated in a way that works... though the rosette function between matplotlib and plotly is unclear...

Screenshot from 2020-12-17 16-07-27
Screenshot from 2020-12-17 16-07-37

benchopt/plotting/plot_objective_curve.py
```diff
@@ -34,11 +33,10 @@ def plot_objective_curve(df, plotly=False, suboptimality=False,
         The rendered figure, used to create HTML reports.
     """
     if plotly:
-        from plotly.validators.scatter.marker import SymbolValidator
-        markers = SymbolValidator().values[::12]
-        print(markers)
+        MARKERS = {i: i for i, v in enumerate(plt.Line2D.markers)}
     else:
-        markers = list(Line2D.markers)
+        MARKERS = {i: v for i, v in enumerate(plt.Line2D.markers)}
```
tomMoral, line 38: If this is not a constant, use lowercase.
```suggestion
        markers = {i: v for i, v in enumerate(plt.Line2D.markers)}
```

benchopt/plotting/plot_objective_curve.py
```diff
@@ -34,11 +33,10 @@ def plot_objective_curve(df, plotly=False, suboptimality=False,
         The rendered figure, used to create HTML reports.
     """
     if plotly:
-        from plotly.validators.scatter.marker import SymbolValidator
-        markers = SymbolValidator().values[::12]
-        print(markers)
+        MARKERS = {i: i for i, v in enumerate(plt.Line2D.markers)}
```
tomMoral, line 36:
```suggestion
        markers = {i: i for i, v in enumerate(plt.Line2D.markers)}
```

benchopt/plotting/plot_objective_curve.py
```diff
@@ -70,7 +68,7 @@ def plot_objective_curve(df, plotly=False, suboptimality=False,
 
         fill_between_x(
             fig, curve['time'], q1, q9, curve['objective_value'],
-            color=CMAP(i), marker=markers[i], label=solver_name, plotly=plotly
+            color=CMAP(i), marker=MARKERS[i], label=solver_name, plotly=plotly
```
tomMoral, line 71:
```suggestion
            color=CMAP(i), marker=markers[i], label=solver_name, plotly=plotly
```

tomMoral (approved):
Just a nitpick, otherwise LGTM.

What is the limitation for the number of plot? above 20?

benchopt/plotting/plot_objective_curve.py
josephsalmon: actually you are the one who got from markers to MARKERS in the first place?

CMAP has the same issue btw;)

benchopt/plotting/plot_objective_curve.py
josephsalmon: and for the markers I think >52 curves should work. For the color, I guess it will cycle earlier (I have not updated that , so it should be approx. 20 as before, except if some modulo is used for cycling on the colormap?)

benchopt/plotting/plot_objective_curve.py
tomMoral: CMAP is a global constant of the module -> uppercase 
I put MARKER as a global constant to but you revert it to local so it is lowercase.

Sorry for my TOC 😅

josephsalmon:
> Just a nitpick, otherwise LGTM.
> 
> What is the limitation for the number of plot? above 20?

So the cmap would hold with cyclic values so >20 is fine. For the markers, I am sure it goes above 50.
Not sure we need more than that.

tomMoral:
Thx a lot @josephsalmon !

PR #112, opened 2020-12-18
format scale convergence curve

tanglef:
Scientific notation on yscale instead of "physician" one


tomMoral (approved):
LGTM! thx @tanglef

PR #113, opened 2020-12-18
ENH skip api

tomMoral:
- Add skip API to avoid long runners.
- Rename `scale` to `dimension`
- Add some test for the runner.

### DIFF
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -19,29 +19,29 @@ class BaseSolver(ParametrizedNameMixin, DependenciesMixin, ABC):
 
     Solvers that derive from this class should implement three methods:
 
-    - set_objective(self, **objective_parameters): prepares the solver to be
-      called on a given problem. **objective_parameters are the output of the
-      method :code:`to_dict` from the benchmark objective. In particular, this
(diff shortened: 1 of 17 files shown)


agramfort:
No skip method ? 

Can you also update the narrative doc ?

tomMoral:
There is a skip method.

I will update the doc when I get the time :)
Now is :beer:  time.

benchopt/base.py
```diff
@@ -48,10 +50,30 @@ class BaseSolver(ParametrizedNameMixin, DependenciesMixin, ABC):
...
+        Parameters
+        ----------
+        objective: benchopt.BaseObjective
+            the objective function for the current optimization problem.
+
+        Returns
+        -------
+        skip : bool
+            Wether or not the solver should be skipped.
+        reason : str
```
agramfort, line 64:
```suggestion
        reason : str | None
```

agramfort (commented):
@tomMoral can you merge with master, update what's new page and document the skip method in the how.rst ? thanks

agramfort (approved):
Great let’s move on

PR #117, opened 2020-12-20
ENH publish for non-commiter create a PR

tomMoral:
For non-commiter on `benchopt/benchopt`, `benchopt publish` now create a PR automatically.


agramfort:
it would be great if someone without push rights on the results repo could try this.

maybe @ngazagna or @tanglef ?

ngazagna:
I'll, try this today

PR #118, opened 2020-12-21
add doc on publish

agramfort:


agramfort:
see https://487-227720375-gh.circle-artifacts.com/0/dev/publish.html#publish

good to go from my end !

PR #121, opened 2020-12-21
x axis scale scientific notation

tanglef:
Scientific notation for the x axis.
Unlike for the y-axis, scientific notations don't behave the same way.

- Without the `.` before the e the labels can be very (very) long,
- Rotating the labels is better otherwise even with only a few digits they sometimes overlap each other.


agramfort:
@tanglef can you show how it looks now?

mathurinm:
I pulled the PR, rerun pip install, pulled the lasso benchmark and I get:
image

I'm not sure this is up to date, but as far as I'm concerned I think this was quite ok, no ?

tanglef:
This is what I get wiith plotly now

Capture d’écran de 2020-12-22 09-08-39

agramfort:
can you call "make" in the benchopt/results repo and show me how this page
looks like?

https://benchopt.github.io/results/benchmark_lasso_benchopt_run_2020-12-16_10h55m50.html

tanglef:
These are the three curves I get doing so. And I'm seeing that the y format sometimes (see the third image) acts weird. Should I force the same behavior I did with the x-axis to keep the same number of digits ? 

Capture d’écran de 2020-12-22 09-54-11
Capture d’écran de 2020-12-22 09-51-51
Capture d’écran de 2020-12-22 09-54-23

agramfort:
oui +1 pour fixer aussl le yticks

>

agramfort:
and I would rather rotate the xticks on the other side. Feels more natural
to me

>

tanglef:
This is how the third one renders now (first picture). I put one more digit for the y axis otw we have several same values and that doesn't really make sense (see picture 2 with all the 1e5).

Capture d’écran de 2020-12-22 10-27-41

Capture d’écran de 2020-12-22 10-23-47

agramfort (approved):
thanks !

agramfort:
Thx @tanglef

PR #126, opened 2020-12-23
Plotly barplot

tanglef:
Make barplots with plotly for results (now all the plots will render nicely).

Below see how it renders. To the left is the old behavior (still accessible when `plotly=False`: this is not a capture of the website but a capture of the "make" page with `plotly=False`, it is exactly as before even with the changes made to adapt with plotly). To the right is with `plotly=True`.

Capture d’écran de 2020-12-23 17-14-11

Capture d’écran de 2020-12-23 17-15-04

### DIFF
--- a/benchopt/plotting/helpers_compat.py
+++ b/benchopt/plotting/helpers_compat.py
@@ -76,22 +76,17 @@ def make_bars(fig, heights, ticks, width, colors, times, plotly=False):
     else:
         colors = [f'rgba{color}' for color in colors]
         xi, _ = zip(*ticks)
-        fig.add_trace(go.Bar(x=xi, y=heights,
-                             width=[width] * len(xi),
-                             marker_color=colors))
         no_cv = [True if col ==
                  "rgba(0.8627, 0.8627, 0.8627)" else False for col in colors]
-        if any(no_cv):
-            pts = zip(list(np.array(xi)[no_cv]),
-                      [h_ / 2 for h_ in np.array(heights)[no_cv]])
-            fig.update_layout(annotations=[
-                go.layout.Annotation(
-                    x=pts_[0],
-                    y=np.log(pts_[1]),
-                    text="Did not converge",
-                    valign="bottom",
-                    textangle=-90, yanchor="middle",
-                    showarrow=False) for pts_ in pts])
+        text_ = ["Did not converge" if no_cv_ else " " for no_cv_ in no_cv]
+        fig.add_trace(go.Bar(x=xi,
+                             y=heights,
+                             width=[width],
+                             marker_color=colors,
+                             text=text_,
+                             textposition="inside",
+                             insidetextanchor='middle',
+                             textangle=-90))
         for idx, x_ in enumerate(xi):
             if not no_cv[idx]:
                 fig.add_trace(
--- a/benchopt/plotting/helpers_compat.py
+++ b/benchopt/plotting/helpers_compat.py
@@ -68,8 +68,9 @@ def make_bars(fig, heights, ticks, width, colors, times, plotly=False):
             ax.bar(x=xi, height=height, width=width,
                    color=colors[idx], edgecolor=edges)
             if colors[idx] == "w":
-                ax.annotate("Did not converge", xy=(xi, height/2), ha='center',
-                            va='center', color='k', rotation=90)
+                ax.text(xi, .5, "Did not converge",
+                        ha="center", va="center", color='k',
+                        rotation=90, transform=ax.transAxes)
             else:
                 plt.scatter(np.ones_like(times[idx]) * xi, times[idx],
                             marker='_', color='k', zorder=10)
--- a/benchopt/plotting/helpers_compat.py
+++ b/benchopt/plotting/helpers_compat.py
@@ -1,4 +1,5 @@
 import matplotlib.pyplot as plt
+import numpy as np
 try:
     import plotly.graph_objects as go
 except ImportError:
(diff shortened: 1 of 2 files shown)


agramfort:
is it doable to put the "do not converge" inside the bar rather than overlapping at the top?

agramfort:
actually the second plot does what I had in mind. Why did you share 2 plots?

tanglef:
> actually the second plot does what I had in mind. Why did you share 2 plots?

I showed two plots exactly because of the "Do not converge". Using the exact same way to get the coordinates of the text, sometimes it gets at the top and sometimes in the middle (which was what I first intended to do and still would like to do...)

agramfort:
Ok then let’s wait and try to fix it. If it is a pain we can postpone

tanglef:
Fixed it for plotly (and the code is simplified  by doing so).

Capture d’écran de 2020-12-23 19-50-55

tanglef:
And to finish, this fixes the "Did not converge" position with Matplotlib. It's all good now.

Capture d’écran de 2020-12-23 23-36-15

benchopt/plotting/helpers_compat.py
```diff
@@ -55,3 +56,43 @@ def add_h_line(fig, val, xlim=None, plotly=False):
...
+                        ha="center", va="center", color='k',
+                        rotation=90, transform=ax.transAxes)
+            else:
+                plt.scatter(np.ones_like(times[idx]) * xi, times[idx],
+                            marker='_', color='k', zorder=10)
+    else:
+        colors = [f'rgba{color}' for color in colors]
+        xi, _ = zip(*ticks)
+        no_cv = [True if col ==
+                 "rgba(0.8627, 0.8627, 0.8627)" else False for col in colors]
```
tomMoral, line 81: Not a big fan of checking for "did not converged" based on the color.
it seems error prone. isn't it possible to use missing value for heights?

benchopt/plotting/plot_histogram.py
```diff
@@ -2,9 +2,10 @@
 import matplotlib.pyplot as plt
 
 from .helpers import _color_palette
+from .helpers_compat import get_figure, make_bars
 
 
-def plot_histogram(df):
+def plot_histogram(df, plotly=False):
```
tomMoral, line 8: Can you add the parameter in the doc string?

tomMoral (commented):
Looks awesome @tanglef !
a couple of comments now, I will try to make a second pass later.

benchopt/plotting/helpers_compat.py
tanglef: Yes it's possible. I changed it with missing values for times (as heights is set to the max if it didn't converge).

benchopt/plotting/helpers_compat.py
```diff
@@ -55,3 +56,43 @@ def add_h_line(fig, val, xlim=None, plotly=False):
         line_dash='dot', line_color='black',
         mode='lines', showlegend=False
     ))
+
+
+def make_bars(fig, heights, ticks, width, colors, times, plotly=False):
```
agramfort, line 61:
```suggestion
def _make_bars(fig, heights, ticks, width, colors, times, plotly=False):
```

I would make this function private in the API

So you have a good excuse not to write a docstring ;)

agramfort (commented):
also bonus you could see how to add the plotly plots to the website as explained in https://sphinx-gallery.github.io/dev/auto_examples/plot_9_plotly.html

but maybe for the next PR ;)

tanglef:
> also bonus you could see how to add the plotly plots to the website as explained in https://sphinx-gallery.github.io/dev/auto_examples/plot_9_plotly.html
> 
> but maybe for the next PR ;)

Sure, I'll look into that when I get the time to do so.

agramfort (approved):
@tomMoral I let you merge if happy

benchopt/plotting/plot_histogram.py
```diff
@@ -40,27 +43,39 @@ def plot_histogram(df):
         df_tol = df_.groupby('stop_val').filter(
             lambda x: x['objective_value'].max() < c_star)
         if df_tol.empty:
+            colors[i] = "w" if not plotly else (.8627, .8627, .8627)
```
tomMoral, line 46:
```suggestion
            colors[i] = "w" if not plotly else PLOTLY_GRAY
```

benchopt/plotting/plot_histogram.py
```diff
@@ -2,15 +2,18 @@
 import matplotlib.pyplot as plt
 
 from .helpers import _color_palette
+from .helpers_compat import get_figure, _make_bars
```
tomMoral, line 5:
```suggestion
from .helpers_compat import get_figure, _make_bars

PLOTLY_GRAY = (.8627, .8627, .8627)
```

benchopt/plotting/plot_histogram.py
```diff
@@ -40,27 +43,39 @@ def plot_histogram(df):
...
-                        va='center', color='k', rotation=90)
-            rect_list.append(rect)
+            height_list.append(df.time.max())
+            times_list.append(np.nan)
             continue
         stop_val = df_tol['stop_val'].min()
         this_df = df_[df_['stop_val'] == stop_val]
-        rect_list.append(ax.bar(
-            x=xi, height=this_df['time'].mean(), width=width, color=colors[i]))
+        height_list.append(this_df['time'].mean())
```
tomMoral, line 53: I think it is safer to use the median time instead of the average one as it is more robust to large deviations caused for instance by overloaded systems:
```suggestion
        height_list.append(this_df['time'].median())
```

tomMoral (approved):
LGTM too! Just a few nitpicks but I can apply them directly. Will merge soon!

Thx for taking the time @tanglef ! If you have some time to do the doc, it would be awesome!

tomMoral:
Failure is unrelated to the PR. It is caused by an update in `celer` master that will be removed from the CI in #131 .
Merging this one.

PR #129, opened 2020-12-28
DOC change doc plot kinds

### LINKED ISSUE #124: DOC : Write on the website what are the plot kinds available
On : 
https://benchopt.github.io/cli.html#benchopt-plot
user don't know what they can write as `<kinds>`.

ngazagna:
Corrects the `help` of plot kinds and closes #124

### DIFF
--- a/benchopt/cli.py
+++ b/benchopt/cli.py
@@ -5,7 +5,7 @@ from pathlib import Path
 from benchopt import __version__
 from benchopt import run_benchmark
 from benchopt.benchmark import Benchmark
-from benchopt.plotting import plot_benchmark
+from benchopt.plotting import plot_benchmark, PLOT_KINDS
 
 from benchopt.config import get_global_setting
 from benchopt.utils.github import publish_result_file
@@ -17,6 +17,9 @@ BENCHMARK_TEST_FILE = Path(__file__).parent / 'tests' / 'test_benchmarks.py'
 
 
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
+
+
+PLOT_KINDS_NAMES = list(PLOT_KINDS.keys())
 
 
 @click.group(context_settings=CONTEXT_SETTINGS, invoke_without_command=True)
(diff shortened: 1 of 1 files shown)


agramfort:
If you write 

closes #124 in pr description so when we merge it will close the issue automatically

agramfort (approved):

tomMoral (commented):
I would like to avoid hard coding the kind of plots in the help directive. could you check if formating with the global variable is possible? if it is too complicated, don't spend hours on this.

ngazagna:
I did this instead : 
https://github.com/ngazagna/benchOpt/blob/6d62edc3f4d38c0e32448b81e3cc551f75293851/benchopt/cli.py#L158-L160
Can we remove the "+" ?

benchopt/cli.py
```diff
@@ -152,7 +155,9 @@ def run(benchmark, solver_names, forced_solvers, dataset_names,
               "folder.")
 @click.option('--kind', '-k', 'kinds',
               multiple=True, show_default=True, type=str,
-              help='Timeout a solver when run for more than <timeout> seconds')
+              help="Specify the type of figure to plot: " +
+              ", ".join("`{0}`".format(k) for k in PLOT_KINDS_NAMES[:-1]) +
```
tomMoral, line 159: Use f-string when possible:
```suggestion
              ", ".join(f"`{k}`" for k in PLOT_KINDS_NAMES[:-1]) +
```

tomMoral (commented):
A small nitpick.

For the formatting without a list, I am super convinced:
Screenshot_20201229-174114

At least put double quotes. but I think a list could be nicer.

benchopt/cli.py
```diff
@@ -19,6 +19,9 @@
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 
 
+PLOT_KINDS_NAMES = list(PLOT_KINDS.keys())
```
tomMoral, line 22: To make sure the names are all formatted the same (even the last one which is not processed at the same time), you could do:
```suggestion
PLOT_KINDS_NAMES = list(f'`{k}`' for k in PLOT_KINDS)
```
and then use the name directly in `join` and for the last one.

ngazagna:
What we did with @tomMoral 
Screenshot_kind

benchopt/cli.py
```diff
@@ -19,6 +19,9 @@
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 
 
+PLOT_KINDS_NAMES = list(f'`{k}`' for k in PLOT_KINDS)
+
+
```
tomMoral, line 24:
```suggestion
```

tomMoral (approved):
A small clean up but other wise, LGTM!

tomMoral:
Thx a lot @ngazagna for taking the time to do this! :fire:

PR #136, opened 2021-03-05
New HTML generation for results visualization

bmalezieux:


benchopt/runner.py
```diff
@@ -396,5 +396,5 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
 
     if plot_result:
         from benchopt.plotting import plot_benchmark
-        plot_benchmark(df, benchmark)
+        plot_benchmark(df, benchmark, save_file)
```
tomMoral, line 399: Overall, I would rather not have to pass `save_file` here. But if it is too complicated, let it be :)

tomMoral (commented):
A few comment, I will read more in detail after the holidays :)

benchopt/runner.py
bmalezieux: It is the main cause of tests failures, I'm trying to remove it but it seems complicated... However, adding it in the other files is not always possible so we will need to discuss about that in more details ^^

bmalezieux:
I fixed the filename error. Now the problem comes from the fact that plotly is not installed.

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,296 @@
...
+    -------
+    None
+    """
+
+    benchmark = output_dir
+    try:
+        print("Removing old html files")
+        shutil.rmtree(benchmark / BUILD_DIR)
+    except FileNotFoundError:
+        print("Building html files")
```
tomMoral, line 223:
```suggestion
    print("Removing old html files")
    if (benchmark / BUILD_DIR).exists():
        shutil.rmtree(benchmark / BUILD_DIR)

     print("Building html files")
```

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,296 @@
...
+
+from ..constants import PLOT_KINDS
+from .plot_histogram import plot_histogram  # noqa: F401
+from .plot_objective_curve import plot_objective_curve  # noqa: F401
+from .plot_objective_curve import plot_suboptimality_curve  # noqa: F401
+from .plot_objective_curve import plot_relative_suboptimality_curve  # noqa: F401 E501
+
+
+ROOT = Path(__file__).parent / "html"
+BUILD_DIR = Path("build")
```
tomMoral, line 20: Instead of `build`, maybe better to call this `html`
```suggestion
HTML_DIR = Path("html")
```

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,296 @@
...
+                    fig = plot_func(df_obj)
+                figures[data_name][objective_name][k] = export_figure(
+                    fig, f"{fname_short}_{n_figure}"
+                )
+                n_figure += 1
+
+    return dict(
+        figures=figures, dataset_names=dataset_names, fname_short=fname_short,
+        objective_names=objective_names, kinds=list(kinds)
+        )
```
tomMoral, line 79:
```suggestion
    )
```

benchopt/plotting/generate_html.py
tomMoral: Actually, I am not sure you want to delete the previous files.

Overall, I would say that when you generate the html, you want:
- to generate the html for the current dataframe.
- update the `benchmark_filename = BUILD_DIR / f"{benchmark}.html"` to  reference this `run`

That way you can aggregate the results and navigate easily all the results you have.

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,296 @@
...
+@click.command()
+@click.option('--pattern', '-k', 'patterns',
+              metavar="<pattern>", multiple=True, type=str,
+              help="Include results matching <pattern>.")
+@click.option('--benchmark', '-b', 'benchmarks',
+              metavar="<pattern>", multiple=True, type=str,
+              help="benchmarks to include.")
+@click.option('--root', type=str, default=None,
+              help="root directory for results files.")
+def plot_benchmark_html_all(patterns=(), benchmarks=(), root=None):
```
tomMoral, line 253: this should be a function whithout click and a corresponding click command should be put in `benchopt/cli/process_results.py`

tomMoral (commented):
It is a good start!

It would be nice if you could add some docstring to explain what each function is doing, to make this easier to follow.

also, I would like to ba able to aggregate the html plots as we go, instead of deleting the previous one.

We can talk about it IRL.

benchopt/plotting/generate_html.py
bmalezieux: When you say navigate, you want another html file to reference all the runs, as a "home page" ?

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,334 @@
...
+HTML_DIR_FIGURES = HTML_DIR / "figures"
+
+TEMPLATE_INDEX = ROOT / "templates" / "index.mako.html"
+TEMPLATE_BENCHMARK = ROOT / "templates" / "benchmark.mako.html"
+TEMPLATE_RESULT = ROOT / "templates" / "result.mako.html"
+TEMPLATE_LOCAL_RESULT = ROOT / "templates" / "local_result.mako.html"
+
+
+def generate_plot_benchmark(df, kinds, fname):
+    """Generate all possible plots for a given benchmark.
```
tomMoral, line 30:
```suggestion
    """Generate all possible plots for a given benchmark run.
```

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,334 @@
...
+
+    return dict(
+        figures=figures, dataset_names=dataset_names, fname_short=fname_short,
+        objective_names=objective_names, kinds=list(kinds)
+    )
+
+
+def export_figure(fig, fig_name):
+    """
+    Export figure to svg
```
tomMoral, line 84:
```suggestion
    Export figure to HTML or svg.
```

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,334 @@
...
+
+    Parameters
+    ----------
+    fig : plotly.graph_objs._figure.Figure
```
tomMoral, line 88:
```suggestion
    fig : plotly.graph_objs.Figure
```

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,334 @@
...
+
+    fig_name : str
+        Name to be given to the figure.
+
+    Returns
+    -------
+    str
+        Path to the figure.
+    """
+    print(type(fig))
```
tomMoral, line 99: debug print?
```suggestion
```

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,334 @@
...
+    result = dict(
+        fname=fname,
+        datasets=datasets,
+        **generate_plot_benchmark(df, kinds, fname)
+    )
+    results.append(result)
+
+    for result in results:
+        result['page'] = \
+            f"{str(result['fname_short']).replace('.csv', '.html')}"
```
tomMoral, line 145: This seems weird as `results` is guaranteed to have only 1 element no?
Maybe remove the `results` and return `[result]`?

tomMoral: Also, most of the code can be factorized with `get_results`.
It would maybe be better to have an option in `get_results` that allow to pass directly a dataframe and compute the associated element list of fnames?

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,334 @@
...
+
+    return results
+
+
+def render_benchmark(results, benchmark):
+    """ Render HTML from results using the template """
```
tomMoral, line 199:
```suggestion
    "Render HTML pages with all plots for 1 run of the benchmark."
```

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,334 @@
...
+        results=results,
+        benchmark=benchmark,
+        max_rows=15,
+        nb_total_benchs=len(results),
+        last_updated=datetime.now(),
+    )
+
+
+def render_index(benchmarks):
+    """ Render a home page to index the results from the template """
```
tomMoral, line 211:
```suggestion
    "Render the benchmark's home page to index all runs."
```

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,334 @@
...
+                    input_encoding="utf-8").render(
+        benchmarks=benchmarks,
+        nb_total_benchs=len(benchmarks),
+        max_rows=15,
+        last_updated=datetime.now(),
+    )
+
+
+def copy_static(benchmark=None):
+    """ Copy static files """
```
tomMoral, line 222:
```suggestion
    "Copy static files in the HTML output folder."
```

benchopt/plotting/generate_html.py
tomMoral: Actually, I am not sure what this one does.

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,334 @@
...
+        dst = benchmark / HTML_DIR / 'static'
+    else:
+        dst = HTML_DIR / 'static'
+    if dst.exists():
+        shutil.rmtree(dst)
+    shutil.copytree(ROOT / 'static', dst)
+
+
+def render_all_results(results, benchmark, local=True):
+    """ Create an html file containing the plot from results """
```
tomMoral, line 233:
```suggestion
    "Create an html file containing the plot from a benchmark run."
```

tomMoral (commented):
A few more comments, I will try to make an extra pass in the afternoon.

benchopt/plotting/generate_html.py
bmalezieux: Actually, it is not the same function (it does not take a list of dataframes but a list of csv files, the comments are not accurate). I already tried to factorize and I remember it was tricky.

tomMoral:
I did a few edits, let me know what you think of it.

bmalezieux:
It's very nice ! I think we've reached the expected behavior :)

benchopt/plotting/generate_html.py
```diff
@@ -0,0 +1,431 @@
...
+    ----------
+    benchmark_names : list of str
+        A list of all benchmark names that have been rendered.
+
+    Returns
+    -------
+    rendered : str
+        A str with the HTML code for the index page.
+    static_dir : str
+        Relative path from HTML root to the static files.
```
agramfort: I don't think static_dir is returned

agramfort (approved):
good to go!

merge when green @tomMoral

agramfort:
thx @bmalezieux and @tomMoral !

Issue #88, opened 2020-12-07
ENH gain time for some solvers by returning list of times and iterates


mathurinm:
When we code a solver specifically for a benchmark, we can return a list of `(time, iterate)` (or two lists), so that we get one full convergence curve per run, instead of having to run with different max_iter or tol. I think the time to get a curve should go from quadratic in max_iter to linear.

This means we trust the user to dump the correct time, but the curves could be altered anyways with the current process, so IMO we should assume good faith.

agramfort:
this was a design decision from the start to support all systems and
programming languages no matter
how they use time estimates.

>

mathurinm:
The time is a major bottleneck for me. If you take the example from the doc:
![image](https://user-images.githubusercontent.com/8993218/101355432-d71a3180-3896-11eb-872b-03778caa1afc.png)
It could be run in less than 5 * 2s, but it takes close to a minute on my machine.

I agree *we should keep the default behavior* which is very robust, but that doesn't mean we shouldn't allow for more efficient ones when possible

agramfort:
it's a trade off. It depends if you consider computation time free. I would
just use dragos in the background and move on

>

tomMoral:
Small note @mathurinm the complexity is not qudratic but O(T logT) :wink: 

I agree that adding this option might help diffuse `benchopt`.
When you develop a new solver and you want to use `benchopt` to evaluate it (which we would like to promote), you might want to reduce the time to be able to iterate more quickly so having such option would help a lot.

To me, it seems like a low complexity enhancement as it should be quite easy to support both.
The main issue here is that we need to agree on a good API.

mathurinm:
I believe that assuming an access to drago-like is restrictive for the potential user base. Being able to iterate quickly would be nice IMO

mathurinm:
solved by implementing the callback stopping strategy in #137

jolars:
Any chance of reconsidering this? I understand that the callback strategy can be used to solve this but that this would require that the solvers are coded within the benchopt framework. I'm sure that's okay in many cases, but at least on the project I'm working on right now we're coding our algorithm as a separate module (which I think makes most sense) and not directly in benchopt.

Setting aside the issue of overhead from the current approach, it may also be useful to just have the more fine-grained information from the full progress of the solver. Some solvers do not converge monotonically or have otherwise interesting behavior which doesn't necessarily show with the current approach (thinking of accelerated gradient descent methods or stochastic methods for instance). 

[...]

PR #137, opened 2021-03-05
ENH add API to run solver to convergence with callbacks

tanglef:
Enable less repetitive evaluations for some solvers like PGD.
From [issue88](https://github.com/benchopt/benchOpt/issues/88).


benchopt/runner.py
```diff
@@ -23,6 +23,16 @@
...
+RHO = 1.5
+EPS = 1e-10
+
+
+def get_next(x, rho=RHO, strategy="iteration"):
+    if strategy == "iteration":
+        return max(x + 1, min(int(rho * x), MAX_ITER))
+    else:
+        assert strategy == 'tolerance'
+        def get_next(x): return max(x / rho, MIN_TOL)
```
tomMoral, line 35: Oups! :)
```suggestion
        return max(x / rho, MIN_TOL)
```

benchopt/runner.py
```diff
@@ -158,11 +157,12 @@ def get_next(x): return max(x / rho, MIN_TOL)
 
     id_stop_val = 0
     stop_val = 1
+    rho_ = RHO
```
tomMoral, line 160:
```suggestion
    rho = RHO
```

benchopt/runner.py
```diff
@@ -189,11 +189,11 @@ def get_next(x): return max(x / rho, MIN_TOL)
         delta_objective = prev_objective_value - objective_value
         delta_objectives.append(delta_objective)
         if delta_objective == 0:
-            rho *= 1.2
+            rho_ *= 1.2
```
tomMoral, line 192:
```suggestion
            rho *= 1.2
```

benchopt/runner.py
```diff
@@ -189,11 +189,11 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
         if len(delta_objectives) > PATIENCE:
             delta_objectives.pop(0)
         prev_objective_value = objective_value
-        stop_val = get_next(stop_val)
+        stop_val = get_next(stop_val, rho=rho_, strategy=solver.stop_strategy)
```
tomMoral, line 196:
```suggestion
        stop_val = get_next(stop_val, rho=rho, strategy=solver.stop_strategy)
```

benchopt/runner.py
```diff
@@ -205,6 +205,36 @@ def get_next(x): return max(x / rho, MIN_TOL)
     return curve, status
 
 
+def get_callback(curve_one_rep, cb_status, objective, solver,
+                 meta_rep, deadline):
+    # make documentation
+    info = get_sys_info()
```
tomMoral, line 211: Looking back at the code, I think I would rather define `curve_one_rep` and `cb_status` in this function as it will make it clearer what it is.
Sorry for the change of mind ^^

benchopt/runner.py
```diff
@@ -205,6 +205,36 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+
+    def callback(it, beta_hat_i):
+        t0 = time.perf_counter()
+        cb_status["delta_t"] += t0 - cb_status["time"]
+        stop = False
+        if it == cb_status["next_stopval"]:
+            objective_dict = objective(beta_hat_i)
+            # Add system info in results
+            dict_ = dict(**meta_rep, solver_name=str(solver),
```
tomMoral, line 220: Solver should be added in `meta_rep`, this will make it easier to read the code.

benchopt/runner.py
```diff
@@ -265,11 +295,25 @@ def run_one_solver(benchmark, objective, solver, meta, max_runs, n_repetitions,
             # Force the run if needed
             deadline = time.time() + timeout / n_repetitions
 
-            curve_one_rep, status = run_one_to_cvg_cached(
-                benchmark=benchmark, objective=objective, solver=solver,
-                meta=meta_rep, max_runs=max_runs, deadline=deadline,
-                progress_str=progress_str, force=force
-            )
+            if hasattr(solver, "run_all_in_one"):
+                max_iter = int(2 * 1.5 ** (max_runs - 1))
```
tomMoral, line 299: This should use the value of `RHO`
```suggestion
                max_iter = int(2 * RHO ** (max_runs - 1))
```

tomMoral (commented):
A couple comments. This looks cool! :)

benchopt/runner.py
tanglef: I made both changes in the last commit :)

doc/how.rst
```diff
@@ -152,6 +152,17 @@ with no other dependencies. Here is an example:
 
 .. literalinclude:: ../benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd.py
 
+For solvers that allow access to the each iterates of the solution, the ``run`` method can be
+pass and instead use ``run_all_in_one``. The distinction is in the use of a ``callback`` parameter.
+This is a function that must be placed at the beginning of each iteration and at the very end
+of the method. It takes the number of the iterate and its value as arguments. Here is an example
+in the same situation as above:
```
tomMoral, line 159:
```suggestion
For solvers that allow access to the each iterates of the solution, it is possible to implement
``run_with_cb``. The distinction is that it is called with a ``callback`` that should be called at
each iteration with parameters the number of iterate ``it`` and the current value of the iterate
``beta``. Here is an example in the same situation as above:
```

doc/how.rst
```diff
@@ -152,6 +152,17 @@ with no other dependencies. Here is an example:
...
+
+.. literalinclude:: ../benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd_allinone.py
+  :pyobject: Solver.run
```
tomMoral, line 162:
```suggestion
```

tomMoral (commented):
Some leftover comment in the doc

benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd_allinone.py
```diff
@@ -0,0 +1,75 @@
+from benchopt import BaseSolver, safe_import_context
+
+with safe_import_context() as import_ctx:
+    import numpy as np
+    from scipy import sparse
+
+
+class Solver(BaseSolver):
+    name = 'Python-PGD-all-in-one'  # proximal gradient, optionally accelerated
```
tanglef, line 9:
```suggestion
    name = 'Python-PGD-with-cb'  # proximal gradient, optionally accelerated
```

benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd_allinone.py
```diff
@@ -0,0 +1,75 @@
...
+            rk = b_k1 @ b_k
+            if abs(rk - rk1) < 1e-10:
+                break
+
+            # re normalize the vector
+            b_k = b_k1 / np.linalg.norm(b_k1)
+
+        return rk
+
+    def run_with_cb(self, callback):
```
tanglef, line 51:
```suggestion
    def run_with_cb(self, callback):  # instead of the classic run
```

doc/how.rst
tanglef:
```suggestion
For solvers that allow access to each iterate of the solution, it is possible to implement
``run_with_cb``. The distinction is that is calls a ``callback`` that should be called at
each iteration with parameters the number of the iterate ``it`` and the current value of the iterate
``beta``. Here is an example in the same situation as above:
```

doc/how.rst
```diff
@@ -152,6 +152,17 @@ with no other dependencies. Here is an example:
...
+.. literalinclude:: ../benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd_allinone.py
+  :pyobject: Solver.run_all_in_one
```
tanglef, line 164:
```suggestion
  :pyobject: Solver.run_with_cb
```

tanglef (commented):
Just a few tweaks + use `` run_with_cb`` everywhere instead of ``run_all_in_one``

tanglef:
> Some leftover comment in the doc
See last review if is seems ok for you.

benchopt/base.py
```diff
@@ -87,7 +103,7 @@ def set_objective(self, **objective_dict):
         """
         ...
 
-    @abstractmethod
+    @not_implemented
```
agramfort, line 106: this was correct. It's an ABC class. abstractmethod in an ABC class are the methods to be implemented by subclasses. It's checked at runtime.

benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd_allinone.py
```diff
@@ -0,0 +1,75 @@
...
+                t_old = t_new
+                t_new = (1 + np.sqrt(1 + 4 * t_old ** 2)) / 2
+                w_old = w.copy()
+                z -= self.X.T @ (self.X @ z - self.y) / L
+                w = self.st(z, self.lmbd / L)
+                z = w + (t_old - 1.) / t_new * (w - w_old)
+            else:
+                w -= self.X.T @ (self.X @ w - self.y) / L
+                w = self.st(w, self.lmbd / L)
+        self.w = w
```
agramfort, line 72: You should not have to duplicate the full code. run method should here call run_with_cb using a dummy callback.

benchopt/base.py
tomMoral: The abstract class does not allow you to implement one of 2 methods. Here we want to have either `run` or `run_with_cb` implemented. so `not_implemented` is a mechanism to allow to check this while still having API definition in the class.

benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd_allinone.py
tomMoral: `run` is not implemented in this class so not sure what you mean?
As this is a test solver, I would say it is fine to remove the part about acceleration to avoid having too much duplicated code.

tomMoral:
your changes look good to me!

benchopt/base.py
agramfort: how about we enforce that run is always available but run_with_cb is optional. If it's available it's preferred if option says so. This would avoid duplicating all solvers and to use the same solver code with an without tracing using a callback?

tomMoral: I don't understand why we would need to duplicate the solver?
The solver should just implement one of the 2, not both. Then we can either call with a dummy cb or the cb that log everything.  I don't think requiring to implement both would be super user friendly.

benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd_allinone.py
agramfort: the rationale is for me:

- use callback if you can as it will speed up the benchmark generation
- do not duplicate solver code but provide by default what works the fastest
- so I would move all solvers to the callback setup if possible
- and have an option to disable the callback just to check that it does not affect results

let's talk on discord if it can help

tomMoral:
After some discussion appart with @agramfort, we think the best way to do this would actually to only have one `run` method but add a `stop_strategy = 'callback'` when implementing the callback. This way, we can keep only one abstract method and the API is simple.

Also, we mention that the callback should keep track of the number of iteration by itself in `cb_status`, so it can simply be called as `callback(w)`. This way we will avoid user that forget to increment `it` :)

tanglef:
> After some discussion appart with @agramfort, we think the best way to do this would actually to only have one `run` method but add a `stop_strategy = 'callback'` when implementing the callback. This way, we can keep only one abstract method and the API is simple.
> 
> Also, we mention that the callback should keep track of the number of iteration by itself in `cb_status`, so it can simply be called as `callback(w)`. This way we will avoid user that forget to increment `it` :)

All right, so we go back to the abstractmethod decorator for `run` right?
I'll work on that!

tomMoral:
Yep! Thanks @tanglef for cooping with our back and forth :)

tanglef:
So, after looking at the logs the error comes from... rpy2 (which isn't touched at all in this PR).
And i doesn't seem to be any sort of clash with the strategy.

benchopt/base.py
```diff
@@ -94,11 +94,20 @@ def run(self, stop_val):
         This function should not return the parameters which will be
         retrieved by a subsequent call to get_result.
 
+        If `stop_strategy` is set to `"callback"`, then it should call the
+        callback at each iteration. The callback will compute the time,
+        the objective function and store the relevant quantities fo BenchOpt.
```
tomMoral, line 99:
```suggestion
        the objective function and store relevant quantities for BenchOpt.
```

benchopt/base.py
```diff
@@ -94,11 +94,20 @@ def run(self, stop_val):
...
+        Else, the `stop_val` parameter should be specified.
+
         Parameters
         ----------
-        stop_val : int | float
+        stop_val : int | float | callable
             Value for the stopping criterion of the solver for. It allows to
             sample the time/accuracy curve in the benchmark.
+            If it is a callable, then it should act as a callback. This
+            callback that should be called after each iteration with parameter
```
tomMoral, line 108:
```suggestion
            callback should be called once for each iteration with argument
```

benchopt/base.py
```diff
@@ -94,11 +94,20 @@ def run(self, stop_val):
...
+            `beta_hat_it`. The callback returns False when the
+            computations should be stopped.
```
tomMoral, line 110:
```suggestion
            computations should stop.
```

benchopt/base.py
```diff
@@ -94,11 +94,20 @@ def run(self, stop_val):
         This function should not return the parameters which will be
         retrieved by a subsequent call to get_result.
 
+        If `stop_strategy` is set to `"callback"`, then it should call the
```
tomMoral, line 97:
```suggestion
        If `stop_strategy` is set to `"callback"`, then `run` should call the
```

benchopt/runner.py
```diff
@@ -205,6 +205,65 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
 
+def get_callback(objective, max_iter, deadline,  meta):
+    """Get callback function and store informations for the curve.
+
+    Parameters
+    ----------
+    objective : instance of BaseObjective
+        The objective to minimize.
+    maxiter : maximum number of iterations in the run.
+    deadline : maximum time allowed to perform the run.
```
tomMoral, line 216:
```suggestion
    max_iter : int
        Maximum number of iterations to run for.
    deadline : float
        Deadline after which to stop the computations. This will be
        used to respect the timeout for each solver.
```

benchopt/runner.py
```diff
@@ -205,6 +205,65 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+    meta : dict
+        Metadata passed to store in Cost results.
+        Contains objective, data, dimension
```
tomMoral, line 219:
```suggestion
        Contains objective and data names, problem dimension, etc.
```

benchopt/runner.py
```diff
@@ -205,6 +205,65 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+                return False
+
+            cb_status["next_stopval"] = min(
+                get_next(cb_status["next_stopval"]), max_iter
+            )
+        cb_status["it"] += 1
+        cb_status["time"] = time.perf_counter()
+        return True
+
+    return curve, cb_status, callback
```
tomMoral, line 264: Actually, would it be cleaner with an object? Not sure it is worth it.

doc/how.rst
```diff
@@ -112,7 +112,8 @@ A solver requires to define three methods:
 
    - ``run(stop_value)``: This method takes only one parameter that controls the stopping
      condition of the solver. This is typically a number of iterations ``n_iter``
-     or a tolerance parameter ``tol``. This is controled by the ``stop_strategy``,
+     or a tolerance parameter ``tol`` or a ``callback`` function that will be called until a
+     number of iteration. This is controled by the ``stop_strategy``,
```
tomMoral, line 116:
```suggestion
     or a tolerance parameter ``tol``. Alternatively, a ``callback`` function that will be
     called at each iteration can be passed. The callback returns False once the
     computation should stop. The parameter is controlled by the ``stop_strategy``,
```

doc/how.rst
```diff
@@ -133,6 +134,10 @@ This ``stop_strategy`` can be:
       the running time. The parameter is called ``tol`` and should be
       a positive float.
 
+    - ``'callback'``: in this case, the ``run`` method of the solver
+      will call at each iteration a callback function that computes the
+      objective until a high number of iteration is reached.
```
tomMoral, line 139:
```suggestion
    - ``'callback'``: in this case, the ``run`` method of the solver
      should call at each iteration the provided callback function. It will
      compute and store the objective and return False once the computations
      should stop.
```

tomMoral (approved):
This looks quite good! Just a few comments on the doc.

If you have some time there are 2 things that might be nice:
- Create an object for the callback. We can have a callable object with attribute. This will be cleaner than having a dict and better encapsulate everything. We can also have method like `get_curve` and `get status`, or `curve, status = cb.get_result()`.
- Add an option in the client to force the `runner`  to call  all solver with `run`, by passing a callback that just check `deadline` and `max_iter`. (typically, pass  `objecytive=None` in the callback`). This way, we can check easily if we have overhead, with the same code.

If you don't have time, don't feel pressured into doing these, we can merge this one and do it in next PRs.

tanglef:
> This looks quite good! Just a few comments on the doc.
> 
> If you have some time there are 2 things that might be nice:
> 
>     * Create an object for the callback. We can have a callable object with attribute. This will be cleaner than having a dict and better encapsulate everything. We can also have method like `get_curve` and `get status`, or `curve, status = cb.get_result()`.
> 
>     * Add an option in the client to force the `runner`  to call  all solver with `run`, by passing a callback that just check `deadline` and `max_iter`. (typically, pass  `objecytive=None` in the callback`). This way, we can check easily if we have overhead, with the same code.
> 
> 
> If you don't have time, don't feel pressured into doing these, we can merge this one and do it in next PRs.

Thanks.
I don't think the first point (making callback into an object) will pose much of a problem. So yes maybe in the same PR.
For the second one, maybe in another one? I'm not entirely sure (yet) of the changes needed for that.

And don't worry, there's no pressure from anyone. I work on it when I have the time to and it's not a "chore" :smiley:

tomMoral:
Ok then, let's do the object thing in this PR and we can work on the option in another PR.
I wait for the object one to merge.

As for the failure, I will look it up ASAP.

tanglef:
The error is the same (so no red flag that alerts me).
On my machine I was able to run the PGD with callback LASSO example (but it might be good to check on someone else's to be safe) :smiley:

benchopt/base.py
```diff
@@ -94,11 +94,20 @@ def run(self, stop_val):
...
+
         Parameters
         ----------
-        stop_val : int | float
+        stop_val : int | float | callable
             Value for the stopping criterion of the solver for. It allows to
             sample the time/accuracy curve in the benchmark.
+            If it is a callable, then it should act as a callback. This
+            callback should be called once for each iteration with argument
+            `beta_hat_it`. The callback returns False when the
```
agramfort, line 109: we should not call the optimization parameters beta. it's used in linear models but a generic name should be x or maybe theta

doc/how.rst
```diff
@@ -152,6 +159,14 @@ with no other dependencies. Here is an example:
 
 .. literalinclude:: ../benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd.py
 
+For solvers that allow access to each iterate of the solution, using ``"callback"``
+as a ``stop_strategy`` implies a slight modification for ``run``. A ``callback`` should be called at
+each iteration with parameter the current value of the iterate ``beta``.
```
agramfort, line 164: I am questioning this choice of the naming beta for param of optim

doc/how.rst
```diff
@@ -152,6 +159,14 @@ with no other dependencies. Here is an example:
...
+Here is an example in the same situation as above:
+
+.. literalinclude:: ../benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd_allinone.py
```
agramfort, line 167:
```suggestion
.. literalinclude:: ../benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd_callback.py
```

benchopt/runner.py
```diff
@@ -205,6 +205,77 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+            ))
+
+            if self.deadline is not None and time.time() > self.deadline:
+                self.cb_status["status"] = 'timeout'
+                return False
+            if self.cb_status["it"] == self.max_iter:
+                return False
+
+            self.cb_status["next_stopval"] = min(
+                get_next(self.cb_status["next_stopval"]), self.max_iter
```
agramfort, line 259:
```suggestion
                get_next(self.cb_status["next_stopval"], strategy="iteration"), self.max_iter
```
to be explicit that Callback only works with iterations

benchopt/runner.py
```diff
@@ -205,6 +205,77 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+        self.cb_status["time"] = time.perf_counter()
+        return True
+
+    def get_results(self):
+        """Get the results stored by the callback
+
+        Returns
+        -------
+        curve : dict
+            Details on the run and the objective value obtained.
```
agramfort, line 271: what are the keys and values?

benchopt/runner.py
```diff
@@ -205,6 +205,77 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+        cb_status : dict
+            Instantiated dict containing informations for the time
+            callback is called.
```
agramfort, line 274: what are the keys?

benchopt/runner.py
```diff
@@ -205,6 +205,77 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+        self.objective = objective
+        self.max_iter = max_iter
+        self.deadline = deadline
+        self.meta = meta
+        self.info = get_sys_info()
+        self.curve = []
+        self.cb_status = {
+                'status': 'unfinished',
+                "it": 0,
+                'next_stopval': 0,
```
agramfort, line 236: why do you need to store next_stopval? I would just keep it as attribute of Callback.

benchopt/runner.py
agramfort: also it's init as a list above

benchopt/runner.py
```diff
@@ -205,6 +205,77 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+        }
+
+    def __call__(self, beta_hat_i):
+        t0 = time.perf_counter()
+        self.cb_status["delta_t"] += t0 - self.cb_status["time"]
+        if self.cb_status["it"] == self.cb_status["next_stopval"]:
+            objective_dict = self.objective(beta_hat_i)
+            self.curve.append(dict(
+                    **self.meta, stop_val=self.cb_status["it"],
+                    time=self.cb_status["delta_t"],
```
agramfort, line 248: storing  delta_t in a variable called time seems wrong. Also why do you have time in curve and in the cb_status? it seems redundant

benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd_allinone.py
```diff
@@ -0,0 +1,76 @@
...
+    stop_strategy = "callback"
+
+    def set_objective(self, X, y, lmbd):
+        self.X, self.y, self.lmbd = X, y, lmbd
+
+    def st(self, w, mu):
+        w -= np.clip(w, -mu, mu)
+        return w
+
+    def compute_lipschitz_cste(self, max_iter=100):
```
agramfort, line 23:
```suggestion
    def compute_lipschitz_cst(self, max_iter=100):
```
in English it's contant and not contante

tanglef:
So I removed the dicts that were useful when not in a class but now we can treat them as attributes of Callback.
For the `beta` notation, I put an `x` in `base.py` and in the doc considering that we don't call it anywhere else I didn't explicity name it.
And lastly, the example with R/Julia needed to exclude Python-Python-callback file (o.w the fourth curve appeared on the plots and the callback has no place there i.m.o).

agramfort:
it's good to go from your end?

tanglef:
Yes

benchopt/runner.py
```diff
@@ -23,6 +23,16 @@
 MAX_ITER = int(1e6)
 MIN_TOL = 1e-15
 INFINITY = 3e38  # see: np.finfo('float32').max
+RHO = 1.5
+EPS = 1e-10
+
+
+def get_next(x, rho=RHO, strategy="iteration"):
```
agramfort, line 30: x is called stop_val elsewhere. Can you rename here?

benchopt/runner.py
```diff
@@ -205,6 +205,75 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+            The objective to minimize.
+        max_iter : int
+            Maximum number of iterations to run for.
+        deadline : float
+            Deadline after which to stop the computations. This will be
+            used to respect the timeout for each solver.
+        meta : dict
+            Metadata passed to store in Cost results.
+            Contains objective and data names, problem dimension, etc.
+        """
```
agramfort, line 225:
```suggestion
class Callback:
    """Get callback function and store informations for the curve.

    Parameters
    ----------
    objective : instance of BaseObjective
        The objective to minimize.
    max_iter : int
        Maximum number of iterations to run for.
    deadline : float
        Deadline after which to stop the computations. This will be
        used to respect the timeout for each solver.
    meta : dict
        Metadata passed to store in Cost results.
        Contains objective and data names, problem dimension, etc.
        
    Attributes
    ----------
    TODO
    """

    def __init__(self, objective, max_iter, deadline, meta):
```

benchopt/runner.py
```diff
@@ -205,6 +205,75 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+        return True
+
+    def get_results(self):
+        """Get the results stored by the callback
+
+        Returns
+        -------
+        curve : list
+            Details on the run and the objective value obtained.
+        status : string
```
agramfort, line 271:
```suggestion
        status : 'done' | 'diverged' | 'timeout'
```

agramfort (commented):
also can you share a plot you get with this PR on a real benchmark? I would be convinced by some end result

thx

tanglef:
The results I get on the LASSO with a few solvers (the callback/not callback are almost identical)

merged_callback_res

benchopt/runner.py
```diff
@@ -268,8 +272,7 @@ def get_results(self):
         -------
         curve : list
             Details on the run and the objective value obtained.
```
```suggestion
        curve : list
            Details on the run and the objective value obtained
            at each evaluated iteration.
```

benchopt/base.py
```diff
@@ -94,11 +94,20 @@ def run(self, stop_val):
...
+
         Parameters
         ----------
-        stop_val : int | float
+        stop_val : int | float | callable
             Value for the stopping criterion of the solver for. It allows to
             sample the time/accuracy curve in the benchmark.
+            If it is a callable, then it should act as a callback. This
+            callback should be called once for each iteration with argument
+            the current iterate `x`. The callback returns False when the
```
tomMoral, line 109: We called this `parameters` in the `get_result`. I would stick to this or change it there too.
```suggestion
            the current iterate `parameters`. The callback returns False when the
```

benchopt/runner.py
```diff
@@ -205,6 +205,78 @@ def get_next(x): return max(x / rho, MIN_TOL)
     return curve, status
 
 
+class Callback:
```
tomMoral, line 208: This `Callback` should be private as it is not part of the private API.
```suggestion
class _Callback:
```

benchopt/runner.py
```diff
@@ -205,6 +205,78 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+
+    def __init__(self, objective, max_iter, deadline, meta):
+        """Get callback function and store informations for the curve.
```
tomMoral, line 211:
```suggestion
        """Callback function to store information for the curve.
```

benchopt/runner.py
```diff
@@ -205,6 +205,78 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+
+    def __call__(self, x):
+        t0 = time.perf_counter()
+        self.time_iter += t0 - self.time_callback
+        if self.it == self.next_stopval:
+            objective_dict = self.objective(x)
+            self.curve.append(dict(
+                    **self.meta, stop_val=self.it,
+                    time=self.time_iter,
+                    **objective_dict, **self.info
```
tomMoral, line 251:
```suggestion
                **self.meta, stop_val=self.it,
                time=self.time_iter,
                **objective_dict, **self.info
```

benchopt/runner.py
```diff
@@ -205,6 +205,78 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+        return True
+
+    def get_results(self):
+        """Get the results stored by the callback
+
+        Returns
+        -------
+        curve : list
+            Details on the run and the objective value obtained.
+        status : 'done' | 'diverged' | 'timeout'
```
tomMoral, line 275:
```suggestion
        status : 'done' | 'diverged' | 'timeout' | 'unfinished'
```

benchopt/runner.py
```diff
@@ -260,16 +332,24 @@ def run_one_solver(benchmark, objective, solver, meta, max_runs, n_repetitions,
...
-                meta=meta_rep, max_runs=max_runs, deadline=deadline,
-                progress_str=progress_str, force=force
-            )
+            if solver.stop_strategy == "callback":
+                max_iter = int(2 * RHO ** (max_runs - 1))
+                callback = Callback(
+                    objective, max_iter, deadline, meta_rep
+                )
+                solver.run(callback)
+                curve_one_rep, status = callback.get_results()
```
tomMoral, line 346: This is neat with the object `callback`! Thx for doing this @tanglef

benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd_callback.py
```diff
@@ -0,0 +1,76 @@
...
+        L = self.compute_lipschitz_cst()
+        n_features = self.X.shape[1]
+        w = np.zeros(n_features)
+        if self.use_acceleration:
+            z = np.zeros(n_features)
+
+        it = 0
+        t_new = 1
+        while callback(w):
+            it += 1
```
tomMoral, line 62:
```suggestion
```

benchopt/tests/test_benchmarks/dummy_benchmark/solvers/python_pgd_callback.py
```diff
@@ -0,0 +1,76 @@
...
+        return rk
+
+    def run(self, callback):  # instead of the number of iteration or tolerance
+        L = self.compute_lipschitz_cst()
+        n_features = self.X.shape[1]
+        w = np.zeros(n_features)
+        if self.use_acceleration:
+            z = np.zeros(n_features)
+
+        it = 0
```
tomMoral, line 59:
```suggestion
```

tomMoral (commented):
A few more comments but it is starting to look good.

In the plot you shared, both implementation are roughly overlapped?

tanglef:
Yes they overlap (if you zoom, you can see at the beginning and in the "elbows" some blue get out. On some runs it's a little more visible (but not by much)

benchopt/runner.py
```diff
@@ -205,6 +205,78 @@ def get_next(x): return max(x / rho, MIN_TOL)
...
+            Deadline after which to stop the computations. This will be
+            used to respect the timeout for each solver.
+        meta : dict
+            Metadata passed to store in Cost results.
+            Contains objective and data names, problem dimension, etc.
+
+        Attributes
+        ----------
+        TODO
+        """
```
agramfort, line 229:
```suggestion
class _Callback:
    """Callback function to store information for the curve.

    Parameters
    ----------
    objective : instance of BaseObjective
        The objective to minimize.
    max_iter : int
        Maximum number of iterations to run for.
    deadline : float
        Deadline after which to stop the computations. This will be
        used to respect the timeout for each solver.
    meta : dict
        Metadata passed to store in Cost results.
        Contains objective and data names, problem dimension, etc.

    Attributes
    ----------
    TODO
    """
    def __init__(self, objective, max_iter, deadline, meta):

```

please also document the Attributes section

agramfort (commented):
any chance to merge with green CIs?

tanglef:
Green CI it is! (This one was way easier than expected considering the current issue with rpy2 in the LASSO repo)

tomMoral:
thx a lot @tanglef !
The CI problem is easier on this repo compared to the other because it is only r here and not glmnet 😅

agramfort:
feel free to merge https://github.com/benchopt/benchOpt/pull/148

>

PR #139, opened 2021-03-08
Enable autocompletion for BENCHMARK argument

gdurif:
When click autocompletion is activated, `benchopt run` suggests the list of sub-directories in current directory for the BENCHMARK argument.

### DIFF
--- a/benchopt/cli/main.py
+++ b/benchopt/cli/main.py
@@ -32,8 +32,10 @@ def get_datasets(ctx, args, incomplete):
 
 def get_benchmark(ctx, args, incomplete):
     skip_import()
-    benchmarks = [b.path for b in os.scandir('.') if b.is_dir()]
-    return [os.curdir] + [b for b in benchmarks if incomplete in b]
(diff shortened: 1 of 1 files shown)


benchopt/cli/main.py
```diff
@@ -29,10 +30,17 @@ def get_datasets(ctx, args, incomplete):
     return [d for d in datasets if incomplete.lower() in d]
 
 
+def get_benchmark(ctx, args, incomplete):
+    skip_import()
+    benchmarks = [b.path for b in os.scandir('.') if b.is_dir()]
+    return [os.curdir] + [b for b in benchmarks if incomplete in b]
```
agramfort, line 36: I would check if the folder contains an objective.py file to only show the folders that are likely to be benchmark folders

gdurif: Done in e4f8f3b

agramfort (approved):

agramfort:
thx @gdurif

Issue #130, opened 2021-01-12
BUG fail to run default benchmark (PyPI version)


gdurif:
Hi,

On a fresh install (inside a clean python virtual environment) via `pip`, I cannot run any of the standard benchmarks

```bash
$ pip install benchopt
$ git clone https://github.com/benchopt/benchmark_ols
$ benchopt run benchmark_ols 
Traceback (most recent call last):
  File "/xxxx/benchopt/.pyenv/bin/benchopt", line 11, in <module>
    sys.exit(main())
  File "/xxxx/benchopt/.pyenv/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/xxxx/benchopt/.pyenv/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/xxxx/benchopt/.pyenv/lib/python3.6/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/xxxx/benchopt/.pyenv/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/xxxx/benchopt/.pyenv/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/xxxx/benchopt/.pyenv/lib/python3.6/site-packages/benchopt/cli.py", line 80, in run
    validate_dataset_patterns(benchmark, dataset_names)
  File "/xxxx/benchopt/.pyenv/lib/python3.6/site-packages/benchopt/utils/checkers.py", line 35, in validate_dataset_patterns
    datasets = list_benchmark_datasets(benchmark)
  File "/xxxx/benchopt/.pyenv/lib/python3.6/site-packages/benchopt/util.py", line 104, in list_benchmark_datasets
    return _list_benchmark_classes(benchmark_dir, 'Dataset')
  File "/xxxx/benchopt/.pyenv/lib/python3.6/site-packages/benchopt/util.py", line 91, in _list_benchmark_classes
    classes.append(_load_class_from_module(module_filename, class_name))
  File "/xxxx/benchopt/.pyenv/lib/python3.6/site-packages/benchopt/utils/dynamic_modules.py", line 46, in _load_class_from_module
    module = _get_module_from_file(module_filename)
[...]

PR #142, opened 2021-03-11
temporary install instruction while PyPI version is not up-to-date

gdurif:
Temporary update of installation instruction:
```
$ git clone https://github.com/benchopt/benchOpt
$ pip install -e benchOpt
```
to replace `pip install benchopt` while PyPI version is 1.0.0 and thus not compatible with current version of the doc (c.f. #130)

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -39,7 +39,8 @@ This package can be installed through `pip` using:
 
 .. code-block::
 
-    $ pip install benchopt
+    $ git clone https://github.com/benchopt/benchOpt
+    $ pip install -e benchOpt
 
 This will install the command line tool to run the benchmark. Then, existing
 benchmarks can be retrieved from git or created locally. To discover which
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -35,12 +35,21 @@ Learn how to :ref:`how`.
 Install
 --------
 
-This package can be installed through `pip` using:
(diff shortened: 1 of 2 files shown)


doc/index.rst
```diff
@@ -39,7 +39,8 @@ This package can be installed through `pip` using:
 
 .. code-block::
 
-    $ pip install benchopt
+    $ git clone https://github.com/benchopt/benchOpt
+    $ pip install -e benchOpt
```
agramfort, line 43: don't assume people have git on their machine:

use:

`pip install -U https://github.com/benchopt/benchOpt/archive/master.zip`

besides I would suggest the 2 version. If you want to install the last release use this.... To install the latest development version use this...

thanks

gdurif: One might argue that `git` is used in several other section of the doc when cloning benchmarks.

Nonetheless, in e676220 I added the two possibility (last release and latest dev version). I also added a note (to be removed after the next release) recommending to use the development version at the moment (because of the discrepancy between the doc and PyPI version).

gdurif: Regarding the `git` question, it could  be a requirement, and we could write benchopt commands to wrap `git clone` and `git pull` to download and update standard benchmarks.

agramfort (approved):

agramfort:
looks good https://611-227720375-gh.circle-artifacts.com/0/dev/index.html

thx @gdurif

PR #144, opened 2021-03-17
ENH add xfail config file for benchmark test

tomMoral:
Add an API to allow `xfailing` specific solver in benchmark tests easily.

One should add a file `xfail.py` with a function `xfail_test_name` and use the fixeture `xfail_check` to be able to call this function in the test. See `test_benchmakr.py::test_solver_install` for instance.

### DIFF
--- a/benchopt/benchmark.py
+++ b/benchopt/benchmark.py
@@ -117,6 +117,18 @@ class Benchmark:
         "Get the location for the config file of the benchmark."
         return self.benchmark_dir / 'config.ini'
 
+    def get_xfail_file(self):
+        """Get the location for the xfail file for the benchmark.
+
+        This file will serve to check is a test should be xfailed on specific
+        solvers and platforms when we have install or running problems.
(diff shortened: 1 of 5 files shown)


benchopt/benchmark.py
```diff
@@ -117,6 +117,18 @@ def get_config_file(self):
         "Get the location for the config file of the benchmark."
         return self.benchmark_dir / 'config.ini'
 
+    def get_xfail_file(self):
+        """Get the location for the xfail file for the benchmark.
+
+        This file will serve to check is a test should be xfailed on specific
+        solvers and platforms when we have install or running problems.
```
tomMoral, line 124:
```suggestion
        This file will be used to check if a test should be xfailed on specific
        solvers and platforms when we have installation or running issues.
```

agramfort (approved):

agramfort:
thx @tomMoral !

PR #145, opened 2021-03-17
FIX unset R_HOME in all subprocesses

tomMoral:


tomMoral:
Closing in favor of #162

PR #146, opened 2021-03-20
[ENH]  better hovers with plotly

tanglef:
Now the overs show the whole name of the solver (better when different solvers with the same prefix showing)I
And for the quantiles traces, it removes the "trace .." on the right side of the hover box.


agramfort:
thx @tanglef !

can you see how to make the contrast always a bit better? See:

Screenshot 2021-03-21 at 09 10 20

tanglef:
The authors decided to only let users customize the main box of the hover text.
So for accessibility, we can remove the side box and put the name in the main one (the contrast is the same now).

Capture d’écran de 2021-03-21 12-33-26

agramfort (approved):
works great

thx @tanglef

PR #149, opened 2021-03-27
Stop the solver with callback

tanglef:
Use delta objective to stop the solver when using callbacks.
O.w, it went on until MAXITER was reached and because of the timeout only did one run and results were biased (see comment below, thank you @agramfort for noticing the problem).

See [nlls PR](https://github.com/benchopt/benchmark_nnls/pull/5) for a plot.

### DIFF
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -328,6 +328,8 @@ def run_one_solver(benchmark, objective, solver, meta, max_runs, n_repetitions,
     states = []
 
     with exception_handler(tag, pdb=pdb):
+        if solver.stop_strategy == "callback":
+            n_repetitions += 1  # warmup
         for rep in range(n_repetitions):
             if show_progress:
                 progress_str = (
@@ -354,7 +356,8 @@ def run_one_solver(benchmark, objective, solver, meta, max_runs, n_repetitions,
                     meta=meta_rep, max_runs=max_runs, deadline=deadline,
                     progress_str=progress_str, force=force
                 )
-
+            if rep == 0 and solver.stop_strategy == "callback":
+                continue
             curve.extend(curve_one_rep)
             states.append(status)
 
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -357,8 +357,6 @@ def run_one_solver(benchmark, objective, solver, meta, max_runs, n_repetitions,
     states = []
 
     with exception_handler(tag, pdb=pdb):
-        if solver.stop_strategy == "callback":
-            n_repetitions += 1  # warmup
         for rep in range(n_repetitions):
             if show_progress:
                 progress_str = (
@@ -385,8 +383,6 @@ def run_one_solver(benchmark, objective, solver, meta, max_runs, n_repetitions,
                     meta=meta_rep, max_runs=max_runs, deadline=deadline,
                     progress_str=progress_str, force=force
                 )
-            if rep == 0 and solver.stop_strategy == "callback":
-                continue
             curve.extend(curve_one_rep)
             states.append(status)
 
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -236,6 +236,7 @@ class _Callback:
         self.it = 0
         self.time_iter = 0.
         self.next_stopval = 0
+        self.prev_objective = 1e10
         self.time_callback = time.perf_counter()
 
     def __call__(self, x):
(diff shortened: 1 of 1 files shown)


tanglef:
EDIT: when doing a run without callback, there is some sort of warmup done by the first step, so there is no harm done for the rest of the run. For example, on the plot in nlls PR, we see the PGD in blue at the very beginning much higher in time.

With a callback, this warmup impacts the whole run (more exactly the first rep), as we can see on the same plot with the large shaded orange area... So I made a warmup turn for callback strategy and we get this (which is i.m.o much better, we don't have any danger of seeing this big polygone over everyone anymore). And we can also see here in light blue for PGD acceleration=False that at first, it takes more time (and the csv file tells us that it's only more for the first run).

Capture d’écran de 2021-03-27 22-47-44

If you want to put a warmup for everybody or chose to do something else instead (or even at another time), feel free to tell me. 

PS: the warmup may also help for Numba/packages that need a computation first, to only count the run without the computation that happens once.

benchopt/runner.py
```diff
@@ -236,6 +236,7 @@ def __init__(self, objective, max_iter, deadline, meta):
         self.it = 0
         self.time_iter = 0.
         self.next_stopval = 0
+        self.prev_objective = 1e10
```
agramfort, line 239:
```suggestion
        self.prev_objective = np.inf
```
?

benchopt/runner.py
```diff
@@ -254,7 +255,12 @@ def __call__(self, x):
                 return False
             if self.it == self.max_iter:
                 return False
+            current_obj = objective_dict["objective_value"]
+            if (-EPS <= self.prev_objective - current_obj <= EPS):
+                self.status = 'done'
+                return False
 
+            self.prev_objective = current_obj
```
agramfort, line 263: is prev_objective in the class docstring?

agramfort (commented):
sorry I am lost with the "warmup" thing.

we have things like https://github.com/benchopt/benchmark_lasso/blob/master/solvers/cd.py#L48 to cache the numba compilation time.

benchopt/runner.py
tanglef: It wasn't but it is now

benchopt/runner.py
tanglef: I put `math.inf` as numpy is not imported in the runner

tanglef:
I didn't know for the compilation thanks for that.
And what I call "warmup" is simply one more rep that isn't stored for the plots/medians (just like what you did with numba, but not made by the user in the solver file, rather directly into the runner).

I don't know why this happens (there is no compilation with PGD as far as I know), but it does

See for example another I found for the PGD with MCP [here in benchopt results](https://benchopt.github.io/results/benchmark_mcp_benchopt_run_2020-12-16_10h35m29.html) at the first point there is a large area, meaning that there was a run (the first) which took way longer than the followings.

And it wasn't really impactful without callbacks as it only changed one value of one run
But with callbacks, it impacts the whole first run (technically still the 1st one of the 1st run, but there is intra dependance for the runs now). And without correction, this leads to a big part of the plot shaded (like I posted on your nlls PR) and it isn't good looking.

agramfort:
> I didn't know for the compilation thanks for that.
> And what I call "warmup" is simply one more rep that isn't stored for the plots/medians (just like what you did with numba, but not made by the user in the solver file, rather directly into the runner).

I am a bit uncomfortable with this. A repetition can be a lot of time.
For large datasets I always set -r 1

> I don't know why this happens (there is no compilation with PGD as far as I know), but it does

can you add to the PGD class the call to run(1) to see if it fixes the
pb like we do for numba?
maybe it's the time to load numpy linalg in memory?

> And it wasn't really impactful without callbacks as it only changed one value of one run
> But with callbacks, it impacts the whole first run (technically still the 1st one of the 1st run, but there is intra dependance for the runs now). And without correction, this leads to a big part of the plot shaded (like I posted on your nlls PR) and it isn't good looking.

let me know if the suggestions above help

tanglef:
Yes, I didn't think of that...

Capture d’écran de 2021-03-28 22-03-01

That is the results with a run for PGD in nlls. The blue shade (standard PGD with iteration strategy) for the first point is removed with a `self.run(1)` and the orange shade (PGD with callback) with `self.run(lambda _: False)` so we do not enter the while loop.

agramfort (approved):
thx @tanglef 

@tomMoral feel free to MRG

tomMoral:
Actually, I had something a bit more complex in mind for two reasons:

I would like the behaviors to be the same for all `stop_strategy` with factorized code.
I would like to be able to simply change the behvior to introduce better stopping criterion.

I proposed PR #151 for this, let me know what you think about it.

tomMoral:
Closing this one as we merged #151.
Thx for the proposal @tanglef !

PR #151, opened 2021-04-14
ENH add stopping criterion

tomMoral:
This is a concurrent solution for #149.

I propose to refactor the code for the stopping criterion with 2 parts:

- One which check global stuff like `timeout`, `max_runs` and make sure we stop when we diverged.
- One which is for now fixed but should be made objective specific: the actual stopping criterion.

For now, I implemented the `SufficientDescentCriterion`, which check the same as what we were checking before.
But we could add other easily and make the choice depend on the `objective`.

### DIFF
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -120,9 +120,8 @@ def run_one_to_cvg(benchmark, objective, solver, meta, stopping_criterion,
     -------
     curve : list of Cost
         The cost obtained for all repetitions.
-    status : str
-        The status on which the solver was stopped. Should be one of
-        {'done', 'timeout', 'max_runs', 'diverged', }
+    status : 'done' | 'diverged' | 'timeout' | 'max_runs'
+        The status on which the solver was stopped.
     """
 
     # Create a Memory object to cache the computations in the benchmark folder
@@ -131,15 +130,9 @@ def run_one_to_cvg(benchmark, objective, solver, meta, stopping_criterion,
 
     # compute initial value
     stopping_criterion.show_progress('initialization')
-    init_stop_val = (
-        0 if solver.stop_strategy == 'iteration' else INFINITY
-    )
-    call_args = dict(
-        objective=objective, solver=solver, meta=meta
-    )
-    cost = run_one_resolution_cached(
-        stop_val=init_stop_val, **call_args
-    )
+    init_stop_val = (0 if solver.stop_strategy == 'iteration' else INFINITY)
+    call_args = dict(objective=objective, solver=solver, meta=meta)
+    cost = run_one_resolution_cached(stop_val=init_stop_val, **call_args)
     curve = [cost]
     stop, status, is_flat = stopping_criterion.should_stop_solver(curve)
 
@@ -191,8 +184,8 @@ class _Callback:
         A dictionary with info from the current system.
     curve : list
         The convergence curve stored as a list of dict.
-    status : 'done' | 'diverged' | 'timeout' | 'max_runs'
-        The convergence status.
+    status : 'running' | 'done' | 'diverged' | 'timeout' | 'max_runs'
+        The status on which the solver is or was stopped.
     time_iter : float
         Computation time to reach the current iteration.
         Excluding the times to evaluate objective.
@@ -262,7 +255,8 @@ class _Callback:
         -------
         curve : list
             Details on the run and the objective value obtained.
-        status : 'done' | 'diverged' | 'timeout' | 'unfinished'
+        status : 'done' | 'diverged' | 'timeout' | 'max_runs'
+            The status on which the solver was stopped.
         """
         return self.curve, self.status
 
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -7,7 +7,7 @@ from .benchmark import _check_name_lists
 from .utils.sys_info import get_sys_info
 from .utils.pdb_helpers import exception_handler
 
-from .stopping_criterion import SufficientDescentCondition
+from .stopping_criterion import SufficientDescentCriterion
 
 from .utils.colorify import colorify
 from .utils.colorify import LINE_LENGTH, RED, GREEN, YELLOW
@@ -318,7 +318,7 @@ def run_one_solver(benchmark, objective, solver, meta, max_runs, n_repetitions,
 
             meta_rep = dict(**meta, idx_rep=rep, solver_name=str(solver))
 
-            stopping_criterion = SufficientDescentCondition._get_instance(
+            stopping_criterion = SufficientDescentCriterion._get_instance(
                 max_runs=max_runs, timeout=timeout / n_repetitions,
                 progress_str=progress_str
             )
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -14,6 +14,7 @@ from .utils.colorify import LINE_LENGTH, RED, GREEN, YELLOW
 
 
 # Get config values
+from .config import DEBUG
 from .config import RAISE_INSTALL_ERROR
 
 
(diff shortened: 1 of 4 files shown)


benchopt/stopping_criterion.py
```diff
@@ -0,0 +1,192 @@
...
+        cost_curve : list of dict
+            List of dict containing the values associated to the objective at
+            each evaluated points.
+
+        Returns
+        -------
+        stop : bool - wether or not we should stop the algorithm.
+        status : str - reason why the algorithm was stopped if stop is True.
+        is_flat : bool - indicate if 2 successive evaluation yielded the same
+            objective value. Useful for moving faster on tolerance.
```
agramfort, line 65:
```suggestion
        stop : bool
            Whether or not we should stop the algorithm.
        status : str
            Reason why the algorithm was stopped if stop is True.
        is_flat : bool
            Indicate if 2 successive evaluation yielded the same
            objective value. Useful for moving faster on tolerance.
```

benchopt/stopping_criterion.py
```diff
@@ -0,0 +1,192 @@
...
+        cost_curve : list of dict
+            List of dict containing the values associated to the objective at
+            each evaluated points.
+
+        Returns
+        -------
+        stop : bool - wether or not we should stop the algorithm.
+        progress : float - measure of how far the solver is from convergence.
+            This should be in [0, 1], 0 meaning no progress and 1 meaning
+            that the solver has converged.
```
agramfort, line 137:
```suggestion
        stop : bool
            Whether or not we should stop the algorithm.
        progress : float
            Measure of how far the solver is from convergence.
            This should be in [0, 1], 0 meaning no progress and 1 meaning
            that the solver has converged.
```

benchopt/stopping_criterion.py
```diff
@@ -0,0 +1,192 @@
...
+            **self.kwargs
+        )
+
+
+class SufficientDescentCriterion(StoppingCriterion):
+    """Stopping criterion based on sufficient descent.
+
+    The solver will be stopped once successive evaluations do not make enough
+    progress. The number of successive evaluation and the definition of
+    sufficient progress is controled by ``eps`` and ``patience``.
```
agramfort, line 153: add proper docstring

benchopt/stopping_criterion.py
```diff
@@ -0,0 +1,192 @@
...
+
+        Parameters
+        ----------
+        cost_curve : list of dict
+            List of dict containing the values associated to the objective at
+            each evaluated points.
+
+        Returns
+        -------
+        stop : bool - wether or not we should stop the algorithm.
```
agramfort, line 173: idem as above

doc/whats_new.rst
```diff
@@ -34,6 +34,10 @@ API
 - ``Solver.skip`` can now be used to skip objectives that are incompatible
   for the Solver, by `Thomas Moreau`_ (:gh:`113`).
 
+- ``Solver`` can now use ``stop_strategy = 'callback'`` to allow for
```
agramfort, line 37:
```suggestion
- ``Solver`` can now uses ``stop_strategy = 'callback'`` to allow for
```

benchopt/stopping_criterion.py
```diff
@@ -0,0 +1,192 @@
...
+        # Modify the criterion state:
+        # - compute the number of run with the curve. We need to remove 1 as
+        #   it contains the initial evaluation.
+        # - compute the delta_objective for debugging and stalled progress.
+        n_eval = len(cost_curve) - 1
+        objective_value = cost_curve[-1]['objective_value']
+        delta_objective = self._prev_objective_value - objective_value
+        self._prev_objective_value = objective_value
+
+        # default for value for is_flat
```
tanglef, line 76:
```suggestion
        # default value for is_flat
```

benchopt/stopping_criterion.py
```diff
@@ -0,0 +1,192 @@
...
+            stop = True
+            status = 'timeout'
+
+        elif n_eval == self.max_runs:
+            stop = True
+            status = 'max_runs'
+
+        elif delta_objective < -1e10:
+            stop = True
+            status = 'diverged'
```
tanglef, line 91: This says that the objective should be decreasing, otherwise the solver is not converging and there is not point to keep it going (if I get it right).
But what about cases like with logreg_l2's  `copt[accelerated=True,llinesearch=False,solver=pgd]`  ([see on benchopt.results](https://benchopt.github.io/results/benchmark_logreg_l2_benchopt_run_2021-03-18_02h11m46.html) with the `simulated[200, 500]` dataset, the purple curve) where the objective increases at one point but the solver still converges at the end (also happens with [cvxpy in logreg_l1](https://benchopt.github.io/results/benchmark_logreg_l1_benchopt_run_2020-12-20_18h19m52.html) in the first iterations)

doc/whats_new.rst
tanglef: `use` is actually correct.

benchopt/runner.py
```diff
@@ -212,32 +172,26 @@ class _Callback:
...
+        Object to check if we need to stop a solver.
     meta : dict
         Metadata passed to store in Cost results.
         Contains objective and data names, problem dimension, etc.
     info : dict
         A dictionary with info from the current system.
     curve : list
         The convergence curve stored as a list of dict.
-    status : 'done' | 'diverged' | 'timeout' | 'unfinished'
+    status : 'done' | 'diverged' | 'timeout' | 'max_runs'
```
tanglef, line 194: `get_results` docstring for `status` has an `unfinished` that should be replaced now.

tanglef (commented):
Way more complete than my little patch!
I'm just wondering about the non-decreasing objective part but o.w looks great.

benchopt/runner.py
tomMoral: good catch!

benchopt/runner.py
```diff
@@ -358,35 +324,32 @@ def run_one_solver(benchmark, objective, solver, meta, max_runs, n_repetitions,
 
             meta_rep = dict(**meta, idx_rep=rep, solver_name=str(solver))
 
-            # Force the run if needed
-            deadline = time.time() + timeout / n_repetitions
+            stopping_criterion = SufficientDescentCriterion._get_instance(
```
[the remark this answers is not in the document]
tomMoral: After talking IRL, this design is in order to allow for easy sub-classing.
When the user subclass the `StoppingCriterion`, he can define a `__init__` function without calling the `super().__init__` as we have this factory that take care of this.
I will add some comments in the class design.

benchopt/stopping_criterion.py
tomMoral: Here, we are checking if the objective exploded (increased of 10^10). So it should only kick in in really bad situation I think. But let me know if you see this kick in some cases?

doc/whats_new.rst
tomMoral: @agramfort  got carried by the momentum of adding `s` to all my docstrings :rofl:

benchopt/stopping_criterion.py
tanglef: Yes, sorry, I read it with ` e-10`  (not the same quantity at all ^^' ) so it's all good

agramfort:
i tried this branch on the OLS benchmark and on boston dataset. It does not work as before. I see negative percentage and the solver does not seem to converge.

agramfort:
in the end it's converging but I still see negative percentages

benchopt/stopping_criterion.py
tomMoral: To be completely fair, I did the same multiple time :grin:

tomMoral:
Indeed, I did a typo in cleaning the code it appears. The negative percentage are fixed.

agramfort (approved):

agramfort:
thx @tomMoral !

PR #154, opened 2021-04-21
MTN migrate CI on github action

tomMoral:
Tentative to migrate CI on github action.


setup.cfg
```diff
@@ -68,4 +68,3 @@ doc =
     sphinx_gallery
     pillow
     scikit-learn
-    sklearn-contrib-lightning @ git+https://github.com/scikit-learn-contrib/lightning.git
```
agramfort, line 71: lightning est sur la doc

https://benchopt.github.io/

à moins que tu veux qu'on skip if not installed?

tomMoral: It is indeed skipped when not installed and I this deps make it impossible to push on pypi so we need to remove it.

agramfort:
ok !

>

PR #135, opened 2021-03-05
ENH separate install and run commands

gdurif:
ChangeLog

- New command `benchopt install <benchmark>` to install the requirements and dependencies for the provided benchmark, with management of environment (possible installation in current env), explicite verbosity during installation process, ask for confirmation when installing in current env or in user env, possible to force installation, to select solvers/datasets.
- Command `benchopt run <benchmark>` does not install anything anymore.
- Doc update.
- Message `"BenchOpt is running"` when calling `benchopt run <benchmark>`.
- Minor fixes: typo in a comment

PR #155, opened 2021-04-22
CLN remove allow_install

tomMoral:
The rational of allow install was to avoid silently installing stuff on `benchopt run`.
Now that per #135, `install` and `run` are spearated, the user knows that he is doing an install so the allow install is not really relevant anymore.

this PR supress it as it might be confusing.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -29,6 +29,10 @@ Changelog
 - Move most CI to github action, with auto-release on pypi,
   by `Thomas Moreau`_ (:gh:`150`, :gh:`154`).
 
+- Remove ``BENCHOPT_ALLOW_INSTALL`` and always install to requested env as
+  user now must request explicitely the install,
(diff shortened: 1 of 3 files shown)


doc/whats_new.rst
```diff
@@ -68,7 +72,7 @@ CLI
 - Add ``benchopt config`` command to allow easy configuration of ``benchopt``
   using the CLI, by `Thomas Moreau`_ (:gh:`128`).
 
-- Add ``benchopt install`` command to install benchmark requirements
+  - Add ``benchopt install`` command to install benchmark requirements
```
agramfort, line 75: revert?

agramfort (approved):
LGTM otherwise

doc/whats_new.rst
tomMoral: yep, gone too fast... :)
Thx!

agramfort:
thx @tomMoral !

PR #160, opened 2021-05-03
FIX: Fixing an issue with the dimension in test_benchmark_objective

bmalezieux:
I replaced np.zeros(*dimension) by np.zeros(dimension) in test_benchmark_objective to handle multidimensional shapes output by dataset.get_data().


agramfort:
This seems to have broken the CI

tomMoral:
the broken CI is unrelated.
It is an issue with env solving and julia that is present on master I believe: https://github.com/benchopt/benchOpt/runs/2490725536?check_suite_focus=true

PR #161, opened 2021-05-07
Update results page

tanglef:
This PR modifies for the results page:
- the index is in a grid
- the table contains main and subsidiary system informations (main=cpu/ram/cuda, sub=plateform/num_threads/processor) and displays them if available
- we can filter the table from the main informations (for when there will be a lot of files)
- informations + scipy/numpy versions appear if needed above plots that are up to date with modifications from PR that didn't make the cut to the main repo in the transfer.
- Responsive to different screen sizes + ok chrome / firefox

See the video attached to tell me where modifications are needed :slightly_smiling_face: 

https://user-images.githubusercontent.com/57089823/117458298-6dd44280-af4a-11eb-926f-792e8a72fcfc.mp4

### DIFF
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -69,6 +69,21 @@ def generate_plot_benchmark(df, kinds, fname, fig_dir, benchmark_name):
                 plot_func = globals()[PLOT_KINDS[k]]
                 try:
                     fig = plot_func(df_obj, plotly=True)
+                    if plot_func != "plot_histogram":
+                        if len(df_obj["solver_name"].unique()) < 10:
+                            fact_ = 10
+                        else:
+                            fact_ = 100
+                        height = 1000 + fact_ * len(objective_names)
+                        fig.update_layout(legend={"xanchor": "center",
+                                                  "yanchor": "top",
+                                                  "y": -.2,
+                                                  "x": .5
+                                                  },
+                                          autosize=False,
+                                          width=1000,
+                                          height=height
+                                          )
                 except TypeError:
                     fig = plot_func(df_obj)
                 figures[data_name][objective_name][k] = export_figure(
@@ -144,11 +159,14 @@ def get_results(fnames, kinds, root_html, benchmark_name, copy=False):
         datasets = list(df['data_name'].unique())
         main_info = ['system-cpus', 'system-ram (GB)', "version-cuda"]
         sub_info = ['platform', 'system-processor', 'env-OMP_NUM_THREADS']
+        ter_info = ["version-numpy", "version-scipy"]
         display_sysinfo = ["cpu", "ram (GB)", "cuda",
-                           "processor", "nb threads"]
+                           "processor", "nb threads",
+                           "numpy", "scipy"]
         main_dic = dict.fromkeys(main_info, "")
         sub_dic = dict.fromkeys(sub_info, "")
-        sysinfo = {"main": main_dic, "sub": sub_dic,
+        ter_dic = dict.fromkeys(ter_info, "")
+        sysinfo = {"main": main_dic, "sub": sub_dic, "ter": ter_dic,
                    "all_names": display_sysinfo}
         if "platform" in df:
             platform = (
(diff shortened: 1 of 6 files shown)


benchopt/plotting/generate_html.py
```diff
@@ -142,7 +157,35 @@ def get_results(fnames, kinds, root_html, benchmark_name, copy=False):
...
+            disp = 0
+            for level, all_keys in enumerate(
+                [main_info, sub_info[1:], ter_info]
+            ):
+                level = hierarchy[level]
+                for idx, key in enumerate(all_keys):
+                    val = df[key].unique()[0]
+                    if not pd.isnull(val):
+                        sysinfo[level][key] = {display_sysinfo[disp]: val}
+                    disp += 1
```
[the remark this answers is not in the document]
tanglef: Not really actually. I needed a criterion for backward compatibility.
So if there is the column `platform`  we know we're in recent versions and have all the keys. Only then we can go check them all and see if they have a value associated.
But I agree it's rough... When I'll make it into a function I'll try to do that better.

tanglef:
Thanks. I've put the button on top-right + some modifications. I'll work on "functionize" in the python file.
For the benchmark part, I'll see what I can do.
PS: in the last test, the error actually comes from Julia.

tomMoral:
thanks for the update.
for the failure with Julia, this is also on the main branch. this is due to
unstable install on conda I guess.


Le ven. 7 mai 2021 à 19:42, Tanguy Lefort ***@***.***> a
écrit :

benchopt/plotting/generate_html.py
```diff
@@ -142,7 +157,35 @@ def get_results(fnames, kinds, root_html, benchmark_name, copy=False):
 
         df = pd.read_csv(fname)
         datasets = list(df['data_name'].unique())
-
+        main_info = ['system-cpus', 'system-ram (GB)', "version-cuda"]
```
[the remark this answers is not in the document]
tanglef: See latest commit :slightly_smiling_face:

benchopt/plotting/generate_html.py
```diff
@@ -205,36 +205,39 @@ def get_sysinfo(df):
...
+                 ("version-cuda", 'cuda')
+                 ],
+        "sub": [('platform', 'platform'),
+                ('system-processor', 'processor'),
+                ('env-OMP_NUM_THREADS', 'nb threads')
+                ],
+        "ter": [("version-numpy", "numpy"),
+                ("version-scipy", "scipy")
+                ]
+    }
```
tomMoral, line 220: As it is a constant, maybe put it at the top of the file or before the function?

benchopt/plotting/html/templates/benchmark.mako.html
```diff
@@ -111,37 +109,37 @@ <h1>
...
-      % for idx_info, all_info in enumerate(result['sysinfo']['main'].values()):
-      <div class="all_info">
-        <div class="info">
-          % for idx in range(len(all_info)):
-          <b>${list(all_info)[idx]}</b>: ${list(all_info.values())[idx]}
+      <ul style="list-style-type: none; margin-top: 0;">
+      % for idx_info, (keys_info, vals_info) in enumerate(result['sysinfo']['main'].items()):
+        % if vals_info != "":
+          <li>
+            <b>${keys_info}</b>: ${vals_info}
```
tomMoral, line 120: You can remove the `enumerate` and this is only `key, val` no? not lists?
```suggestion
      % for name, val in result['sysinfo']['main'].items():
        % if val != "":
          <li>
            <b>${name}</b>: ${val}
```

tomMoral (approved):
A couple more comments but LGTM otherwise!!
Thx a lot @tanglef

benchopt/plotting/html/templates/benchmark.mako.html
tanglef: I'm currently looking for a way to remove this enumerate. Because I don't like that the + icon top-right of the cell to expand is index-dependant.

tanglef: Now it's better :slightly_smiling_face:

tomMoral (approved):
Not sure what was the problem but if you are satisfied with this, let's merge this one!

PR #164, opened 2021-05-12
[WIP] Fixes for results page

tanglef:
Observed issues:

- [ ] : index figures way too high (IOS)
- [ ] : navigation bar can be too large on benchmark page (screen size does not seem to be OS dependent)

Navigation bar issue seem fixable using percentages (need more testing for now cross platform).

### DIFF
--- a/benchopt/plotting/html/static/main.css
+++ b/benchopt/plotting/html/static/main.css
@@ -257,7 +257,7 @@ td.number {
 
 .sidenav {
 	height: 100%;
-	width: 200px;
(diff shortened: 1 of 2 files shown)


agramfort (approved):
Merge if happy. I will test when deployed

tomMoral:
The failure is due to `julia` still being unstable with `conda install`.
Let's not be too conservative on the `results` page for now and merge fast :)

Thx a lot @tanglef for the quick fix!

PR #167, opened 2021-05-28
Style mobile sidebar

tanglef:
A menu-type filter for mobiles that can replace the temporary fix-style.
The sidebar for non-mobile is not changed.

Before             |  After
:-------------------------:|:-------------------------:
Capture d’écran de 2021-05-28 15-01-27 | Capture d’écran de 2021-05-28 14-59-44

### DIFF
--- a/benchopt/plotting/html/static/main.css
+++ b/benchopt/plotting/html/static/main.css
@@ -268,6 +268,10 @@ td.number {
   transition: 0.5s;
 }
 
+.sidenav .closebtn{
+  display: none;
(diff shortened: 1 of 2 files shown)


agramfort (approved):

tomMoral (approved):
LGTM! thx @tanglef 

As a side note, I think it would be nice to add comments in the CSS to make sure this is maintainable in the long run. I guess simply describing which class is for which situation would make the file easier to understand.

PR #176, opened 2021-06-28
ENH add API for per solver stopping criterion

tomMoral:


benchopt/stopping_criterion.py
```diff
@@ -28,33 +40,110 @@ class StoppingCriterion():
...
+        )
 
-        # Store critical parameters for hashing and reconstruction
+        self.kwargs = kwargs
+        self.strategy = strategy
+
+    def get_runner_instance(self, max_runs=1, timeout=None, progress_str=None,
+                            solver=None):
+        """Copy the stopping criterion and set the parameters that depends on
+        how benchopt runner is callled.
```
agramfort, line 74:
```suggestion
        how benchopt runner is called.
```

benchopt/stopping_criterion.py
```diff
@@ -28,33 +40,110 @@ class StoppingCriterion():
...
+        max_runs : int
+            The maximum number of solver runs to perform to estimate
+            the convergence curve.
+        timeout : float
+            The maximum duration in seconds of the solver run.
+        progress_str : str or None
+            Format string to display the progress of the solver.
+        solver : BaseSolver
+            The solver for which this stopping criterion is called. Used to get
+            overridden ``stop_strategy`` and ``get_next``.
```
agramfort, line 87: Returns?

benchopt/stopping_criterion.py
```diff
@@ -63,6 +152,9 @@ def should_stop_solver(self, cost_curve):
 
         Parameters
         ----------
+        stop_val : int | float
+            Corresponds to stopping criterion of the underlying algorithm, such
+            as tol or max_iter.
```
agramfort, line 157:
```suggestion
            as ``tol`` or ``max_iter``.
```

agramfort (commented):
how do you want to proceed with the doc? 
I would first assess it works for the use case we have in mind before polishing the doc but there delay a bit the merge

agramfort (approved):
merge if happy @tomMoral

PR #177, opened 2021-07-01
Scales plotly option

tanglef:
- Add an option to toggle between semilogy plots and loglog with plotly (cc @mathurinm). It presents iself as a dropdown menu and the active element is showed. Default is loglog (current plots). (Of course after clicking it shrinks back to a normal menu, the image below is to show how options are displayed)
- Comments in CSS added for clarity.

Capture d’écran de 2021-07-01 16-57-16

### DIFF
--- a/benchopt/plotting/html/static/hover_index.css
+++ b/benchopt/plotting/html/static/hover_index.css
@@ -59,11 +59,15 @@
 .grid figure figcaption,
 .grid figure figcaption > a {
 	position: absolute;
-	top: 0em;
+	top: 0em; /* no space offset top */
 	left: 0;
 	width: 100%;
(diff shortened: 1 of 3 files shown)


mathurinm:
Thanks @tanglef it rocks. I tested it locally and it works as expected :tada: 

really really small comment, the box seems to overlap a bit with the title of the plot 
image

agramfort (approved):
nice !

tanglef:
Because the spacing for  the title is very "figure-dependant" I rather put the options on the side to avoid having to make ugly fixes for now.

tomMoral (approved):
LGTM.

Thanks a lot @tanglef for adding the comments!

Issue #178, opened 2021-07-02
ENH : refine solver name matching


mathurinm:
currently `-s celer` also selects `python-pgd[use_acceleration=True]` because it contains `celer`

PR #181, opened 2021-07-07
FIX pattern should *start* with solver or dataset name

mathurinm:
cf #178 
I think we could even change it to "^{p} {opening bracket or end of string}.*"   
at least with this change `-s celer` does not trigger `pgd[use_acceleration=False]"

WDYT @tomMoral ?

### DIFF
--- a/benchopt/tests/__init__.py
+++ b/benchopt/tests/__init__.py
@@ -11,7 +11,7 @@ DUMMY_BENCHMARK_PATH = TEST_BENCHMARK_DIR / 'dummy_benchmark'
 # Pattern to select specific datasets or solvers.
 SELECT_ONE_SIMULATED = r'simulated*500*rho=0\]'
 SELECT_ONE_PGD = r'python-pgd*step_size=1\]'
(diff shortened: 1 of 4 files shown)


tomMoral (approved):
LGTM, thx @mathurinm .

Could you simply add an entry in the `doc/whats_new.rst`?

benchopt/tests/__init__.py
```diff
@@ -11,6 +11,7 @@
 # Pattern to select specific datasets or solvers.
 SELECT_ONE_SIMULATED = r'simulated*500*rho=0\]'
 SELECT_ONE_PGD = r'python-pgd*step_size=1\]'
+SELECT_ONE_OBJECTIVE = r'dummy sparse regression*reg=0.1\]'
```
tomMoral: To avoid issues with spaces in the test:
```suggestion
SELECT_ONE_OBJECTIVE = r'dummy*reg=0.1\]'
```

tomMoral (commented):
for the failing test, I would say use this simple fix?

agramfort (approved):
thx @mathurinm

PR #186, opened 2021-07-14
add --profile option to run command

agramfort:
cc @PABannier

example of usage using OLS benchmark. The diff:

Screenshot 2021-07-14 at 11 06 10

then the usage:

Screenshot 2021-07-14 at 11 05 59

### DIFF
--- a/benchopt/cli/main.py
+++ b/benchopt/cli/main.py
@@ -75,7 +75,8 @@ main = click.Group(
 @click.option('--profile', 'do_profile',
               flag_value='True', default=False,
               help="Will do line profiling on all functions with @profile "
-                   "decorator.")
+                   "decorator. The profile decorator needs to be imported "
+                   "with: from benchopt.utils import profile")
 @click.option('--env', '-e', 'env_name',
               flag_value='True',
               help="Run the benchmark in a dedicated conda environment "
@@ -92,11 +93,11 @@ def run(benchmark, solver_names, forced_solvers, dataset_names,
         plot=True, html=True, pdb=False, do_profile=False,
         env_name='False'):
 
-    from benchopt.globals import globals
     from benchopt.runner import run_benchmark
+    from benchopt.utils.profiling import use_profile
 
-    # to set DO_PROFILE before calling validate_solver_patterns
-    globals["DO_PROFILE"] = do_profile
+    if do_profile:
+        use_profile()  # needs to be called before validate_solver_patterns
 
     # Check that the dataset/solver patterns match actual dataset
     benchmark = Benchmark(benchmark)
@@ -111,7 +112,8 @@ def run(benchmark, solver_names, forced_solvers, dataset_names,
             dataset_names=dataset_names,
             objective_filters=objective_filters,
             max_runs=max_runs, n_repetitions=n_repetitions,
-            timeout=timeout, plot_result=plot, html=html, pdb=pdb
+            timeout=timeout, plot_result=plot, html=html, pdb=pdb,
+            do_profile=do_profile
         )
         return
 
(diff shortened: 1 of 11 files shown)


tomMoral (commented):
Not a big fan of the globals. I propose another way to do it.
Let me know if you feel that I am nitpicking too much, I tend to do that sometime.

It is just that I don't really like the idea of a global dictionary.
Also, I feel this is a bit brittle because if `profile` is imported before the set of the glolbal variable, this can fail. The proposed solution should be robust in any cases as the decision is taken when the decorator is called.

benchopt/profiling.py
```diff
@@ -0,0 +1,14 @@
...
+    profile = line_profiler.LineProfiler()
+else:
+    class Profile:
+        def __call__(self, func):
+            return func
+
+        def print_stats(self):
+            return
+
+    profile = Profile()
```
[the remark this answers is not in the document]
agramfort: Works for me. Don’t hesitate to push here.

benchopt/runner.py
```diff
@@ -497,4 +504,8 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
     if plot_result:
         from benchopt.plotting import plot_benchmark
         plot_benchmark(save_file, benchmark, html=html)
+
+    if do_profile:
+        print_stats()
```
tomMoral, line 509: You can simply call this as the Profiler object will only be created if `do_profile` is `True` and at least one function as been decorated.
```suggestion
    print_stats()
```

benchopt/runner.py
```diff
@@ -402,6 +406,9 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
     """
     print("BenchOpt is running")
 
+    if do_profile:
+        use_profile()
+
```
tomMoral, line 411: Do you need to call this again?

tomMoral (commented):
Just nitpicking but I am not sure we need the `do_profile` in the `run_benchmark` function.

agramfort:
Yes I can simplify if we assume benchopt is only used from CLI. I can then remove the do profile parameter. Shall I do this ?

tomMoral:
I would say yes

agramfort:
@tomMoral let me know how you would add a test for this.

benchopt/cli/main.py
```diff
@@ -85,9 +91,14 @@
               "datasets, see the command `benchopt install`.")
 def run(benchmark, solver_names, forced_solvers, dataset_names,
         objective_filters, max_runs, n_repetitions, timeout,
-        plot=True, html=True, pdb=False, env_name='False'):
+        plot=True, html=True, pdb=False, do_profile=False,
+        env_name='False'):
 
     from benchopt.runner import run_benchmark
+    from benchopt.utils.profiling import use_profile
```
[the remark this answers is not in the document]
tomMoral: WDYT @agramfort ?

benchopt/cli/main.py
```diff
@@ -104,6 +116,10 @@ def run(benchmark, solver_names, forced_solvers, dataset_names,
             max_runs=max_runs, n_repetitions=n_repetitions,
             timeout=timeout, plot_result=plot, html=html, pdb=pdb
         )
+
+        if do_profile:
+            print_stats()
```
tomMoral, line 121: You can also simply do as it will only print stat if they have been recorded.
```suggestion
        print_stats()
```

tomMoral (commented):
For the test, I would add the `profile` decorator to one of the  solver in the dummy benchmark and add a test in `test_cli.py` similar to `test_benchopt_run_in_env`, where you check that with `--profile` you have a profiler report and that without it, you don't?

agramfort:
@tomMoral it's a mystery... it seems the USE_PROFILE is reset back to False although it's set to True first....

agramfort:
Merge if happy @tomMoral

tomMoral:
Merged! thx @agramfort !

PR #190, opened 2021-07-21
larger markers, use semilogy scale by default

mathurinm:
1) I changed the default scale to semilogy in plotly. I find loglog very confusing, as it exaggerates the importance of early iterations ; it also does not reflect how crushingly superior an algorithm can be.

2) I increased the marker scales, as I found them hard to read (given that the colors are already quite similar)

Before marker size change (notice how the two red curves are hard to distinguish):
image

After:
image


agramfort (approved):

tomMoral (approved):
Seems nice!

PR #191, opened 2021-07-21
Fix #189

### LINKED ISSUE #189: xlim does not seems to always be updated when solvers are selected/deselected
Sometimes unticking a solver changes the x axis, sometimes it does not (e.g:
image

reproduce:: select the first config in the following report, tick and untick PGD: xlim changes. removing more solvers does not reduce the xlim.

https://www.dropbox.com/s/kicnhajwt3jwgov/benchmark_bcd_benchopt_run_2021-07-21_10h25m04.html?dl=0

tanglef:
This is a fix for the issue #189 about the xlim that did not really adapt to the selected curves because of the horizontal dotted black line.

### DIFF
--- a/benchopt/plotting/helpers_compat.py
+++ b/benchopt/plotting/helpers_compat.py
@@ -56,11 +56,9 @@ def add_h_line(fig, val, xlim=None, plotly=False):
         plt.xlim(xlim)
         return fig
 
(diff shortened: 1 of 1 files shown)


mathurinm:
I pulled the PR and it works as expected. thanks a lot @tanglef

PR #199, opened 2021-09-07
update index

agramfort:


agramfort:
here is how it looks now:

https://943-227720375-gh.circle-artifacts.com/0/dev/index.html#some-available-benchmarks

PR #200, opened 2021-09-07
ENH multi objective plotting

tomMoral:
Add a prefix to all `objective` outputs and then print all of them so we can have multiple curves.


agramfort (approved):
@tomMoral merge if happy

PR #202, opened 2021-09-08
Fix doc CI

tanglef:
Circle CI doc build fails. 

It would seem that the support for the parameter `use_2to3` was dropped in the last version (58.0) of setuptools (as it was deprecated for some time now). But sphinx-bootstrap uses it.
Setuptools developers are working on some fixes since the release, but in the mean time maybe we should pin the version to build the doc.


agramfort:
thx @tanglef

PR #207, opened 2021-09-14
Reverse setuptools version

tanglef:
Now that sphinx-bootstrap-theme has released the version 0.8.0 we should no longer need to pin setuptools version.
(pending that other packages don't use the `use_2to3` parameter).


agramfort:
CIs are red

tomMoral (approved):
LGTM! thx @tanglef

PR #208, opened 2021-09-14
Remove inf values from plotting

tanglef:
The reason why the `logreg_l2` plots were "weird looking" was that there was a point that took a very long time to be computed that had for value `inf`: so not visible.
By removing the `inf` values in the plots, axis scales are much more relevant (see before/after)

| Before | After  |
|---|---|
| With inf values|Without inf values|

### DIFF
--- a/benchopt/plotting/plot_objective_curve.py
+++ b/benchopt/plotting/plot_objective_curve.py
@@ -44,8 +44,7 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
     dataset_name = df['data_name'].unique()[0]
     objective_name = df['objective_name'].unique()[0]
     title = f"{objective_name}\nData: {dataset_name}"
-
-    df.query(f"{obj_col} not in [inf, -inf]", inplace=True)
+    df.query(f"`{obj_col}` not in [inf, -inf]", inplace=True)
     y_label = "F(x)"
     if suboptimality:
         eps = 1e-10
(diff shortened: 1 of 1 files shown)


agramfort:
thx @tanglef

it seems it breaks the CIs. Can you have a look?

tanglef:
> thx @tanglef
> 
> it seems it breaks the CIs. Can you have a look?

It seems that the whole CI (not just this PR) is broken on Macos (and cancels the others)

agramfort:
you have time to investigate?

>

tanglef:
I can try but that won't be today. Also, it seems that its spreading to other repos.

agramfort (approved):
@tomMoral merge if happy

benchopt/plotting/plot_objective_curve.py
```diff
@@ -45,6 +45,8 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
     objective_name = df['objective_name'].unique()[0]
     title = f"{objective_name}\nData: {dataset_name}"
 
+    inf = float("inf")
+    df = df[df[obj_col] != inf]
```
tomMoral, line 49: Simpler and also handeling the negative case:
```suggestion
    df = df.query(f"{obj_col} not in [inf, -inf]")
```

tomMoral (approved):
A small nitpick

tanglef:
I will look at why this is red

tomMoral:
ok sorry, this is once again a `pandas` instability with queries and numexpr.
revert to your initial solution and let's merge it that way...

tanglef:
This should work, if not I'll revert back

tomMoral:
GReat! thx for the fix @tanglef

PR #209, opened 2021-09-14
descending order based on date

agramfort:
for some reason this does not work

help welcome @tanglef 

basically I want to see the latest results appear first

### DIFF
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -233,10 +233,10 @@ def get_sysinfo(df):
         if key in df:
             if key == 'platform':
                 return (
-                        df["platform"].unique()[0] +
-                        df["platform-release"].unique()[0] + "-" +
-                        df["platform-architecture"].unique()[0]
-                    )
+                    df["platform"].unique()[0] +
+                    df["platform-release"].unique()[0] + "-" +
+                    df["platform-architecture"].unique()[0]
+                )
             else:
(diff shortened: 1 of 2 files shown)


benchopt/plotting/generate_html.py
```diff
@@ -495,7 +495,7 @@ def plot_benchmark_html_all(patterns=(), benchmarks=(), root=None,
         fnames = []
         for p in patterns:
             fnames += (benchmark / 'outputs').glob(f"{p}.csv")
-        fnames = sorted(set(fnames))
+        fnames = sorted(set(fnames))[::-1]
```
tanglef, line 498: The elements are not sorted here but in the JS table because of jquery/aria anyway, so this can be removed

tanglef (changes requested):
To do that, we actually need to touch the JS in `benchmark.mako.html`.
I will do a PR on your branch with the changes needed

agramfort:
can you have a look?

we try to push these days the final touches to benchopt to push a new
release
that should be the first stable release.

thx !

>

tanglef:
I can't push directly on your branch so I made a PR (https://github.com/agramfort/benchOpt/pull/1) on your forked version of the repo. If that is ok for you, you can merge it on your branch and it will be all good here.

tanglef:
And the CI seem to be good today :)

agramfort:
works like a charm @tanglef  thx heaps !

PR #210, opened 2021-09-27
FIX use conda install for sklearn

tomMoral:
try to fix failures from https://github.com/benchopt/benchOpt/runs/3718219572?check_suite_focus=true
If this works, we will try to reproduce the segfault for `sklearn` in a separate PR.


agramfort (approved):
🍺

PR #211, opened 2021-09-28
DBG reproduce sklearn segfault

tomMoral:
Installing scikit-learn 1.0 from pip on OSX with julia and R seems to cause a segfault at import.
This PR tries to pin point why.


tomMoral:
This one has been solved in the nightly build of sklearn. closing this.

PR #214, opened 2021-10-06
Plolty colors match matplotlib

### LINKED ISSUE #213: ENH/BUG: choose solver colors on a wider range
For 4 solvers we already have 2 red ones:
image

I don't understand why this happens. When we pick the color we use CMAP(i): https://github.com/benchopt/benchOpt/blob/master/benchopt/plotting/plot_objective_curve.py#L84

If I show this palette outside of benchopt, it looks good and colors 3 and 4 are orange, not red. 
image

tanglef:
Colors in plotly need to be RGBa to match with matplotlib (*255).
(Fixes #213 @mathurinm)


agramfort (approved):

agramfort:
looks good
https://benchopt.github.io/results/benchmark_lasso_benchmark_lasso_benchopt_run_2021-09-13_04h21m53.html
!

>

PR #218, opened 2021-10-06
Constant trace selector

### LINKED ISSUE #215: Selected solvers are not constant across curves types
Changing the type of curves forgets about the currently selected solvers (and displays the solvers which were selected the last time this curve was displayed).

I think it may be more intuitive to share selected solvers across plot types

tanglef:
Fixes #215 
Traces are the same for objectives, objective values, type of axis and kind of plot (histogram is not modified).
They only reinitialize themselves when changing dataset.

### DIFF
--- a/benchopt/plotting/html/templates/result.mako.html
+++ b/benchopt/plotting/html/templates/result.mako.html
@@ -63,7 +63,7 @@
 
 <!-- Add selector to display only one plot -->
 <label for='dataset_selector'>Dataset:</label>
-<select id='dataset_selector' onchange="showMe(this);">
+<select id='dataset_selector' onchange="showMe(this); set_visibility();">
 % for data in result['dataset_names']:
   <option value="${data}">${data}</option>
 %endfor
@@ -170,6 +170,9 @@
 
   var obj = document.getElementById('dataset_selector');
   obj.onchange();
+  var prev_dataset = obj.value;
+  var obj = document.getElementById("objective_selector");
+  obj.onchange();
   var prev_obj_selector = obj.value;
   var obj = document.getElementById('objective_column');
   obj.onchange();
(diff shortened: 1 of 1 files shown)


benchopt/plotting/html/templates/result.mako.html
```diff
@@ -169,10 +170,37 @@ <h1>Result on ${benchmark} benchmark</h1>
...
+    var kind = document.getElementById('plot_kind').value;
+    var obj_select = document.getElementById("objective_selector").value;
+    var obj_col = document.getElementById("objective_column").value;
+    if(kind !== "histogram"){
+      if(prev_kind != null){
+        // old graph values
+        var graph_id = document.querySelector('[id$="' + prev_obj_selector + prev_obj_col + prev_kind + '"]').id;
+        var graph = document.getElementById(document.getElementById(graph_id).getElementsByTagName("div")[1].id);
+        var traces_visib = graph.data.map(trace => trace.visible);
+        // new graph values
```
tomMoral, line 191:
```suggestion
        // Set the visible traces to the one from the previous graph.
        // This part ensures that the visible solvers are always the
        // across the different plots (same dataset).
```

benchopt/plotting/html/templates/result.mako.html
```diff
@@ -169,10 +170,37 @@ <h1>Result on ${benchmark} benchmark</h1>
...
+        var graph_id = document.querySelector('[id$="' + obj_select + obj_col + kind + '"]').id;
+        var graph = document.getElementById(document.getElementById(graph_id).getElementsByTagName("div")[1].id);
+        for(i=0; i < graph.data.length; i++){
+          graph.data[i].visible = traces_visib[i];
+        }
+        Plotly.redraw(graph);
+      }
+      prev_kind = kind;
```
tomMoral, line 199: A suggestion: maybe store the graph_id so it is simpler code to read?
Not mandatory of course.

```suggestion
      // Store the current graph id so we can update the visible traces on the next change.
      prev_kind = kind;
```

tomMoral:
There is a problem when you change the dataset. The behavior you implemented is fine (it resets to all solvers are visible).
However, if you change the dataset, `set_visible` does not pick up the right map because the doc id is computed independently of the dataset.

I guess you need to store the dataset id and update it when changing dataset.

tanglef:
Now this should work. Changing dataset sets all traces to visible.
Also thanks, but I assure you, I'm no JS wizard, I started learning JS for Benchopt :sweat_smile:
And I think one day I'll have to learn how to use jquery properly to make it way less time consuming to load a result page :slightly_smiling_face:

PR #223, opened 2021-11-03
Add axis types semilog and linear

### LINKED ISSUE #222: ENH: add normal scale and semilogx as axis types
For some objectives it may make sense to have a normal plot, without logscale (here I have vanishing values in my custom objective) 
image

we may also add a semilogx option. 

@tanglef the plotly :mage:

tanglef:
Fixes #222 by adding semilog-x and linear graph axis types (poke @mathurinm )

### DIFF
--- a/benchopt/plotting/html/static/result.js
+++ b/benchopt/plotting/html/static/result.js
@@ -101,9 +101,19 @@ function changeScale(e, allIds) {
     switch (e.value) {
       case "loglog":
         layout.xaxis.type = "log";
+        layout.yaxis.type = "log";
         break;
       case "semilog-y":
         layout.xaxis.type = "linear";
+        layout.yaxis.type = "log";
+        break;
+      case "semilog-x":
+        layout.xaxis.type = "log";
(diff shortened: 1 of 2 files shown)


mathurinm:
Looking good! Thx @tanglef 
image
image

agramfort:
merge if happy

>

mathurinm:
Is it easy to keep the same `axis type` selected as long as `value` (here, objective value and support size) does not change ? 

Currently when I change Objective, it goes back to semilogy scale

tanglef:
You are right, but that means there's a bigger flaw about the current (and past) implementations.
The axis type is the same across "kinds" but the dropdown menu still changes when we change the type.
I will look into that and find a way to make it right for objectives too.

tanglef:
The way I found to make the axis type "independent" was to remove the id list made from Python that was passed to Js through the `onChange` method and update directly the plots concerned using Js `onChange`.

tomMoral (approved):
This LGTM! Thanks a lot @tanglef ! :)

PR #227, opened 2021-11-04
Display options with newline and temporary ci fix setuptools 

tanglef:
- Display options for BenchOpt results a little more organized (poke @mathurinm)
- `setuptools` released the version 58.5.0 that broke the CI (and a lot of others like scipy's), like last time until they fix the problem we can pin the version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,6 +57,7 @@ jobs:
             name: Get Python running
             command: |
               mamba install python=3.8 julia r-base rpy2 numpy cython -yq
+              pip install --upgrade "setuptools<58.5"
               pip install --upgrade --progress-bar off julia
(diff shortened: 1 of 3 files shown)


mathurinm:
Tested, looks good to me:
image


Maybe it's the occasion to have something more explicit than "Kind"? What do you think of "Plot type", or just "Plot"?

tanglef:
Is `chart type` ok? (naming convention in plotly's doc)

tanglef:
The result:
Capture d’écran de 2021-11-04 09-50-31

tanglef:
And it's also the occasion to remove all underscores in titles (and the benchmark that was repeated) and buttons 
Capture d’écran de 2021-11-04 10-06-05

tomMoral (approved):
LGTM!  The change in the UI are great

Also, thx for the CI fix. hopefully we can remove the deps pin quickly.

PR #231, opened 2021-11-05
Multiple level calls in html generation CLI

### LINKED ISSUE #230: benchopt generate-results produces html giving a File not found
```
(base) ➜  lasso git:(flash) ✗ pwd
/home/mathurin/workspace/benchmarks/lasso
(base) ➜  lasso git:(flash) ✗ benchopt generate-results -b . 
Rendering benchmark: .
Processing outputs/benchopt_run_2021-11-05_08h30m08.csv
Processing outputs/benchopt_run_2021-11-05_08h30m38.csv
Processing outputs/benchopt_run_2021-11-05_08h34m48.csv
Solver sklearn did not reach precision 1e-06.
Processing outputs/benchopt_run_2021-11-05_08h36m28.csv
Processing outputs/benchopt_run_2021-11-05_09h05m37.csv
Solver sklearn did not reach precision 1e-06.
Solver sklearn did not reach precision 1e-06.
Solver sklearn did not reach precision 1e-06.
Solver sklearn did not reach precision 1e-06.
Solver sklearn did not reach precision 1e-06.
Solver sklearn did not reach precision 1e-06.
Writing  results to html.html
Writing results to html/__benchopt_run_2021-11-05_08h30m08.html
Writing results to html/__benchopt_run_2021-11-05_08h30m38.html
Writing results to html/__benchopt_run_2021-11-05_08h34m48.html
Writing results to html/__benchopt_run_2021-11-05_08h36m28.html
Writing results to html/__benchopt_run_2021-11-05_09h05m37.html
Writing index to html/index.html
```

This opens a valid html:
image
but if I try to click on the benchmarks rectangle, I get:
image

tanglef:
Fixes #230 by allowing users to call `generate-results` from multiple levels.
For example:

benchmarks
├─ benchmark_lasso <- we're inside here
├─ benchmark_logreg_l1

you can do `benchopt generate-results -b . -b ../benchmark_logreg_l1` and get a menu page with each benchmark which will not lead to 404 error page (poke @mathurinm).


agramfort:
@mathurinm merge if you're happy

tomMoral (changes requested):
Thanks a lot for the proposed fix. But I think we should change this function to rely on the `Benchmark` object so we avoid accessing folders manually. This requires a bit more changes, I can do it if you don't have time.

tanglef:
So, I didn't respond to take time to understand what the task really would mean using `Benchmark`, and I'm not sure I totally get it.
You would want to use said `Benchmark` object in a common framework for local and the website renderings, but the `results` repository is not made of `Benchmarks`, only folders with csv files  (no `objective.py`, ...) so I'll turn this PR into draft for now and try and work on it when I have the time

tomMoral:
Ok I see what you mean now. I would proceed that way:

- add an optional argument in `Benchmark` to not check that `objective.py` exists so you can still create the `Benchmark` object in `results`
- Use a list of `Benchmark` in both cases with `get_results_files` to list all files.

We might want to update a bit the `publish` command to also select the folder name from `get_output_folder`.

tanglef:
- [x] optional argument `standalone` in `Benchmark` to use a Benchmark even if no `objective.py` file exist
- [x] use a list of `Benchmark` (without `get_results_files` because of the pattern arg) using `benchmark.benchmark_dir` as anchor
- [ ] take into account possible duplicates results (cf #263, but how as there is possibly no `objective.py` target id -whatever we choose it to be- to match with ?)  
- [ ] update the publish command

tanglef:
The command `benchopt generate-results -b .` works with current changes and I think the `publish` and multiple repo names problem should be addresses in another PR with more discussion (out of the current scope)

tanglef:
@tomMoral if it's good for you, it's good for me. I updated with recent modifications. And all checks are good now

benchopt/benchmark.py
```diff
@@ -27,36 +27,37 @@
...
-    benchmark_dir : str or Path-like
-        Folder containing the benchmark. The folder should at least
-        contain an `objective.py` file defining the `Objective`
-        function for the benchmark.
+        Parameters
+        ----------
+        benchmark_dir : str or Path-like
+            Folder containing the benchmark. The folder should at least
+            contain an `objective.py` file defining the `Objective`
+            function for the benchmark.
```
mathurinm, line 38: @tanglef can you document the `standalone` parameter?

benchopt/benchmark.py
```diff
@@ -27,36 +27,40 @@
...
-    def __init__(self, benchmark_dir):
+        Parameters
+        ----------
+        benchmark_dir : str or Path-like
+            Folder containing the benchmark. The folder should at least
+            contain an `objective.py` file defining the `Objective`
+            function for the benchmark.
+        standalone : boolean
+            If no `objective.py` file defines the benchmark, use
+            `standalone=True` in order to visualize available results.
```
mathurinm, line 41: @tomMoral it's the `click` parameter right?

tanglef:
Conflicts with `main` solved

mathurinm (approved):

tomMoral:
I don't think this is correct, what is the difference between standalone and `allow_meta_from_json`?

Sorry I feel like I reimplemented a bit the same logic on the meta data (I worked on this in the context of the plotting config, I did not realize it was similar).

But I am not sure it makes sense to have both?

tanglef:
> I don't think this is correct, what is the difference between standalone and `allow_meta_from_json`?
> 
> Sorry I feel like I reimplemented a bit the same logic on the meta data (I worked on this in the context of the plotting config, I did not realize it was similar).
> 
> But I am not sure it makes sense to have both?

Sorry, I didn't follow this feature release. I just tested it out and the multiple level calls works with `allow_meta_from_json`, just have to remove this part in `benchmark.py` and replacethe `standalone=True` with your `allow_meta_from_json=True`.
Should I do this in another PR ?

tomMoral:
no worries, sorry I did a second implementation instead of merging this one.

a follow up PR would be nice yes

PR #232, opened 2021-11-15
Alternative to remove local file 

### LINKED ISSUE #229: ENH: add a way to delete benchmarks results from html page ?
I tend to have quite many outputs in the end, maybe we could have a `trash` icon on every line  in the right most column that would delete them and the associated csv (after prompting for confirmation) ?
image

Also, when I delete the benchmarks manually (html and csv), they still appear on `lasso.html`.

tanglef:
As discussed to resolve #229, for local rendering only, this adds a select box at the left of the bench result's name (or just below depending on the available space) and a trash icon below the table.

Capture d’écran de 2021-11-15 09-36-42

If selected, then the trash icon displays the path of several files that are inside the `outputs` folder:
- the selected `.csv` datasets,
- the selected `.html` associated generated pages,
- the cached json file (so the next time we call `benchopt plot --all .` for example the deleted benchs are not presented in the results table).

Capture d’écran de 2021-11-15 09-36-50

Then, one can simply click on `copy to clipboard` and paste the command in the terminal from the `outputs` folder.
For the current time, there is also a `hide row` button that will hide the selected rows (until the page is refreshed as it is static) but it can be useful to avoid clutter if we open the results in multiple tabs. 

PS: this also separates the JS for the table page from the html. The only thing to adapt in the future would be the paths of the files (available as `data-csv` and `data-html` attributes of the checkboxes) when the user does not use `benchopt plot` but `generate-results` or a local `make render` (open issue #230).

### DIFF
--- a/benchopt/plotting/html/static/benchmark.js
+++ b/benchopt/plotting/html/static/benchmark.js
@@ -1,0 +1,152 @@
+$(document).ready(function () {
+  $(".summary").dataTable({
+    order: [[0, "desc"]],
+  }); // reorder table by date
+  if (location.hostname === "") {
+    // local file
+    $("[name='checkfiles']").css({
+      display: "block",
+      "margin-left": "auto",
+      float: "left",
+    }); // display checkboxes
+    $("#trashBtn").css({
+      display: "block",
+      float: "left",
+      "margin-bottom": "5vh",
+    });
+  }
+});
+
+$("table").each(function (a, tbl) {
+  var currentTableRows = $(tbl).find("tbody tr").length;
+  $(tbl)
+    .find("th")
+    .each(function (i) {
+      var remove = 0;
+      var currentTable = $(this).parents("table");
+      var tds = currentTable.find("tr td:nth-child(" + (i + 1) + ")");
+      tds.each(function (j) {
+        if ($(this)[0].innerText.length == 0) remove++;
+      });
+      if (remove == currentTableRows) {
+        $(this).hide();
+        tds.hide();
+      }
+    });
+});
+
+function change(ll_item) {
+  var td, i, fil;
+  var filter = new Array();
+  for (item = 0; item < ll_item.length; item++) {
+    filter.push(
+      document
+        .getElementById("select" + ll_item[item].replace(/\s/g, ""))
+        .value.toUpperCase()
+    );
+  }
+  var table = document.getElementById("summary");
+  var tr = table.getElementsByTagName("tr");
+  for (i = 0; i < tr.length; i++) {
+    tr[i].style.display = "";
+    td = tr[i].getElementsByTagName("td")[2];
+    if (td) {
+      for (fil = 0; fil < filter.length; fil++) {
+        if (
+          td.innerHTML.toUpperCase().indexOf(filter[fil]) > -1 &&
+          tr[i].style.display !== "none"
+        ) {
+          tr[i].style.display = "";
+        } else {
+          tr[i].style.display = "none";
+        }
+      }
+    }
+  }
+}
+
+function displaymore(id, loop_index) {
+  var x = document.getElementById("subinfo" + loop_index);
+  $(id).find("svg").toggleClass("fa-plus-circle fa-minus-circle");
+  if (x.style.display === "none") {
+    x.style.display = "block";
+  } else {
+    x.style.display = "none";
+  }
+}
+
+var openbtn = document.getElementById("open");
+if (openbtn) {
+  var closebtn = document.getElementById("close");
+  openbtn.style.display = "none";
+
+  function openSideNav() {
+    navbar = document.getElementById("sidenav");
+    navbar.style.visibility = "visible";
+    main = document.getElementById("main");
+    if ($(window).width() < 900) {
+      navbar.style.height = "100%";
+      main.style.marginLeft = "0";
+    } else {
+      navbar.style.height = "1000px";
+      main.style.marginLeft = "200px";
+    }
+    navbar.style.opacity = "1";
+    openbtn.style.display = "none";
+    closebtn.style.display = "";
+  }
+
+  if ($(window).width() > 900) {
+    openSideNav();
+  } else {
+    closeSideNav();
+  }
+
+  function closeSideNav() {
+    navbar = document.getElementById("sidenav");
+    main = document.getElementById("main");
+    navbar.style.visibility = "hidden";
+    navbar.style.opacity = "0";
+    navbar.style.height = "0";
+    if ($(window).width() > 900) {
+      main.style.marginLeft = "0px";
+    }
+    openbtn.style.display = "";
+    closebtn.style.display = "none";
+  }
+}
+
+$(function () {
+  $("#dialogRm").dialog({
+    autoOpen: false,
+  });
+
+  $("#trashBtn").click(function () {
+    allChecked = document.querySelectorAll("input[name=checkfiles]:checked");
+    delCmd = "rm \\\n <br />"; // n and br for html and copy to clipboard
+    for (check of allChecked) {
+      delCmd += $(check).attr("data-csv") + " \\\n <br />";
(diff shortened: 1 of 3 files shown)


agramfort:
I am +0 on this

I'll let vote those we feel the need for this

tomMoral:
Forgot to mention but I like the feature! I am +1 on this one!

tanglef:
Yes sure. I haven't really had the time to go back on BenchOpt these past few days but I'll take care of it ASAP.

tomMoral:
Thanks for the fix in `.circleci` config!

Just a comment for next time, please try to fix such issue in a separate PR so it is easier to review and merge each task.

tanglef:
If you prefer, I can revert the last commit and open a new PR with the `.circleci` fix now.

mathurinm:
No need to revert it, you can open a new pr, merge it into main and merge main into this one

benchopt/plotting/html/static/benchmark.js
```diff
@@ -0,0 +1,175 @@
...
+      }
+    }
+  }
+}
+
+/**
+ * Display subinfo in the table
+ * Toggle between + and - icon button
+ */
+function displaymore(id, loop_index) {
```
tomMoral, line 79: For consistent naming:
```suggestion
function displayMore(id, loop_index) {
```

benchopt/plotting/html/static/benchmark.js
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// dialog box to get path of files we would like to remove
+$(function () {
+  $("#dialogRm").dialog({
+    autoOpen: false, // only open when needed
+  });
+
+  // click on delete button in the dialog box
+  $("#trashBtn").click(function () {
```
tomMoral, line 143: To be consistent, I would rather use the `onclick` attribute of the HTML button and define this one as a callback?

tomMoral (commented):
Thanks a lot for all the comments!!
Reading this, I can understand what each part is doing. I tried to make the  some of the comments even clearer.

I have a few questions on how to get a consistent organization of the JS file, in particular:

- should we move all parts related to post load formatting of the page to the `ready` callback?
- Should we consistently use callback with defined names and not jquery?

I am by no mean a JS expert so if you have arguments against this, no problem. My arguments are mainly for consistency and maintainability.

tanglef (commented):
Apply some doc changes

benchopt/plotting/html/static/benchmark.js
tanglef: In [tfae2072 commit](https://github.com/benchopt/benchopt/pull/232/commits/fae20727268085a2cac52e5ef78fa99bbe1bd33d)

benchopt/plotting/html/static/benchmark.js
tanglef: It should be the preferred way to handle click events actually (recently learned that)
See [this as ref and more explanations](https://stackoverflow.com/questions/12627443/jquery-click-vs-onclick)

benchopt/plotting/html/static/benchmark.js
```diff
@@ -0,0 +1,175 @@
...
+    $("#trashBtn").css({
+      display: "block",
+      float: "left",
+      "margin-bottom": "5vh",
+    });
+  }
+});
+
+// hide columns without content (if system informations not available)
+$("table").each(function (a, tbl) {
```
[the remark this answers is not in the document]
tanglef: Done you're right

benchopt/plotting/html/static/benchmark.js
```diff
@@ -0,0 +1,175 @@
...
+  var x = document.getElementById("subinfo" + loop_index);
+  $(id).find("svg").toggleClass("fa-plus-circle fa-minus-circle");
+  if (x.style.display === "none") {
+    x.style.display = "block";
+  } else {
+    x.style.display = "none";
+  }
+}
+
+// handle sidebar navigation
```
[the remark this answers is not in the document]
tanglef: Done with the update on the click method too

benchopt/plotting/html/static/benchmark.js
tomMoral: Do you think i would be possible to define the function inside `click` outside and add the callback as it is done for the open and close button?

benchopt/plotting/html/templates/benchmark.mako.html
```diff
@@ -145,7 +147,7 @@ <h1>
       % if disp_sub:
         <!--Case where sub infos are available but somehow main infos are not-->
         % if disp_button:
-          <button onclick="displaymore(this,${idx_res})" id=${"btn_subinfo"+str(idx_res)} class="button buttoncent" style="float:right;">
+          <button onclick="displayMore(this,${idx_res})" id=${"btn_subinfo"+str(idx_res)} class="button buttoncent" style="float:right;">
```
tomMoral, line 150: should we also use the `click` method or would this complicate the code too much?

benchopt/plotting/html/static/benchmark.js
```diff
@@ -0,0 +1,209 @@
+/*********************************************
+ *  JS functions for the benchmark page.
+ *********************************************/
+
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* Format the table once the page has been loaded.
+* - Order the table based on run date.
+* - Add trash button if it is a local file.
```
tomMoral, line 8:
```suggestion
* - Add trash button if it is a local file.
* - Hide sysinfo if the column is empty
```

benchopt/plotting/html/static/benchmark.js
```diff
@@ -0,0 +1,209 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+$(function () {
+  // Sort the table by descending date order (date being in column 0)
+  $(".summary").dataTable({
+    order: [[0, "desc"]],
+  }); // reorder table by date
+
+  if (location.hostname === "") {
```
tomMoral, line 16:
```suggestion
  /*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  * Add trash button if this is a local document
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
  if (location.hostname === "") {
```

benchopt/plotting/html/static/benchmark.js
```diff
@@ -0,0 +1,209 @@
...
+    }); // display checkboxes
+    $("#trashBtn").css({
+      display: "block",
+      float: "left",
+      "margin-bottom": "5vh",
+    });
+  }
+  /*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  * Hide the system information column if it is empty
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
```
tomMoral, line 32:
```suggestion

  // Hide the system information column if it is empty
```

benchopt/plotting/html/templates/benchmark.mako.html
tomMoral: Maybe with something like `data-index` to store the `idx_res`, as describe [here](https://stackoverflow.com/a/53047367/2642845).

tanglef: Done in the latest commit (and we were already using the `data-?` attributes to handle the filenames :slightly_smiling_face: so the code is quite the same).

tomMoral (approved):
LGTM!

The Js is cleaner this way. I will merge this.

Just a leftover question for later: Is it a good practice to have many calls to document ready. not sure about the perf here. I could not find any clear answer to this.

PR #236, opened 2021-11-23
CLN fix print length

tomMoral:
Try to fix the length of the output.
What is missing is actually try to reduce the size of the name of the solver if the console is too small.

An idea would be to `truncate` the parameters with ellipsis: `|---solver[p=1,...] xx% (X / X reps)`


agramfort (approved):

tomMoral:
Let's merge this one as it will fix 90% of the use cases. We can always look for truncated solver names later if needed.

PR #240, opened 2021-11-24
DOC: use less solvers, use suboptimality

mathurinm:
The first figure that users see when they land on the doc page is a bit "heavy" in my opinion, so I removed some solvers, and use a suboptimality plot:
Before:
image

After (on my computer, I'll wait to see the results on CI):
image

### DIFF
--- a/examples/plot_run_benchmark.py
+++ b/examples/plot_run_benchmark.py
@@ -19,7 +19,8 @@ BENCHMARK_PATH = (
 
 try:
     save_file = run_benchmark(
-        Benchmark(BENCHMARK_PATH), ['sklearn', 'lightning'],
+        Benchmark(BENCHMARK_PATH), ['sklearn[liblinear]', 'sklearn[saga]',
+                                    'sklearn[newton-cg]', 'lightning'],
         dataset_names=['Simulated*n_samples=200,n_features=500*'],
         max_runs=100, timeout=20, n_repetitions=3,
         plot_result=False, show_progress=False
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -25,10 +25,10 @@ optimization benchmark should be **as simple as doing**:
     $ benchopt run --env ./benchmark_logreg_l2
 
 Running these commands will fetch the benchmark files, install the benchmark
-requirements in a dedicated environment `benchop_benchmark_logreg_l2` and
+requirements in a dedicated environment `benchopt_benchmark_logreg_l2` and
 give you a benchmark plot on l2-regularized logistic regression:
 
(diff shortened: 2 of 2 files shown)


mathurinm:
Are we using the results of `plot_run_benchmark` elsewhere ? It seems not, in that case maybe we don't need to run for the objective with lambda = 0.01 lambda_max ? That would make doc building faster

examples/plot_run_benchmark.py
```diff
@@ -19,7 +19,8 @@
 
 try:
     save_file = run_benchmark(
-        Benchmark(BENCHMARK_PATH), ['sklearn', 'lightning'],
+        Benchmark(BENCHMARK_PATH), ['sklearn[liblinear]', 'sklearn[saga]',
+                                    'sklearn[newton-cg]', 'lightning'],
```
tomMoral, line 23: Why is `lightning` not on your figure?

tomMoral (approved):
I totally agree!

Just a small question

examples/plot_run_benchmark.py
mathurinm: It was not installed when I ran `plot_run_benchmark.py` 

I had to because I can't build the doc with `make html`, see upcoming issue

agramfort:
looks good https://benchopt.github.io

>

mathurinm:
Any idea why it looks less smooth on CI than on my computer ?
There are spikes on CI that bug me:
image

mathurinm:
Also in case you missed it @ŧommoral
> Are we using the results of plot_run_benchmark elsewhere ? It seems not, in that case maybe we don't need to run for the objective with lambda = 0.01 lambda_max ? That would make doc building faster, and we could increase the number of repetitions

agramfort:
CI machine is not just for us...

>

tomMoral:
> Are we using the results of `plot_run_benchmark` elsewhere ? It seems not, in that case maybe we don't need to run for the objective with lambda = 0.01 lambda_max ? That would make doc building faster

We are only using it in the doc and `README` so we could indeed use only one parameter:
```
$ git grep sphx_glr_plot_run_benchmark
README.rst:.. figure:: https://benchopt.github.io/_images/sphx_glr_plot_run_benchmark_001.png
doc/index.rst:.. figure:: auto_examples/images/sphx_glr_plot_run_benchmark_001.png
```

Yeah I agree that the reason is probably because we are not alone in the CI.
What do you think of median curves?

mathurinm:
I like the median curves (they are currently used, right ? I think it's because with 3 runs, the median is still unstable). In #242 I increased n-repetitions

Issue #239, opened 2021-11-24
PLT: for some objectives, suboptimality, relative suboptimality etc do not make sense


mathurinm:
For example, monitoring sparsity of the iterates for the Lasso, I can't envision a case where one wants to use relative suboptimality.
I can't see an obvious solution right now.

mathurinm:
would it make sense to propose suboptimality/relative suboptimality only for `value` ?

agramfort:
You can have test and train losses.  would you want suboptimality for test
loss?

>

mathurinm:
I think not, I've never seen in a paper : for the test loss F(x^*) depends on the solvers being added to the graph.

So, for me it makes more sense to propose subopt only for `value`, the objective actually being minimized.

agramfort:
let's do this then

>

tomMoral:
I disagree, this is only for your particular use case. Adding this restriction won't improve the lib and might hinder some use cas.
I don't see why we should restrict this type of plots, if they don't make sense, don't look at it :)

I would say in the long run, we can add some meta data to make sure we only plot what make sense but this is a longer term goal. So I would keep it that way for now.

mathurinm:
We can discuss solutions around coffee but having 16 matplotlib figures showing up as the result of our default `plot_run_benchmark` really bothers me (indeed this is my particular usecase)

mathurinm:
Another example : this plot in the doc is wrong: 
https://benchopt.github.io/_images/sphx_glr_plot_run_benchmark_python_R_julia_005.png

F(x*) is read as F at optimum, while it is in fact min_F, F being || . ||_0, thus 0 here.

mathurinm:
One more reason to talk about it: `benchopt run ./benchmark_lasso -s sklearn -s celer -d simulated --no-html -r1` opens **288 figures**. 
At this point I would like to not look at them, but is not possible with matplotlib (with plotly I agree)

PR #243, opened 2021-11-24
display name of objective on the ylabel when --no-html

### LINKED ISSUE #238: PLT: when using --no-html, the value being plotted is not displayed
This can be confusing when one has 16 graphs displayed, for example I am confused as to what is being plotted here:
image

mathurinm:
closes #238 
I was ending up with 16 figures without knowing which was which. 

image

also related to #239 : relative suboptimality for test loss may not be relevant

### DIFF
--- a/benchopt/plotting/plot_objective_curve.py
+++ b/benchopt/plotting/plot_objective_curve.py
@@ -105,7 +105,7 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
     else:
         plt.legend(fontsize=14)
         plt.xlabel("Time [sec]", fontsize=14)
-        plt.ylabel(y_label, fontsize=14)
+        plt.ylabel(obj_col[10:] + ': ' + y_label, fontsize=14)
         plt.title(title, fontsize=14)
         plt.tight_layout()
 
--- a/benchopt/plotting/plot_objective_curve.py
+++ b/benchopt/plotting/plot_objective_curve.py
@@ -109,7 +109,7 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
     else:
         plt.legend(fontsize=14)
         plt.xlabel("Time [sec]", fontsize=14)
-        plt.ylabel(_remove_prefix(obj_col, "objective_") + ': ' + y_label,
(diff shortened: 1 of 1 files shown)


benchopt/plotting/plot_objective_curve.py
```diff
@@ -105,7 +105,7 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
     else:
         plt.legend(fontsize=14)
         plt.xlabel("Time [sec]", fontsize=14)
-        plt.ylabel(y_label, fontsize=14)
+        plt.ylabel(obj_col[10:] + ': ' + y_label, fontsize=14)
```
agramfort, line 108: where does the 10 come from?

mathurinm: it's because all objectives start with `objective_` : https://github.com/benchopt/benchOpt/blob/master/benchopt/base.py#L296

mathurinm: in python 3.9 I'd use `removeprefix` to be clearer, but we support 3.7 and 3.8 too

agramfort: I would do something like this:

```
def remove_prefix(text, prefix):
    if text.startswith(prefix):
        return text[len(prefix):]
    return text  # or whatever
so it's explicit
```

benchopt/plotting/plot_objective_curve.py
```diff
@@ -105,7 +109,8 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
     else:
         plt.legend(fontsize=14)
         plt.xlabel("Time [sec]", fontsize=14)
-        plt.ylabel(y_label, fontsize=14)
+        plt.ylabel(_remove_prefix(obj_col, "objective_") + ': ' + y_label,
```
tomMoral, line 112:
```suggestion
        plt.ylabel(f"{_remove_prefix(obj_col, 'objective_')}: {y_label}",
```

tomMoral (approved):
Fixed a nitpick but LGTM!

PR #246, opened 2021-11-25
plot histogram only for objective_value

### LINKED ISSUE #245: What is histogram for objectives other than value ?
Related to #239

Launching a Lasso benchmark on a single dataset, single problem:
```
benchopt run . -s sklearn -s celer -d simulated\[n_samples=100,n_features=10000,rho=0\] --no-html -r1 -p "Lasso Regression[reg=0.1,fit_intercept=False]" 
```
I get 3 histograms, apparently one for value, one for support_size (where convergence is measured compared to the smallest support > this is shaky) and one for duality gap

image

mathurinm:
closes #245 

I do this because I don't think we know what convergence means for a value which is not the minimized objective.

### DIFF
--- a/benchopt/plotting/__init__.py
+++ b/benchopt/plotting/__init__.py
@@ -69,8 +69,8 @@ def plot_benchmark(fname, benchmark, kinds=None, display=True, plotly=False,
                     if kind not in PLOT_KINDS:
                         raise ValueError(
                             f"Requesting invalid plot '{kind}'."
-                            "Should be in:\n"
-                            f"{PLOT_KINDS}")
+                            f"Should be in:\n{PLOT_KINDS}"
+                        )
                     # plot histogram only when convergence has a meaning
                     if kind == "histogram":
                         if obj_col != "objective_value":
--- a/benchopt/plotting/__init__.py
+++ b/benchopt/plotting/__init__.py
@@ -65,24 +65,28 @@ def plot_benchmark(fname, benchmark, kinds=None, display=True, plotly=False,
 
                 plot_id = get_plot_id(benchmark.name, df_obj)
 
-                for k, obj_col in itertools.product(kinds, obj_cols):
-                    if k not in PLOT_KINDS:
+                for kind, obj_col in itertools.product(kinds, obj_cols):
+                    if kind not in PLOT_KINDS:
                         raise ValueError(
                             f"Requesting invalid plot '{k}'. Should be in:\n"
                             f"{PLOT_KINDS}")
-                    plot_func = globals()[PLOT_KINDS[k]]
+                    # plot histogram only when convergence has a meaning
+                    if kind == "histogram":
+                        if obj_col != "objective_value":
+                            continue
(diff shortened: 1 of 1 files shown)


benchopt/plotting/__init__.py
```diff
@@ -65,24 +65,29 @@ def plot_benchmark(fname, benchmark, kinds=None, display=True, plotly=False,
...
 
-                for k, obj_col in itertools.product(kinds, obj_cols):
-                    if k not in PLOT_KINDS:
+                for kind, obj_col in itertools.product(kinds, obj_cols):
+                    if kind not in PLOT_KINDS:
                         raise ValueError(
-                            f"Requesting invalid plot '{k}'. Should be in:\n"
+                            f"Requesting invalid plot '{kind}'."
+                            "Should be in:\n"
                             f"{PLOT_KINDS}")
```
tomMoral, line 73:
```suggestion
                            f"Should be in:\n{PLOT_KINDS}"
                        )
```

benchopt/plotting/__init__.py
```diff
@@ -65,24 +65,29 @@ def plot_benchmark(fname, benchmark, kinds=None, display=True, plotly=False,
...
-                    plot_func = globals()[PLOT_KINDS[k]]
+                    # plot histogram only when convergence has a meaning
+                    if kind == "histogram":
+                        if obj_col != "objective_value":
+                            continue
```
tomMoral, line 77:
```suggestion
                    # For now only plot histogram and suboptimality for
                    # objective_value for which we monitor convergence
                    # XXX - find a better solution
                    if obj_col != "objective_value" and (
                            kind == "histogram" or "subopt" in kind):
                        continue
```

tomMoral (commented):
If we do this one, we might as well do for both `histogram` and `suboptimality`.

I don't really like this solution as `suboptimality/histogram` can make sense in many other cases. But I understand that having tens of windows pop-up randomly can be bothering so if you really need this, let's do this as a temporary fix.

Note that you can already filter the kind of plot by using `benchopt config -b . set plots convergence_curve` in the benchmark folder. This cannot yet be tweaked to select part of the loss.

benchopt/plotting/__init__.py
tomMoral, line 77:
```suggestion
                    # For now only plot histogram and suboptimality for
                    # objective_value for which we monitor convergence
                    # XXX - find a better solution
                    if obj_col != "objective_value" and (
                            kind == "histogram" or "subopt" in kind):
                        continue
```

mathurinm:
To which benchmarks/metric where histogram make sense are you thinking?

tomMoral:
> To which benchmarks/metric where histogram make sense are you thinking?

For any value that converges, you may want to look at the histogram or suboptimality.
I am mainly thinking of bi-level objectives when I say that there can be multiple curves converging, but also if you monitor the loss and and a proxy such as the error or the F1 score for ML training, you can look at such curves too.

However,  having 288 windows for `matplotlib` as you reported in #239 is not acceptable either so I agree that we should merge this temporary fix.

PR #244, opened 2021-11-25
fix channels priorities

mathurinm:
I also set capture_stdout to False, even without CONDA_DEBUG since env creation takes a long time and people want feedback e.g. #90 

Fix #241

PR #254, opened 2021-12-02
Quiet flag in dependencies install

mathurinm:
finishes what was started in #244


tomMoral (approved):
LGTM! thanks @mathurinm  :)

PR #256, opened 2021-12-03
Circle ci with main instead of master

tanglef:
Replace all mentions of master as main in `.circleci/config.yml`


tomMoral:
thx a lot for the fix @tanglef !

PR #258, opened 2021-12-03
Toggle shaded areas

### LINKED ISSUE #228: ENH/PLT: add way to toggle shaded areas on and off
Sometimes I use `-r 10` to get less bumpy curves, but it may get hard to read with the shaded areas and it could be nice to disable them. It's not a critical feature, but if it's easy to do then WDYT ?
image

tanglef:
This introduces a button to toggle on/off the shaded areas (and thus resolves #228 )
| On | Off |
:----------------------:|:------------------------------:
Capture d’écran de 2021-12-03 15-08-23 | Capture d’écran de 2021-12-03 15-08-29

### DIFF
--- a/benchopt/plotting/helpers_compat.py
+++ b/benchopt/plotting/helpers_compat.py
@@ -19,7 +19,7 @@ def fill_between_x(fig, x, q1, q9, y, color, marker, label, plotly=False):
         marker_size=10, name=label, legendgroup=label,
         hoverlabel=dict(namelength=-1),
         hovertemplate='%{text} <br> (%{x:.1e},%{y:.1e}) <extra></extra>',
-        text=[label for _ in x]
+        text=[label for _ in x], showlegend=True,
     ))
     fig.add_trace(go.Scatter(
         x=q1, y=y, mode='lines', showlegend=False,
--- a/benchopt/plotting/html/static/hover_index.css
+++ b/benchopt/plotting/html/static/hover_index.css
@@ -4,162 +4,227 @@
 
 /* Place rectangles in a grid */
 .grid {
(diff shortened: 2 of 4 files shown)


agramfort:
it does not work for me. When I hide the quantiles it removes some of the
legend. It should not

>

tanglef:
Do you have a case where it happens I could reproduce ? (I fixed the histogram, but can't seem to reproduce that issue for objective curves)

agramfort:
in the screenshot you shared do you see that the legend is not the same?

>

tanglef:
It's just me when I made the rectangle to take part of my screen in capture. But if you want I can take with the whole legend:
| off | on |
:----------:|:--------:
Capture d’écran de 2021-12-03 18-25-21 | Capture d’écran de 2021-12-03 18-25-14

tanglef:
To be even more precise as to what I turn on/off: I only toggle on/off the visibility of traces in the plotly graph that have no name attribute (and if a trace is in the legend, then it has a name so its visibility is not modified and thus the legend is not either).

agramfort:
https://user-images.githubusercontent.com/161052/144647177-b18990b3-3b3b-42e4-86c2-b87b4f1d28c6.mov

tanglef:
Okay, I saw [a post](https://community.plotly.com/t/scatter-plot-legend-not-showing-for-a-single-trace/21713) in plotly doc about that (the legend disappears because this is now a single trace graph and thus we need to specify that the legend is visible). I'll take care of it tomorrow or later (can't right now).
Thank you for noticing that!

agramfort (approved):
works for me !

agramfort:
thx @tanglef

PR #261, opened 2021-12-08
Use timeit instead of time

jolars:
Benchopt currently uses the time module, but I believe a better option is to use timeit instead. See https://stackoverflow.com/questions/17579357/time-time-vs-timeit-timeit and https://docs.python.org/2/library/timeit.html#timeit.default_timer for context, but basically what `timeit.timeit()` does is to disable garbage collection when timing, which makes for more consistent benchmarking since garbage collection will be scheduled intermittently throughout benchmarking and affect timings inconsistently between runs.

This is not completely uncontroversial, because I think methods that are leaner, i.e. waste less resources, will trigger garbage collection less often, but this seems like a minor issue to me since most of the time garbage collection will just happen for completely unrelated reasons.

This of course won't affect garbage collection on the R side (and I'm unsure about how cython works in this regard).

### DIFF
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -1,4 +1,4 @@
-import time
+import timeit
 from datetime import datetime
 
 from .utils import product_param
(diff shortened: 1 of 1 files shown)


jolars:
Sorry, I totally bungled this pull request. Disregard this.

PR #264, opened 2021-12-11
ENH list all possible objectives when unmatched pattern option is passed to -p

### LINKED ISSUE #224: ENH: List all possible objectives when unmatched pattern option is passed to -p
this is done for datasets/solvers, but not for objectives

bmalezieux:
All permutations of parameters values are displayed when filters don't match objective attributes. However, I'm not sure of the format of the pattern...

Fixes #224

### DIFF
--- a/benchopt/benchmark.py
+++ b/benchopt/benchmark.py
@@ -8,7 +8,8 @@ from .base import BaseSolver, BaseDataset
 from .utils.colorify import colorify, YELLOW
 from .utils.dynamic_modules import _load_class_from_module
 from .utils.parametrized_name_mixin import product_param
-from .utils.parametrized_name_mixin import _list_all_parametrized_names
+from .utils.parametrized_name_mixin import (_list_all_parametrized_names,
+                                            _list_all_objective_params)
 
 from .utils.conda_env_cmd import install_in_conda_env
 from .utils.conda_env_cmd import shell_install_in_conda_env
@@ -276,6 +277,14 @@ class Benchmark:
         all_solvers = _list_all_parametrized_names(*self.get_solvers())
 
         _validate_patterns(all_solvers, solver_patterns, name_type='solver')
+
+    def validate_objective_filters(self, objective_params):
+        "Check that all objective filters match at least one objective setup"
+
+        # List all choices of objective parameters
+        all_params = _list_all_objective_params(self.get_benchmark_objective())
+
+        _validate_patterns(all_params, objective_params, name_type="objective")
 
 
 def _check_name_lists(*name_lists):
--- a/benchopt/benchmark.py
+++ b/benchopt/benchmark.py
@@ -8,8 +8,7 @@ from .base import BaseSolver, BaseDataset
 from .utils.colorify import colorify, YELLOW
 from .utils.dynamic_modules import _load_class_from_module
 from .utils.parametrized_name_mixin import product_param
-from .utils.parametrized_name_mixin import (_list_all_parametrized_names,
-                                            _list_all_objective_params)
+from .utils.parametrized_name_mixin import _list_all_parametrized_names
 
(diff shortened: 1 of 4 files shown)


benchopt/cli/main.py
```diff
@@ -158,7 +159,7 @@ def run(benchmark, solver_names, forced_solvers, dataset_names,
     solvers_option = ' '.join(['-s ' + s for s in solver_names])
     forced_solvers_option = ' '.join(['-f ' + s for s in forced_solvers])
     datasets_option = ' '.join(['-d ' + d for d in dataset_names])
-    objective_option = ' '.join(['-p ' + p for p in objective_filters])
+    objective_option = ' '.join(['-p ' + f"'{p}'" for p in objective_filters])
```
tomMoral, line 162: Why do you need this?

benchopt/benchmark.py
```diff
@@ -277,6 +278,14 @@ def validate_solver_patterns(self, solver_patterns):
 
         _validate_patterns(all_solvers, solver_patterns, name_type='solver')
 
+    def validate_objective_filters(self, objective_params):
+        "Check that all objective filters match at least one objective setup"
+
+        # List all choices of objective parameters
+        all_params = _list_all_objective_params(self.get_benchmark_objective())
```
tomMoral, line 285: You should simply use `_list_all_parametrized_names`. No need for an extra function here.

tomMoral: I would call this `all_objectives` to be consistent with the previous functions.

tomMoral (changes requested):
You can reuse `_list_all_parametrize_names` for this PR.

benchopt/cli/main.py
bmalezieux: It seems that this doesn't work for names that contain blank spaces. For instance, `benchopt run -e -p "Convolutional Sparse Coding[reg=0.1]" benchmark_csc/` returns the following error: `Error: Got unexpected extra arguments (Sparse Coding[reg=0.1])` without adding this fix.

benchopt/cli/main.py
```diff
@@ -158,7 +159,7 @@ def run(benchmark, solver_names, forced_solvers, dataset_names,
     solvers_option = ' '.join(['-s ' + s for s in solver_names])
     forced_solvers_option = ' '.join(['-f ' + s for s in forced_solvers])
     datasets_option = ' '.join(['-d ' + d for d in dataset_names])
```
tomMoral, line 161:
```suggestion
    forced_solvers_option = ' '.join([f"-f '{s}'" for s in forced_solvers])
    datasets_option = ' '.join([f"-d '{d}'" for d in dataset_names])
```

benchopt/cli/main.py
tomMoral, line 162: Ok, let's do this for all parameters then:
```suggestion
    objective_option = ' '.join([f"-p '{p}'" for p in objective_filters])
```

benchopt/benchmark.py
```diff
@@ -277,6 +277,20 @@ def validate_solver_patterns(self, solver_patterns):
...
+        # List all choices of objective parameters
+        all_objectives = _list_all_parametrized_names(
+            self.get_benchmark_objective()
+        )
+
+        _validate_patterns(
+            all_objectives,
+            objective_params,
+            name_type="objective"
+        )
```
tomMoral, line 292: Doesn't it pass on a single line?

tomMoral (approved):
A few more comments but overall it looks good!

benchopt/benchmark.py
bmalezieux: No, it is the reason why I changed the name `all_objectives` by something shorter in the previous commit

tomMoral (approved):
LGTM! thx @bmalezieux

PR #281, opened 2021-12-22
API deprecate -p shortcut in favor of -o for Objective

### LINKED ISSUE #225: CLI: could -p be -o ?
-o for Objective, as -s for Solver and -d for Dataset

mathurinm:
fixes #225 
I have an issue, the warnings are captured even with BENCHOPT_DEBUG=1, I think they should not because the user won't be notified of the deprecation.

It remains to adapt the test if you agree with this @tomMoral @agramfort

### DIFF
--- a/benchopt/cli/main.py
+++ b/benchopt/cli/main.py
@@ -99,10 +99,9 @@ def run(benchmark, solver_names, forced_solvers, dataset_names,
         plot=True, html=True, pdb=False, do_profile=False,
         env_name='False', old_objective_filters=None):
     if len(old_objective_filters):
-        print("should warn but nothing is printed")
         warnings.warn(
             'Using the -p option is deprecated, use -o instead',
-            DeprecationWarning
+            FutureWarning,
         )
         objective_filters = old_objective_filters
 
--- a/benchopt/cli/main.py
+++ b/benchopt/cli/main.py
@@ -1,4 +1,5 @@
 import click
+import warnings
 from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


benchopt/cli/main.py
```diff
@@ -93,7 +97,14 @@
 def run(benchmark, solver_names, forced_solvers, dataset_names,
         objective_filters, max_runs, n_repetitions, timeout,
         plot=True, html=True, pdb=False, do_profile=False,
-        env_name='False'):
+        env_name='False', old_objective_filters=None):
+    if len(old_objective_filters):
+        print("should warn but nothing is printed")
+        warnings.warn(
```
mathurinm, line 103: @agramfort do you know why the warning is captured ? The print is executed if you use `-p` but the warning is not displayed

agramfort:
maybe a call to catch_warnings somewhere?

mathurinm:
@tomMoral do you know where this is coming from ? If we call `warnings.resetwarnings()` of `warnings.setfilter("always")` before `warnings.warn`, then the warnings are shown (in addition it shows useful DeprecationWarning for `stping_strategy`). 

So something is catching the warnings somewhere ?

agramfort (commented):
this requires a what's new API change entry.

no doc to be updated?

mathurinm:
I added the whatsnew entry, the doc is OK because it's generated from the python click file and there are no examples where we filter objectives

agramfort (approved):

PR #284, opened 2021-12-24
PLT use 1 more significant digit for time axis in plotly

### LINKED ISSUE #234: PLT: some xticks have the same label
I think this is a precision error, the last tick being 120 s and the one before 100, but since there is a single digit, they are both displayed as 1e+2 
image

mathurinm:
Before:
image
After:
image

closes #234


tanglef:
All good for me for the x-axis, thanks.

Quick question though: if it's bothering for the x-axis, it is easily editable (as you did), but as you see in your "After", it happens sometimes in the y-axis too, only for a few plots.
Do you know any way we could adapt the format to the values (with the difficulty that when changing the axis-scale the values also change and we don't want 3 significant digits all the time)

tanglef (approved):

TomDLT (approved):

PR #289, opened 2022-01-20
ENH add X_density argument to make_correlated_data to generate sparse design matrix

mathurinm:


benchopt/datasets/tests/test_dataset_simulated.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 import numpy as np
+from scipy import sparse
```
agramfort, line 3: Use nested import to avoid paying the price of import

benchopt/datasets/simulated.py
```diff
@@ -1,12 +1,13 @@
 """Helper to generate simulated data in benchopt."""
 import numpy as np
 from numpy.linalg import norm
+from scipy import sparse
```
agramfort, line 4: Use nested import

benchopt/datasets/simulated.py
```diff
@@ -1,12 +1,13 @@
...
 
 from ..utils.checkers import check_random_state
 
 
 def make_correlated_data(
-        n_samples=100, n_features=50, n_tasks=1, rho=0.6, snr=3,
+        n_samples=100, n_features=50, n_tasks=1, rho=0.6, snr=3, X_density=1,
```
agramfort, line 10: Add it at the end to avoid breaking public API

agramfort (commented):
Also don’t forget to update what’s new. 

From the 🚠

benchopt/datasets/tests/test_dataset_simulated.py
mathurinm: can you explain more ?

benchopt/datasets/simulated.py
```diff
@@ -84,6 +89,11 @@ def make_correlated_data(
     else:
         X = rng.randn(n_samples, n_features)
 
+    if X_density != 1:
```
agramfort, line 92: import scipy sparse here or just after the docstring. We do not want to have to import scipy sparse when importing benchopt. Typically  it slows down a lot the autocomplete

mathurinm:
done @agramfort

agramfort (approved):
@tomMoral merge if happy

benchopt/datasets/tests/test_dataset_simulated.py
tomMoral: :+1:

doc/whats_new.rst
```diff
@@ -37,10 +37,13 @@ API
 - 'stop_strategy' attribute is replaced by 'stopping_strategy' to harmonize
   with 'stopping_criterion', by `Benoît Malézieux`_ (:gh:`274`).
 
-- Add `import_from` method in `safe_import_context` to allow importing common
+- Add ``import_from`` method in ``safe_import_context`` to allow importing common
```
tomMoral, line 40: thx!

benchopt/datasets/simulated.py
```diff
@@ -7,7 +7,7 @@
 
 def make_correlated_data(
         n_samples=100, n_features=50, n_tasks=1, rho=0.6, snr=3,
-        w_true=None, density=0.2, random_state=None):
+        w_true=None, density=0.2, random_state=None, X_density=1):
```
tomMoral, line 10: I disagree with @agramfort, I think it is okay to put it here as it is more consistent and good practice is to always use keyword for such helper.

```suggestion
        w_true=None, density=0.2, X_density=1, random_state=None):
```

tomMoral (approved):
LGTM, just a question about the position of the parameter.

PR #293, opened 2022-02-25
Raise FutureWarning instead of DeprecationWarning if stop_strategy is used

tanglef:
Raise Future and not Deprecation warning in the `base.py` file (cf @mathurinm  #281)

### DIFF
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -57,7 +57,7 @@ class BaseSolver(ParametrizedNameMixin, DependenciesMixin, ABC):
             warnings.warn(
                 "'stop_strategy' attribute is deprecated, "
                 "use 'stopping_strategy' instead",
-                DeprecationWarning
+                FutureWarning
             )
             return self.stop_strategy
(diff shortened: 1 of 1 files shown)


agramfort:
Not sure to follow here. If you deprecate i would say it’s a deprecation warning you should raise

tanglef:
Also, #281 corrected the Deprecated to Future type of warnings in cli `main.py`, so this is simply an harmonization change really.

agramfort:
thx @tanglef

mathurinm:
@agramfort you won't see them if you have not asked for or they are not in your main. 
Put this in a `test_warnings.py`:

```python
import warnings
warnings.warn("truc", DeprecationWarning)
A = 1
```

and in a ipython shell run `from test_warnings import A`. The warning will not show up.

agramfort:
>
> and in a ipython shell run from test_warnings import A. The warning will
> not show up.
>

yes ok but when I run the file I do this the warning. Using I get
deprecation warnings on execution
unless we deprecate a module

PR #294, opened 2022-02-25
Load results script if not found

### LINKED ISSUE #291: HTML is not rendered properly when folder does not contain static/result.js
WIth @qb3 and other we often share benchmark results by sending the html file alone, and then all the figures are displayed on the page instead of a single graph if `static/result.js` is not present in the same folder as the html.

tanglef:
Fixes #291 by loading the JS file from `results` website if directory not found.

**NB:** We can't directly load the JS file from the `plotting` directory of the benchopt repo as Github prevented this use in 2013

### DIFF
--- a/benchopt/plotting/html/templates/result.mako.html
+++ b/benchopt/plotting/html/templates/result.mako.html
@@ -135,7 +135,11 @@
 
     <a class="backtotop" title="Back to top" href="#top"><i class="fas fa-level-up-alt"></i></a>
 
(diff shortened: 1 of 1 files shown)


agramfort:
I don't like that. I would suggest to embed directly the JS in the file so it works in the future and without an internet connection. If you update https://benchopt.github.io/results/static/result.js in the future you can break many existing files.

agramfort:
to be clear I mean just insert / copy the .js in the html.

PR #297, opened 2022-03-10
Unpin 3.8 version in env creation 

mathurinm:


tomMoral:
The test in the main repo should already check that julia and R are installable and working.
But I would be more confident if `benchopt test benchmark_lasso` passes with this change anyway.

mathurinm:
spams does not support python 3.9: 
```
UnsatisfiableError: The following specifications were found
to be incompatible with the existing python installation in your environment:

Specifications:

  - python-spams -> python[version='2.7.*|3.5.*|3.6.*|>=2.7,<2.8.0a0|>=3.6,<3.7.0a0|>=3.8,<3.9.0a0|>=3.7,<3.8.0a0|>=3.5,<3.6.0a0']

Your python: python=3.9
```
@gdurif is there a chance to have python-spams working with python 3.9 ?

gdurif:
@mathurinm We have a new official repository [here](https://github.com/getspams/spams-python) associated with the [`spams`](https://pypi.org/project/spams/) package on PyPI.

`python-spams` is obsolete (and will be removed from PyPI soon, c.f. https://github.com/getspams/spams-python/issues/12).

I will do the PR on https://github.com/benchopt/benchmark_lasso/ to fix this.

mathurinm:
Moving to python 3.9 breaks spams, but there is work on the spams side to fix it.

On the other side, bumping version  fixes installation of cython solvers such as celer and cd_solver that are very interesting for the benchmark, so I am strongly in favor of temporarily removing spams, and readding it after a new release as we did for modopt.

@agramfort would you be ok with that ?

tomMoral:
I am against removing the solver and readding. just skip it in the test for now.

As for python version, why pin the version and not jsut use the latest one? I think we pinned python to version `3.8` because julia and R were not working with `3.9` (see #89 )
Now if we don't need to pin the version, it would be the best, so we don't have to chose when to move on to the next version.

benchopt/utils/conda_env_cmd.py
```diff
@@ -22,7 +22,7 @@
   - conda-forge
   - nodefaults
 dependencies:
-  - python=3.8
+  - python=3.9
```
tomMoral, line 25:
```suggestion
  - python
```

tomMoral:
Note that the python version installed on the CI is not necessarily the same as the one from the created environment.

benchopt/utils/conda_env_cmd.py
tomMoral: This would install the latest available python so `3.9` for now and as `3.10` roll out, it will move on naturally.

mathurinm:
Testing in https://github.com/benchopt/benchmark_lasso/pull/114

mathurinm:
In benchmark_lasso, `python-spams` and `snapml` fail (requires <= 3.9) https://snapml.readthedocs.io/en/latest/installation.html#installation-guide

Maybe pinning to 3.9 and updating regularly ensures install doesn't break everytime a new python version is released  @tomMoral ?

tomMoral:
I am not sure what would reduce the most the maintenance burden.

If these packages are failing, we can just skip them and ping their maintainers, as this is not `benchopt` fault.
Else we need to regularly check, and it is very hard to know when updating will not break any packages in any benchmark.
So I am more in favor of explicit break in CI I guess as `python` is not updated so often.

tomMoral (approved):
LGTM!

PR #300, opened 2022-03-14
FIX pytest configuration for v7.1

tomMoral:
The current failure is due to a change of behavior with `pytest7.1`, from pytest-dev/pytest#9493.
This makes it not possible to have twice the same `conftest.py`

I found out that we can change this by relying on a plugin, that can be loaded at runtime with option `-p` in pytest.
this PR implement this as this make it cleaner (we are directly loading the installed configuration as a fixture, not relying on path resolution from pytest that we don't control very well).

Overall, this is cleaner, so hopefully the CI is green tonight.


tomMoral:
Merging to get a green CI :) :green_heart:

PR #302, opened 2022-03-18
fix quantiles

### LINKED ISSUE #301: [BUG plotly] Quantile curves selection isn't linked to toggled-off traces
As of now, we can toggle off the trace in the legend, toggle off the quantile area. But those two are not linked together so when turning them on again we can have something like that (no main curve which is ok, but with the quantiles, which are not wanted here):

Capture d’écran de 2022-03-18 10-13-32

tanglef:
Fixes #301 by transferring "main" trace `visible` property to quantile curves by legendgroup.


tanglef:
Circle CI seem to be broken, the final update about github authentication protocols has been made...

```
Merging 302
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

TomDLT (approved):
Nice!
The code makes sense, I tested the PR and it fixes #301 and both issues describes in #317.

mathurinm:
Thanks a lot @tanglef and @TomDLT

PR #319, opened 2022-03-31
DOC rm fit_intercept in logreg L2 example

mathurinm:
https://github.com/benchopt/benchmark_logreg_l2/pull/27 removed this parameter from the benchmark, which broke the doc.

This is a fix


tomMoral (approved):
LGTM, thx @mathurinm

Issue #321, opened 2022-04-07
Possibility to skip a dataset based on some objective parameters


zaccharieramzi:
I think I didn't see a way to do this, so I was wondering whether it was possible to skip a dataset altogether depending on some objective parameters.

My usecase is rather simple to explain: a parameter in my objective controls the type of CV DL model to be used. In the case of VGG, it's not possible to have images smaller than 32x32, which is the case for MNIST, so I would like to skip this dataset in this case.

I can try to look into how to implement that if you want.

tomMoral:
It seems a nice feature to add, and not too complicated.
I would say we need to do the same as what we do with the solver: add a method `skip` in the objective and skip when it return `True`.

zaccharieramzi:
But is there a simple way to access the objective's params, rather than the objective dict?
Or would we need to pass the discriminating params in the objective dict?

zaccharieramzi:
Actually in the current framework, it would make more sense to have the objective skip the dataset, rather than the dataset skip the objective.

I am going to rather implement that.

tomMoral:
Yes that would make more sense as the order for the runner is:
- instanciate the dataset
- instanciate the objective and use `_set_dataset`.
- instanciate the solver and use `_set_objective`.

the call to skip shoud be done in `_set_dataset`, as it is done in `Solver.set_objective`.

zaccharieramzi:
Yes I did exactly that! ready to review

tomMoral:
Closed in #323

PR #323, opened 2022-04-07
allowed objective to skip some datasets

zaccharieramzi:
This solves #321 

However, I don't really know how I should test this. I understand there is a dummy benchmark, but should I add a case where I skip a dataset i the objective? Like create a fake dataset and always skip it?


benchopt/tests/test_benchmarks/dummy_benchmark/objective.py
```diff
@@ -16,6 +20,11 @@ def set_data(self, X, y):
         self.X, self.y = X, y
         self.lmbd = self.reg * self._get_lambda_max()
 
+    def skip(self):
```
tomMoral, line 23:
```suggestion
    def skip(self, X, y):
```

tomMoral (approved):
I would change a bit the API to match the one from `Solver.skip`.
Otherwise, LGTM!, thx @zaccharieramzi

zaccharieramzi:
Yup makes total sense, just added in your suggestions @tomMoral

benchopt/tests/test_benchmarks/dummy_benchmark/objective.py
```diff
@@ -16,6 +20,11 @@ def set_data(self, X, y):
         self.X, self.y = X, y
         self.lmbd = self.reg * self._get_lambda_max()
 
+    def skip(self, X, y):
+        if np.all(self.X == 0):
```
tomMoral, line 24: whoops, i forgot one!

```suggestion
        if np.all(X == 0):
```

PR #325, opened 2022-04-12
WIP support passing options through config files

mathurinm:
proof of concept to start discussion.

Example of config file: `ex.yml`
```yaml
objectives:
  - Lasso Regression[fit_intercept=True,reg=0.5]
datasets:
  - simulated
  - leukemia
solvers:
  - celer
  - cd
repetitions: 1
```

These config files can be committed on repos for examples, shared across collaborators, are easier to maintain/modify/keep track of when the command to launch the benchmark is long

Questions: 
- do we warn when a CLI option is replaced by an option in the config ?
- is there a way to avoid hardcoding the replacement of CLI option by specified config options ? 
- discuss names in yml config vs parameters names in benchopt code 
- support forced, pdb, html, do_profile, env_name in the config ?

### DIFF
--- a/benchopt/cli/main.py
+++ b/benchopt/cli/main.py
@@ -1,3 +1,4 @@
+import yaml
 import click
 import warnings
 from pathlib import Path
@@ -65,6 +66,8 @@ main = click.Group(
 @click.option('--timeout',
               metavar="<int>", default=100, show_default=True, type=int,
               help='Timeout a solver when run for more than <timeout> seconds')
+@click.option('--file', 'config_file', default=None,
+              help="YAML configuration file")
 @click.option('--plot/--no-plot', default=True,
               help="Whether or not to plot the results. Default is True.")
 @click.option('--html/--no-html', default=True,
@@ -95,7 +98,7 @@ main = click.Group(
               "named <env_name>. To install the required solvers and "
               "datasets, see the command `benchopt install`.")
 def run(benchmark, solver_names, forced_solvers, dataset_names,
-        objective_filters, max_runs, n_repetitions, timeout,
+        objective_filters, max_runs, n_repetitions, timeout, config_file,
         plot=True, html=True, pdb=False, do_profile=False,
         env_name='False', old_objective_filters=None):
     if len(old_objective_filters):
@@ -104,6 +107,18 @@ def run(benchmark, solver_names, forced_solvers, dataset_names,
             FutureWarning,
         )
         objective_filters = old_objective_filters
+
+    if config_file is not None:
+        with open(config_file, "r") as f:
+            config = yaml.safe_load(f)
+        # TODO automated wya to do this ?
+        # Todo warn if if config overrides CLI argument ?
+        solver_names = tuple(config.get("solvers", False)) or solver_names
+        dataset_names = tuple(config.get("datasets", False)) or dataset_names
+        repetitions = config.get("repetitions", False) or repetitions
+        objective_filters = tuple(config.get("objectives", False)) or objective_filters
+    # import ipdb
+    # ipdb.set_trace()
 
     from benchopt.runner import run_benchmark
 


jolars:
Great idea! But isn't this syntax more kosher:

```yaml
objectives:
  lasso:
    fit_intercept: true
    reg: 0.5
datasets:
  - simulated
  - leukemia
solvers:
  - celer
  - cd
repetitions: 1
```

or even:

```yaml
objectives:
  - model: lasso
    fit_intercept: true
    reg: 0.5
  - model: slope
    fit_intercept: true
datasets:
  - simulated
  - leukemia
solvers:
  - celer
  - cd
repetitions: 1
```

    do we warn when a CLI option is replaced by an option in the config ?

Shouldnt CLI options take precedence? Or are you talking about defaults?

mathurinm:
I pushed a solution for the names differing in config file (where we want to use the name of the click options, e.g. `--solver`) and what is passed to the click decorated function (a variable `solver_names`).

To avoid hardcoding the dict, we could rely on click automapping from option to variable name (strip '--', replace '-' by '_') and do the reverse mapping automatically in `_get_run_args`.

WDYT @tomMoral @TomDLT

benchopt/cli/main.py
```diff
@@ -28,14 +59,14 @@
...
-@click.option('--objective-filter', '-o', 'objective_filters',
+@click.option('--objective-filter', '-o', OPT_TO_VAR['--objective-filter'],
               metavar='<objective_filter>', multiple=True, type=str,
               help="Filter the objective based on its parametrized name. This "
               "can be used to only include one set of parameters.")
 @click.option('--old_objective-filter', '-p', 'old_objective_filters',
               multiple=True, type=str,
               help="Deprecated alias for --objective_filters/-o.")
-@click.option('--solver', '-s', 'solver_names',
+@click.option('--solver', '-s', OPT_TO_VAR['--solver'],
```
mathurinm, line 69: an alternative is to remove all the variable names in options, let click do the conversion automatically and convert back when needed in `_get_run_args`. this way no need for `OPT_TO_VAR`

We keep to chose the variable names when we declare all the variables that unpack the result of `_get_run_args`

mathurinm:
- Following @jolars suggestion I chose that CLI args override those specified in config_file. click knows when a parameter has its default value, so that's OK.

- I think in the config file parameters should be passed with their CLI option name (without the --) to ease documentation so it requires a bit of work to match the parameters from config_file with the ones from CLI, because click as already changed their names. Hence I have removed specific variable names in `click.option` (not an API breaking change), let click do the automatic conversion, and then assign variable names as the output of `_get_run_args`.

TomDLT:
I agree with both points.

More generally, is there a good reason to have different names for CLI arguments and the function parameters ?

Support for pdb, env_name, etc. would be nice but is not necessary.
Same for autocompletion.

mathurinm:
I'd like to add an example of config.yml file in the doc, on which page does it belong in your opinion? On the CLI page it doesn't look easy to place it properly (right after the run command?)

jolars:
> I'd like to add an example of config.yml file in the doc, on which page does it belong in your opinion? On the CLI page it doesn't look easy to place it properly (right after the run command?)

Right after the run command sounds fine to me.

Klopfe:
I agree with Johan, I think after the run command is a good idea

Klopfe (approved):

tomMoral (commented):
This looks very neat! thanks @mathurinm 

Could you add at least one test in `test_cli.py`? I think mimicking the `TestRunCmd::test_benchopt_run` passing only config should be enough.

benchopt/tests/test_benchmarks/dummy_benchmark/datasets/simulated.py
```diff
@@ -13,8 +13,8 @@ class Dataset(BaseDataset):
     # the cross product for each key in the dictionary.
     parameters = {
         'n_samples, n_features': [
-            (100, 5000),
-            (100, 10000)
+            (100, 500),
+            (100, 1000)
```
tomMoral, line 17: Good call

benchopt/tests/test_cli.py
```diff
@@ -170,6 +171,46 @@ def test_benchopt_run_profile(self):
...
+        max-runs: 1
+        force-solver:
+          - python-pgd[step_size=1.5]
+          - Test-Solver
+        """
+        tmp = tempfile.NamedTemporaryFile(mode="w+")
+        tmp.write(config)
+        tmp.flush()
+
+        run_cmd = [str(DUMMY_BENCHMARK_PATH), '--file', tmp.name, '--no-plot']
```
tomMoral, line 197:
```suggestion
        run_cmd = [str(DUMMY_BENCHMARK_PATH), '--config', tmp.name, '--no-plot']
```

benchopt/tests/test_cli.py
```diff
@@ -170,6 +171,46 @@ def test_benchopt_run_profile(self):
...
+          - {SELECT_ONE_OBJECTIVE}
+        dataset:
+          - {SELECT_ONE_SIMULATED}
+        n-repetitions: 2
+        max-runs: 1
+        force-solver:
+          - python-pgd[step_size=1.5]
+          - Test-Solver
+        """
+        tmp = tempfile.NamedTemporaryFile(mode="w+")
```
tomMoral, line 193: We could use the same one as before no?

tomMoral:
```suggestion
```

tomMoral (approved):
LGTM!

there is just a small leftover `--file` in the test.

tomMoral:
Thanks @mathurinm !!! :)

PR #326, opened 2022-04-13
ENH remove the sidebar in the benchmark result homepage

### LINKED ISSUE #318: Remove the sidebar in the benchmark result homepage
The side menu in the benchmark homepage feels out of place. The design is different from the rest (strong background, round design, big blue drown-down menus) vs (white background, square design, small gray drop-down menus). It holds a major position (sidebar) as if it was a major menu selection, whereas it is just a filtering option.

image

I would rather remove the sidebar, and add the three corresponding filters with drop-down menus next to (and similar to) the "Show X entries" menu.

image

TomDLT:
This PR removes the sidebar, and add dropdown selectors matching "Select X entries".
It also change the sysinfo names to uppercase ("cpu" --> "CPU").

Fixes #318

___
**Before**
___
image
___
**After**
___
image

### DIFF
--- a/benchopt/plotting/html/static/benchmark.js
+++ b/benchopt/plotting/html/static/benchmark.js
@@ -60,7 +60,7 @@ function change(ll_item) {
   for (item = 0; item < ll_item.length; item++) {
     filter.push(
       document
-        .getElementById("select" + ll_item[item].replace(/\s/g, ""))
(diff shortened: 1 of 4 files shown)


agramfort (changes requested):
KISS 

I like it 👍

agramfort (approved):

tomMoral (approved):
LGTM, thx@TomDLT

PR #287, opened 2022-01-07
PLT add plot with iterations

tomMoral:
Add plot relative to iteration numbers.

Also quick start a documentation page on plotting tools. This should be populated at some point.... :(

PR #327, opened 2022-04-13
ENH change plotted quantiles to (q3, q7)

TomDLT:
This PR proposes to use (q3, q7) instead of (q1, q9) in the plots.

For some fast benchmarks, computational time can vary a lot between runs, and the log-plots are highly visually affected by some unfortunate outliers. Using quantiles closer to the median would alleviate this issue by providing a more robust estimate of the computational times.

In a Gaussian distribution with zero-mean and unit standard deviation, the range [-0.5, 0.5] corresponds to the percentiles [30.8, 69.2], so the range (q3, q7) does seem a reasonable choice to me.

### DIFF
--- a/benchopt/plotting/plot_objective_curve.py
+++ b/benchopt/plotting/plot_objective_curve.py
@@ -80,11 +80,11 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
         df_ = df[df['solver_name'] == solver_name]
         curve = df_.groupby('stop_val').median()
 
-        q1 = df_.groupby('stop_val')['time'].quantile(.1)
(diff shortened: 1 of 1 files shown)


agramfort:
I think that using quartile 0.25 and 0.75 would be more standard.

agramfort (approved):

tomMoral:
I agree that this would make it more robust.
I have a concurrent PR changing for `0.2` and `0.8` (see #287). I updated it to use quartile into account.
As the PR also add some new plot (iteration wise + interpolation to avoid weird loooking curves), I will close this one to avoid merge conflicts. Feel free to review theother one so we can merge it fast :)

PR #328, opened 2022-04-18
Disable garbage collection during timing

jolars:
It seems to me like it would make sense to disable garbage collection before timing the run of a solver. 

Let's say you want to benchmark ten different solvers. Then it's likely that during the run of some of these solvers, garbage collection will be initiated. So, for instance, it's possible that garbage collection will occur during the runs for solvers 4 and 10, inflating the timings for these runs.

The usual argument against disabling garbage collection is that methods that consume lots of resources should be punished for that during timing, but the problem with that argument is that there's no guarantee that the heavy resource usage caused by say, solver 9, will actually end up causing garbage collection during that exact run and instead happen during the run of solver 10.

The only downside as I see it here is that we're not handling garbage collection for other languages called through python.

### DIFF
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -1,4 +1,5 @@
 import time
+import gc
 
 from datetime import datetime
 
@@ -67,9 +68,11 @@ def run_one_resolution(objective, solver, meta, stop_val):
     if DEBUG:
         print(f"DEBUG - Calling solver {solver} with stop val: {stop_val}")
 
+    gc.disable()
     t_start = time.perf_counter()
     solver.run(stop_val)
     delta_t = time.perf_counter() - t_start
+    gc.enable()
     beta_hat_i = solver.get_result()
     objective_dict = objective(beta_hat_i)
 


agramfort:
@jolars do you see any real impact of this change on benchmark results?

TomDLT:
What do you think of calling `gc.collect()`, both before starting and before stopping the timer?

jolars:
> @jolars do you see any real impact of this change on benchmark results?

I did some benchmarks on the lasso benchmark with simulated data now and no, there doesn't actually appear to be any real difference now that I look at it.

image

This is just a single example, but it looks roughly like this for every benchmark I've run so far.

> What do you think of calling `gc.collect()`, both before starting and before stopping the timer?

Great idea, but it seems like running a full garbage collection before each timing is quite time consuming.

tomMoral:
Thanks @jolars for the proposition and the benchmark.

I am mainly not in favor of changing the interpreter behavior for the benchmark as this could lead to weird bug.
For instance, for algorithm like ISTA/FISTA, the iterates need to be garbage collected, else the memory consumption would grow linearly with iterations.

But I think having some quantification of this impact is helpful so it is good that you looked at it :pray:
Could you  add the script you run for your benchmark as `scripts/benchmark_gc_impact.py`? This would allow to check this in the future if we reconsider this question.

PR #332, opened 2022-04-21
FIX rename histogram as bar chart

jolars:
What's currently referred to as histograms in the result plots are actually bar charts (or bar plots). This pull request replaces "histogram" with "bar chart" everywhere.


tomMoral (approved):
LGTM

PR #339, opened 2022-04-28
ENH add confidence intervals to plotly results

jolars:
This pull request adds support for confidence intervals to the result plots, but just for the plotly versions of the plots. The idea is to have intervals that show the confidence for the median in addition to quantile plots, so that the width of the intervals decrease as the number of repetitions increase. I've changed the slider to a drop-down list instead. The default is still quantiles.

Here's for instance a benchmark using quantiles and confidence intervals, which was run for 50 repetitions:

image

image

I've used standard percentile bootstrap CIs here because they are always positive and need less babysitting that bias-controlled intervals do. The bootstrapping does add a bit of extra overhead, but I've found it to be largely neglible. But if you would prefer an interval that's faster to compute then there are alternatives.

This functionality could be added to the CLI interface too of course, although it would require a new argument.

My proficiency in javascript is horrible by the way, so I hope I didn't screw anything up.


agramfort (commented):
it seems it can make the plots more readable 👍

PR #345, opened 2022-05-04
CLN rename get_one_beta -> get_one_solution

### LINKED ISSUE #351: API stop returning beta dimension in datasets
For early benchmarks such as Lasso this made sense. Now for resnets or lasso path, this also depends on the objective (e.g. n_lambdas in a path)

tomMoral:
Change name to be clearer, cc @pierreablin.

fixes #351

### DIFF
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -389,7 +389,7 @@ class BaseObjective(ParametrizedNameMixin, DependenciesMixin):
         """
         return False, None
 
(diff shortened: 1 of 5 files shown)


zaccharieramzi:
Shouldn't we maintain backwards's compatibility?

tomMoral:
As this feature was introduced a week ago, I am in favor of skipping backward compat here.

PR #362, opened 2022-05-07
ENH add support for custom parameters in CLI

### LINKED ISSUE #193: ENH: be able to set arbitrary values of solver/dataset/objective parameters via CLI
I think it would be nice to be able to do `benchopt run . -d simulated[n_samples=30,n_features=50]` even if `simulated.py` does not have the values, 30 and 50, as long as `n_samples` and `n_features` are keys of `Dataset.parameters` 

all unspecified parameters would take all their default values

TomDLT:
Fixes #193

This PR makes the following commands work as expected:
- `benchopt run . -d simulated[n_samples=30,n_features=50]`
- `benchopt run . -d simulated[n_features=50, n_samples=30,]`
- `benchopt run . -d simulated[n_samples=[30, 300], n_features=[10, 12]]`
- `benchopt run . -s sklearn[liblinear, lbfgs]`

I added a bunch of unit tests, but I am not super familiar with the test suite, so this might require additional integration tests.


tomMoral (commented):
Nice one @TomDLT ! This looks quite good and it seems to work as expected.

For the test, I would move the current ones in `test_runner.py` and add another one in `test_cli.py::TestRun`, copying a bit what is done in `test_benchopt_run` with different parameters.

benchopt/tests/__init__.py
```diff
@@ -12,9 +12,9 @@
 REQUIREMENT_BENCHMARK_PATH = TEST_BENCHMARK_DIR / 'requirement_benchmark'
 
 # Pattern to select specific datasets or solvers.
-SELECT_ONE_SIMULATED = r'simulated*200*rho=0]'
-SELECT_ONE_PGD = r'python-pgd*step_size=1]'
-SELECT_ONE_OBJECTIVE = r'dummy*reg=0.1]'
+SELECT_ONE_SIMULATED = r'simulated[n_features=200,rho=0]'
+SELECT_ONE_PGD = r'python-pgd[step_size=1]'
+SELECT_ONE_OBJECTIVE = r'dummy[reg=0.1]'
```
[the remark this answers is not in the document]
TomDLT: No it does not work anymore.
We need a convention to isolate the parameters from the rest (I made the assumption that parameters are within "[]"). Also, I am not sure it makes sense to have a regex on the parameters.

The previous filters would need to be changed into:
- `'python-pgd*step_size=1]'`  --> `'python-pgd*[step_size=1]'`
- `'dummy*reg=0.1]'`  --> `'dummy*[reg=0.1]'` 
- `'simulated*200*rho=0]'` --> `'simulated*[n_features=200,rho=0]'`

TomDLT:
Thanks for the suggestions!

One issue we need to solver is that there is not always a parameter checking when creating the instance. For example, if I use the filter `"python-pgd$[wrong_name=2]"`, it will create a solver instance `Python-PGD[step_size=1,wrong_name=2]`.

To fix this, we could check if parameters are in `klass.parameters`, but this list might be incomplete. What do you think?

tomMoral:
> To fix this, we could check if the parameters are in `klass.parameters`, but this list might be incomplete. What do you think?

I think we should design the API in such a way that we can check this. There are 2 options to make this:
- we require that all the parameters are defined in `cls.parameters`.
- we define all parameters in `__init__` and remove the `cls.parameters`.

The current way is to not implement `__init__` but only `parameters`. To me it makes sense as this allow to define ranges of parameters for grid search while it would be weird to make defaults in `__init__` define such list of value.
So I would say it is fine to require that all parameters are in `cls.parameters`. But I am open to discuss this.

Using the new API, I also found some weird cases that we need to investigate:
- with boolean, `arg=True` gave me `arg = 'true'`.
- with string, `arg=18` gave me an integer ^^. this is the expected behavior but as we used to not check type but just name, this tripped my script. We should properly document this.
- we should still support the `*` and deprecate them in the args. So `sol*[arg=1]` shoud match `solver1, solver2` with common `args=1` and `solver1[arg1=1,*]` should raise a warning and have the same behavior as `solver[arg1=1]`.
 
I did not investigate all this, I will continue playing with it to see if I can find other weird cases.

tomMoral:
Another potential change:
- we had some parameters that were coupled, with the synthax `parameters = {'arg1, arg2`: [(False, 0), (True, 0.9), (True, 1.3)]`, which avoided some combination. I don't think this would work anymore?

README.rst
```diff
@@ -26,8 +26,11 @@ be **as simple as doing**:
     conda activate benchopt
     pip install benchopt
     git clone https://github.com/benchopt/benchmark_logreg_l2
-    benchopt install -e benchmark_logreg_l2 -s cd -s sklearn
-    benchopt run -e ./benchmark_logreg_l2 -s cd -s sklearn
+    benchopt install -e ./benchmark_logreg_l2 \
+        -s lightning -s sklearn
+    benchopt run -e ./benchmark_logreg_l2 \
+        -s lightning -s "sklearn[liblinear, newton-cg]" \
+        -d "simulated[n_features=500,n_samples=200]"
```
tomMoral, line 33: Let's make a config file for this, now that we can :)

benchopt/tests/__init__.py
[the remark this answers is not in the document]
tomMoral: I agree with you that it does not make sense to have `*` in the args anymore and that we should raise error on incomplete `[/]` pairs. I also agree for no-regexp in the parameters.

We should add `$` in the `python-pgd` pattern by default, and make it possible to use `python-pgd*[step_size=1]` to select both `python-pgd` and `python-pgd-with_cb`, that way, this will be less surprising to users as using full name should not trigger other ones.

benchopt/tests/test_cli.py
```diff
@@ -201,8 +223,8 @@ def test_benchopt_run_config_file(self):
             run(run_cmd, 'benchopt', standalone_mode=False)
 
         out.check_output(r'Test-Solver:', repetition=11)
-        out.check_output(
-            r'Python-PGD\[step_size=1.5\]:', repetition=11)
+        out.check_output(r'Python-PGD\[step_size=2\]:', repetition=11)
+        out.check_output(r'Python-PGD\[step_size=3\]:', repetition=11)
```
tomMoral, line 227: Nice test!

tomMoral (commented):
I like the test. now we should just check that we manage correctly the boolean and the solver names and we should be good to go.

tomMoral:
Another use case that is not covered when playing with this:
- while passing `s[arg=[a, b]]`, I got a string `arg = "[a, b]"` as I think `[a, b]` is not correctly interpreted as `["a", "b"]`. Not really sure how to change this. I guess we need to recursively evaluate the expression, or require for adding `"` to get proper string representation.

TomDLT:
Good points. Here is the current status:
- [x] Support `Solver['arg1, arg'=[(False, 0), (True, 0.9)]`.
- [x] Support `Solver[arg=[a, b]]`.
- [ ] Support and deprecate `*` in the parameters.
- [x] Require all parameters to be defined in cls.parameters.
- [ ] Update solvers/objectives/datasets in all benchmarks to define all parameters.
- [ ] Add `config_example.yml`. https://github.com/benchopt/benchmark_logreg_l2/pull/32
- [ ] Document the behavior.
- [ ] Future work: support `benchopt install --config ...`

benchopt/tests/__init__.py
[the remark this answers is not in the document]
TomDLT: Well, it would be weird to have at the same time :
- `python-pgd` select `python-pgd` and not `python-pgd-with-cb`
- `dummy` select `Dummy Sparse Regression`

For now, selecting the solver/dataset/objective supports regex, so the user need to explicitly add "$" (`python-pgd$`) to get only `python-pgd`.

benchopt/benchmark.py
```diff
@@ -449,7 +449,7 @@ def get_all_runs(self, solver_names=None, forced_solvers=None,
 
 
 def _check_name_lists(*name_lists):
-    "Normalize name_list ot a list of lowercase str."
+    "Normalize name_list to a list of lowercase str."
     res = []
     for name_list in name_lists:
         if name_list is None:
```
tomMoral, line 455: We need to remove the `.lower()` bellow as this causes issue with the python tokens.

tomMoral (commented):
This fixes the sue cases I tried today.
A few more comments but I feel this is close to being good.

benchopt/tests/__init__.py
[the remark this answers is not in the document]
TomDLT: Makes sense. See bd979bc3bc53af65a81fa86d660a5e159ec59408.

benchopt/benchmark.py
```diff
@@ -502,47 +501,84 @@ def _extract_options(name):
...
+    if len(matches) == 0:
+        return basename, [], {}
+    elif len(matches) > 1:
+        raise ValueError(f"Invalid name (multiple brackets): {name}")
+    else:
+        match = matches[0]
+
         if match[0] != "[" or match[-1] != "]":
-            raise ValueError("Invalid name: {}".format(name))
+            raise ValueError(f"Invalid name (missing bracket): {name}")
```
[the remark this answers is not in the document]
TomDLT: Makes sense, probably a leftover of buggy code.

benchopt/benchmark.py
TomDLT: Well, I feel like it is useful to tolerate both `python-pgd` and `python-PGD`. Another fix would be to support `true` for `True`. See 78ac0cbc863e9d0322f7ac884e74ad5e9fac6332

tomMoral: This is already supported as the regex use `IGNORECASE`

benchopt/benchmark.py
```diff
@@ -502,47 +501,84 @@ def _extract_options(name):
...
     import ast
+    original = string
+
+    # First, replace all quoted names with their hashes, to avoid modification.
+    all_matches = re.findall(r"'[^']*'", string)
+    for match in all_matches:
+        string = string.replace(match, str(hash(match)))
+
+    # Second, add quotes to all variable names (foo -> 'foo').
+    string = re.sub(r"[a-zA-Z][a-zA-Z0-9_-]*", r"'\g<0>'", string)
```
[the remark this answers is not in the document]
TomDLT: No. From "A" to "z" also includes unwanted characters. The [ASCII](https://www.ascii-code.com/) list is:

```
A  B  ...  Y  Z  [  \  ]  ^  _  `  a  b  ...  y  z
```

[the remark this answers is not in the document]
tomMoral: whoops :) good point.

benchopt/benchmark.py
```diff
@@ -466,16 +466,116 @@ def is_matched(name, include_patterns=None, default=True):
...
+    # Sequences with "=" are made into a dict expression {'foo': 'bar'}.
+    # Sequences without "=" are made into a list expression ['foo', 'bar'].
+    if "=" in string:
+        string = "{" + string.replace("=", ":") + "}"
+    else:
+        string = "[" + string + "]"
+
+    # Remove quotes for python language tokens
+    for word in ["True", "False", "None"]:
+        string = string.replace(f"'{word.lower()}'", word)
```
tomMoral, line 569: Not sure we need this one?

benchopt/benchmark.py
```diff
@@ -466,16 +466,116 @@ def is_matched(name, include_patterns=None, default=True):
...
+    >>> _extract_parameters("foo=[bar, baz]")  # {"foo": ["bar", "baz"]}
+    >>> _extract_parameters("foo=1, baz=True")  # {"foo": 1, "baz": True}
+    >>> _extract_parameters("'foo, bar'=[(0, 1),(1, 0)]")
+    >>> # {"foo, bar": [(0, 1),(1, 0)]}
+    """
+    import ast
+    original = string
+
+    # First, replace all quoted names with their hashes, to avoid modification.
+    all_matches = re.findall(r"'[^']*'", string)
```
tomMoral, line 548: shouldn't we also match `"strings"`?
```suggestion
    all_matches = re.findall(r"'[^']*'", string)
    all_matches += re.findall(r'"[^"]*"', string)
```

benchopt/benchmark.py
```diff
@@ -466,16 +466,116 @@ def is_matched(name, include_patterns=None, default=True):
...
+    # Sequences without "=" are made into a list expression ['foo', 'bar'].
+    if "=" in string:
+        string = "{" + string.replace("=", ":") + "}"
+    else:
+        string = "[" + string + "]"
+
+    # Remove quotes for python language tokens
+    for word in ["True", "False", "None"]:
+        string = string.replace(f"'{word.lower()}'", word)
+        string = string.replace(f"'{word}'", word)
```
tomMoral, line 570:
```suggestion
    for token in ["True", "False", "None"]:
        string = string.replace(f"'{token.lower()}'", token)
        string = string.replace(f"'{token}'", token)
```

benchopt/tests/test_runner.py
```diff
@@ -53,3 +56,164 @@ def test_get_one_solution():
...
+    results = filt_(["Test-Dataset[41,42]"])
+    assert len(results) == 2
+    _assert_parameters_equal(results[0][0], dict(n_samples=41))
+    _assert_parameters_equal(results[1][0], dict(n_samples=42))
+
+    # test recursive eval within list, to eval strings
+    results = filt_(["Test-Dataset[n_samples=[foo,True]]"])
+    assert len(results) == 2
+    _assert_parameters_equal(results[0][0], dict(n_samples="foo"))
+    _assert_parameters_equal(results[1][0], dict(n_samples=True))
```
tomMoral, line 156: Not positional ;)

benchopt/tests/test_runner.py
```diff
@@ -53,3 +56,164 @@ def test_get_one_solution():
...
+    # Test the selection of dataset with a positional parameter.
+
+    def filt_(filters):
+        return list(_filter_classes(TEST_DATASET_POSITIONAL, filters=filters))
+
+    results = filt_(["Test-Dataset[42]"])
+    assert len(results) == 1
+    _assert_parameters_equal(results[0][0], dict(n_samples=42))
+
+    results = filt_(["Test-Dataset[41,42]"])
```
tomMoral, line 147: Could you test that this also works with tokens and strings?

benchopt/tests/test_runner.py
```diff
@@ -53,3 +56,164 @@ def test_get_one_solution():
...
+    _assert_parameters_equal(results[0][0], dict(n_samples=41, n_features=42))
+
+    # get grid over one parameter (n_samples)
+    results = filt_(["Test-Dataset[n_samples=[41,42], n_features=19]"])
+    assert len(results) == 2
+    _assert_parameters_equal(results[0][0], dict(n_samples=41, n_features=19))
+    _assert_parameters_equal(results[1][0], dict(n_samples=42, n_features=19))
+
+    # get grid over two parameter (n_samples, n_features)
+    results = filt_(["Test-Dataset[n_samples=[41,42], n_features=[19, 20]]"])
```
tomMoral, line 109: Could you add 2 tests: one with a list of string and one with a list of token?
Just to make sure they don't get mixed up with variables.

doc/cli.rst
```diff
@@ -19,6 +19,22 @@ For example using BASH shell you need to run:
 The `benchopt` command also comes with tab completion for the solver name
 and the dataset name.
 
+.. note::
+    **Optional parameter syntax:** For some CLI parameter (solver, objective,
```
tomMoral, line 23:
```suggestion
    **Optional parameters syntax:** For some CLI parameters (solver, objective,
```

doc/cli.rst
```diff
@@ -19,6 +19,22 @@ For example using BASH shell you need to run:
...
+    dataset), additional parameters can be given with the following syntax:
```
tomMoral, line 24:
```suggestion
    dataset), additional values can be given with the following syntax:
```

tomMoral (approved):
Ok, I think this is my last round of review as I am starting to only nitpick ^^'

LGTM! This is a great feature, I have been using it for 3 days and I don't have any pain points!
Thanks a lot @TomDLT for the hard work.

TomDLT:
Thanks for all the reviews, it really helped to cover all the corner cases.

tomMoral (approved):
All good for me!! thx a lot @TomDLT

tomMoral:
Oh, we forgot to put an entry in whats_new.
can you do it in a follow up Pr @TomDLT ?

jolars:
Thanks, this is really great! It has bugged me constantly :)

PR #363, opened 2022-05-08
ENH add autocompletion for --config

mathurinm:


tomMoral (approved):
Works like a charm! Thx @mathurinm

PR #368, opened 2022-05-10
ENH support solver and dataset install from config file

mathurinm:


tomMoral (approved):
LGTM!

Thx @mathurinm .

PR #384, opened 2022-05-12
ENH use current working directory as default in CLI

### LINKED ISSUE #382: ENH support run command without `.` by default
Reading the doc, it is not always clear that one should run the command as follows:

```
benchopt run . --config pathological.yml
```

One would rather like a command like 

```
benchopt run --config pathological.yml
```
to work out of the box.

tanglef:
Fixes #382
Not exactly sure how to test this though....

### DIFF
--- a/benchopt/cli/completion.py
+++ b/benchopt/cli/completion.py
@@ -46,6 +46,7 @@ def complete_benchmarks(ctx, param, incomplete):
 
 def find_benchmark_in_args(args):
     "Find the benchmark in preceeding args for benchmark dependent completion."
+    args.extend([Path.cwd()])  # default path is current working directory
     for b in args:
         if (Path(b) / "objective.py").exists():
             return Benchmark(b)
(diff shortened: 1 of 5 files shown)


tanglef:
`urllib` error on leukemia. Will trigger back the CI tmrw...

tomMoral (commented):
this seems ok to do that.
Questions:
- How does it fare with autocompletion?
- Could you test this by checking that not passing a repo tells you that you don't have a `objective.py` function in the current repo?

mathurinm:
This is a great feature !

:heavy_check_mark:  The error is raised if we're not in a valid benchmark folder:
```
(base) ➜  ~ benchopt run 
Usage: benchopt run [OPTIONS] [BENCHMARK]
Try 'benchopt run -h' for help.

Error: Invalid value: The folder '/home/mathurin' does not contain `objective.py`.
Make sure you provide the path to a valid benchmark.
```

:checkmark: autocompletion works nice:

image



LGTM !

tomMoral:
I was speaking of the auto completion when not putting `.`. I think it is not working yet as `benchopt run -s <tab>` gives nothing on my side. In the autocompletion command, if no benchmark is found, we should also default to `.`

tanglef:
This finally works

Capture d’écran de 2022-05-13 11-16-55

tomMoral (approved):
LGTM! thansk @tanglef

PR #385, opened 2022-05-13
MNT return dict in get_data in test benchmarks datasets

mathurinm:


tomMoral (approved):
LGTM

tomMoral (commented):
You need to implement `get_one_solution` in the objective when only returning a dict.

PR #386, opened 2022-05-13
Get cuda info without nvcc

### LINKED ISSUE #163: Graphic card information in results
It would be good to have the name of the graphic card in addition to the CUDA version currently saved.
With NVIDIA-smi there is the command:

``` bash
nvidia-smi --query-gpu=name --format=csv,noheader
```

so using `subprocess` module (already used for other informations), we can get it easily. Not sure for others though.

### LINKED ISSUE #378: ENH use pynvml to check conda info
I got several issues on platforms where I have a GPU but no access to `nvcc`, as solvers such as `cuml` relies on the fact that we can find a `nvcc`.
I propose to use `pynvml` to check the GPU info: this will be more robust and this will also allow to access to the name of the GPU, which is more informative (knowing if we have a `V100` or a `Titan`).

tanglef:
Fixes #378 and Fixes #163 by collection GPU name and cuda version
This doesn't use `pynvml` (which is "a wrapper around NVML library" that does, more quickly, what the smi thinner wrapper `nvidia-smi -q -x` does)

### DIFF
--- a/benchopt/utils/sys_info.py
+++ b/benchopt/utils/sys_info.py
@@ -37,8 +37,8 @@ def _get_cuda_version():
         return f"{name}: cuda_{version}"
     except AttributeError:
         import warnings
-        warngings.warn("Found `nvidia-smi` but could not parse cuda version "
-                                  "or device name.")
+        warnings.warn("Found `nvidia-smi` but could not parse cuda version "
+                      "or device name.")
         return None
 
 
--- a/benchopt/utils/sys_info.py
+++ b/benchopt/utils/sys_info.py
@@ -25,7 +25,7 @@ def _get_processor_name():
     return out
 
 
-def _get_cuda_version():
+def get_cuda_version():
     "Return GPU name and CUDA version."
     if which("nvidia-smi") is None:
         return None
@@ -96,7 +96,7 @@ def get_sys_info():
     )
 
     # Info on dependency libs
-    info["version-cuda"] = _get_cuda_version()
+    info["version-cuda"] = get_cuda_version()
     info["version-numpy"] = (np.__version__, _get_numpy_libs())
     info["version-scipy"] = scipy.__version__
 
--- a/benchopt/utils/sys_info.py
+++ b/benchopt/utils/sys_info.py
@@ -26,13 +26,14 @@ def _get_processor_name():
 
 
 def _get_cuda_version():
-    "Return CUDA version."
-    if which("nvcc") is None:
+    "Return GPU name and CUDA version."
+    if which("nvidia-smi") is None:
         return None
-    command = ["nvcc", "--version"]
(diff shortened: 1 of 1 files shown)


benchopt/utils/sys_info.py
```diff
@@ -26,13 +26,14 @@ def _get_processor_name():
...
+    if which("nvidia-smi") is None:
         return None
-    command = ["nvcc", "--version"]
+    command = ["nvidia-smi", "-q", "-x"]
     out = subprocess.check_output(command).strip().decode("utf-8")
-    out = out.splitlines()[-1]  # take only last line
-    return out
+    version = re.search('<cuda_version>(.*)</cuda_version>', out).group(1)
+    name = re.search('<product_name>(.*)</product_name>', out).group(1)
+    return f"{name}: cuda_{version}"
```
tomMoral, line 36:
```suggestion
    try:
        version = re.search('<cuda_version>(.*)</cuda_version>', out).group(1)
        name = re.search('<product_name>(.*)</product_name>', out).group(1)
        return f"{name}: cuda_{version}"
    except AttributeError:
        import warnings
        warngings.warn("Found `nvidia-smi` but could not parse cuda version "
                                  "or device name.")
        return None
```

tomMoral (approved):
Nice, it is indeed pretty neat. Just a small nitpick to better protect in case the output is not parse correctly, as we don't want to lose benchmark results due to bad parsing in the end.

benchopt/utils/sys_info.py
tanglef: yes, didn't think about this case. Thanks

benchopt/utils/sys_info.py
```diff
@@ -26,13 +26,20 @@ def _get_processor_name():
 
 
 def _get_cuda_version():
```
agramfort, line 28: I would now make this function public as we use it in benchmark code

tomMoral: Good call

tanglef:
Doc build fails because of `lasso`'s snapml call to `_get_cuda_version()`

benchopt/utils/sys_info.py
```diff
@@ -25,14 +25,21 @@ def _get_processor_name():
...
-    out = out.splitlines()[-1]  # take only last line
-    return out
+    try:
+        version = re.search('<cuda_version>(.*)</cuda_version>', out).group(1)
+        name = re.search('<product_name>(.*)</product_name>', out).group(1)
+        return f"{name}: cuda_{version}"
+    except AttributeError:
+        import warnings
+        warnings.warn("Found `nvidia-smi` but could not parse cuda version "
+                      "or device name.")
```
tomMoral, line 41:
```suggestion
        warnings.warn(
            "Could not parse cuda version or device name from `nvidia-smi`."
         )
```

PR #387, opened 2022-05-13
FIX use utf8 when writing html report

Badr-MOUFAD:


tomMoral (approved):
LGTM, thx @Badr-MOUFAD

Issue #365, opened 2022-05-10
Website scaling for small screens


TomDLT:
The title bar is a bit long, so for small-to-medium screens it takes up to 3 lines. Note that the layout is correctly changed for very small screens (e.g. phones).
![image](https://user-images.githubusercontent.com/11065596/167529791-715d5ae8-6ac4-41b7-8388-55c67e116e15.png)
____
Also, the side panel cannot be scrolled, so if the screen is too small we cannot see part of the side panel.
![image](https://user-images.githubusercontent.com/11065596/167530142-11bc0715-ed9e-41bc-a35f-8d79481fad12.png)

agramfort:
on safari when I scroll down on the front page I get this now
<img width="763" alt="Screenshot 2022-05-17 at 09 05 20" src="https://user-images.githubusercontent.com/161052/168750014-8c154154-0116-49e7-bf8b-e8900679a081.png">

it works on chrome though

PR #389, opened 2022-05-14
Website improvements

TomDLT:
A bunch of improvements for the website (see each commit).
In particular, it makes the sidebar more compact and with larger indents:
before:imageafter:image

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -75,9 +75,10 @@ For instance, the benchmark for Lasso can be retrieved with:
     git clone https://github.com/benchopt/benchmark_lasso
 
 
-Command line usage
-------------------
+Command line interface
+----------------------
 
+The preferred way to run the benchmarks is through the command line interface.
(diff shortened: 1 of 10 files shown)


agramfort (approved):

mathurinm:
It's much better this way, thanks @TomDLT!

tomMoral:
It is much better but we still have some issues:
image

The body starts before the end of the top bar. Not sure how to fix this.

TomDLT:
This is not a problem of the sidebar, but it is a problem of the navbar at the top. For medium size screens, the content is too long and it creates two or three lines. #365

There is probably a more robust/clean solution, but as a workaround I have removed the search box for medium screens. I also removed the link to "write a benchmark", which is long and already highlighted in the frontpage. It is still present in the "Site" dropdown menu.

tomMoral:
Great, this is very nice now :)

thanks a lot @TomDLT !!!

PR #392, opened 2022-05-14
FIX platform version for windows

mathurinm:


benchopt/utils/terminal_output.py
```diff
@@ -49,7 +49,7 @@ def print_normalize(msg, endline=True, verbose=True):
         MIN_LINE_LENGTH, shutil.get_terminal_size((100, 24)).columns
     )
     n_colors = msg.count('\033') // 2
-    msg = msg.ljust(line_length + n_colors * 11)
+    msg = msg.ljust(line_length + n_colors * 11 - 1)
```
mathurinm, line 52: otherwise every msg is on its own line @tomMoral

benchopt/plotting/generate_html.py
```diff
@@ -234,7 +234,7 @@ def get_val(df, key):
             if key == 'platform':
                 return (
                     df["platform"].unique()[0] +
-                    df["platform-release"].unique()[0] + "-" +
+                    str(df["platform-release"].unique()[0]) + "-" +
                     df["platform-architecture"].unique()[0]
```
tomMoral, line 238: I would put string everywhere to be more robust?

tomMoral (approved):
LGTM, thx for looking into this @mathurinm

benchopt/utils/terminal_output.py
mathurinm: I have not understood why this fixed the issue, and it still does not work for @Badr-MOUFAD so I will revert this one

benchopt/plotting/generate_html.py
```diff
@@ -233,9 +233,9 @@ def get_val(df, key):
         if key in df:
             if key == 'platform':
                 return (
-                    df["platform"].unique()[0] +
-                    df["platform-release"].unique()[0] + "-" +
-                    df["platform-architecture"].unique()[0]
+                    str(df["platform"].unique()[0]) +
+                    str(df["platform-release"].unique()[0]) + "-" +
+                    (df["platform-architecture"].unique()[0])
```
tomMoral, line 238:
```suggestion
                    str(df["platform-architecture"].unique()[0])
```

Badr-MOUFAD:
Regarding the problem with the length of the terminal messages as well as the colors,

It's a platform-related issue, namely Windows, as the ANSI API is not enabled by default in it hence must be done manually. I did that by following this answer https://superuser.com/questions/413073/windows-console-with-ansi-colors-handling#:~:text=Fortunately%2C%20the%20global,explicitly%20disable%20ANSI

**P.S**: This comment is for future related issues.

tomMoral:
Could you try to call this at module init in `terminal_output.py`?

```python
import ctypes

kernel32 = ctypes.windll.kernel32
kernel32.SetConsoleMode(kernel32.GetStdHandle(-11), 7)
```

Badr-MOUFAD:
@tomMoral, 
Just tried that on my local machine, after I disabled ANSI, and it solve the problem.

If it doesn't break the code for other platforms (e.g. Linux), I shall make a pull request.

tomMoral:
(this would break on Linux as `ctypes.windll.kernel32` is only defined on windwos.

PR #397, opened 2022-05-16
[ENH] Change output file name

### LINKED ISSUE #373: ENH possibility to name output file ?
suggestion:
`benchopt run . --output bench_large_scale` would create `outputs/bench_large_scale.csv`

need to check if it erases an existing file or not, but it may make life easier

I propose this because in skglm and in my lasso benchmark we ended up with a lot of 
image

tanglef:
Fixes #373

Still need to do the tests

### DIFF
--- a/benchopt/cli/main.py
+++ b/benchopt/cli/main.py
@@ -55,6 +55,7 @@ def _get_run_args(cli_kwargs, config_file_kwargs):
         "pdb",
         "profile",
         "env_name",
+        "output_name",
         "objective_filter",
         "old_objective_filter",
     ]
@@ -149,6 +150,10 @@ def _get_run_args(cli_kwargs, config_file_kwargs):
               help="Run the benchmark in the conda environment "
               "named <env_name>. To install the required solvers and "
               "datasets, see the command `benchopt install`.")
+@click.option("--output-name", "output_name", default=None,
+              type=click.Path(exists=False),
+              help="Name of the output file. If another result file has "
+              "the same name, appends a number to distinguish them.")
 def run(config_file=None, **kwargs):
     if config_file is not None:
         with open(config_file, "r") as f:
@@ -160,8 +165,8 @@ def run(config_file=None, **kwargs):
     (
         benchmark, solver_names, forced_solvers, dataset_names,
         objective_filters, max_runs, n_repetitions, timeout, n_jobs,
-        plot, html, pdb, do_profile, env_name, deprecated_objective_filters,
-        old_objective_filters
+        plot, html, pdb, do_profile, env_name, output_name,
+        deprecated_objective_filters, old_objective_filters
     ) = _get_run_args(kwargs, config)
 
     if len(old_objective_filters):
@@ -202,7 +207,8 @@ def run(config_file=None, **kwargs):
             objective_filters=objective_filters,
             max_runs=max_runs, n_repetitions=n_repetitions,
             timeout=timeout, n_jobs=n_jobs,
-            plot_result=plot, html=html, pdb=pdb
+            plot_result=plot, html=html, pdb=pdb,
+            output_name=output_name
         )
 
         print_stats()  # print profiling stats (does nothing if not profiling)
@@ -257,6 +263,7 @@ def run(config_file=None, **kwargs):
         rf"{'--plot' if plot else '--no-plot'} "
         rf"{'--html' if html else '--no-html'} "
         rf"{'--pdb' if pdb else ''} "
+        rf"--output-name {output_name}"
         .replace('\\', '\\\\')
     )
     raise SystemExit(_run_shell_in_conda_env(
--- a/benchopt/utils/files.py
+++ b/benchopt/utils/files.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+import warnings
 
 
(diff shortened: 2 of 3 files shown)


benchopt/cli/main.py
```diff
@@ -149,6 +150,10 @@ def _get_run_args(cli_kwargs, config_file_kwargs):
               help="Run the benchmark in the conda environment "
               "named <env_name>. To install the required solvers and "
               "datasets, see the command `benchopt install`.")
+@click.option("--output-name", "output_name", default=None,
```
mathurinm, line 153: I would just call this --output (like we have --config)

benchopt/utils/files.py
```diff
@@ -10,3 +11,21 @@ def rm_folder(folder):
         else:
             f.unlink()
     folder.rmdir()
+
+
+def uniquify_results(file_path):
+    "Add a number to filename if it already exists"
+    parent = file_path.parent
```
mathurinm, line 18: really minor: from a conceptual point of view, don't these belong more L25 ? They are not needed if file_path.exists, I'd define them as close as possible to where they're used

mathurinm (commented):
Nice ! Maybe it's be nice to have an option to overwrite existing results if they exist ? this adds yet another parameter, so I have no strong opinion

benchopt/utils/files.py
tanglef: Yes, changed in the latest commit thanks

tanglef:
> Nice ! Maybe it's be nice to have an option to overwrite existing results if they exist ? this adds yet another parameter, so I have no strong opinion

This is doable, but I think I'd rather make `benchopt clean` take a `file_name` argument as optional input instead of adding the parameter to `run`. Either is fine with me

benchopt/cli/main.py
```diff
@@ -257,6 +263,7 @@ def run(config_file=None, **kwargs):
         rf"{'--plot' if plot else '--no-plot'} "
         rf"{'--html' if html else '--no-html'} "
         rf"{'--pdb' if pdb else ''} "
+        rf"--output {output}"
```
tomMoral, line 266: Does this work when it is not specified? I feel like this will output `None`.

benchopt/cli/main.py
```diff
@@ -149,6 +150,10 @@ def _get_run_args(cli_kwargs, config_file_kwargs):
               help="Run the benchmark in the conda environment "
               "named <env_name>. To install the required solvers and "
               "datasets, see the command `benchopt install`.")
+@click.option("--output", "output", default=None,
+              type=click.Path(exists=False),
+              help="Name of the output file. If another result file has "
+              "the same name, appends a number to distinguish them.")
```
tomMoral, line 156: I find it unclear to know if this is a Path or a filename. Could you clarify?

benchopt/cli/helpers.py
```diff
@@ -33,15 +33,27 @@
...
+    else:
+        for ext in [".csv", ".html"]:
+            if ext == ".html":
+                to_remove = output_folder / f"{benchmark.name}_{filename}"
+            else:
+                to_remove = output_folder / filename
+            file = to_remove.with_suffix(ext)
+            if file.exists():
+                print(f"rm {file}")
+                file.unlink()
```
tomMoral, line 56: Also need to remove it from the benchmark html result page and the cache_list?

benchopt/cli/main.py
```diff
@@ -149,6 +150,10 @@ def _get_run_args(cli_kwargs, config_file_kwargs):
               help="Run the benchmark in the conda environment "
               "named <env_name>. To install the required solvers and "
               "datasets, see the command `benchopt install`.")
+@click.option("--output", "output", default=None,
+              type=click.Path(exists=False),
```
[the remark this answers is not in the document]
tomMoral: As this is only used as a string, I would use it as such
```suggestion
@click.option("--output", 
              type=str, default=None,
```

benchopt/tests/test_cli.py
```diff
@@ -285,6 +285,20 @@ def test_benchopt_caching(self, n_rep):
         out.check_output(r'Python-PGD\[step_size=1\]:',
                          repetition=5*n_rep+1)
 
+    @pytest.mark.parametrize('output_name, file_name', [
+        ('tmp_res', "tmp_res"), ("tmp_res", "tmp_res_1")])
```
tomMoral, line 289: The test should all be indepedent, so you cannot rely on the fact that the first run will create the file.
Also, the file should be cleaned at the end of the test to avoid border effect, so please use `CaptureRunOutput` that will clean up the file afterward and call `run` twice in a single context to make sure we test the `uniquify` function.

tomMoral (commented):
nice one! a few comments to better test it.

benchopt/cli/helpers.py
tanglef: I removed the entry associated to the remove files in the cache.
Using `benchopt plot` this removes the entry in the index html page too.

tanglef:
I made the test independents and the changes about the `output` argument and its doc :slightly_smiling_face:

tomMoral (approved):
Looks great! just a small addition for the doc.

Issue #310, opened 2022-03-22
Bug with benchopt run


EnLAI111:
When I run the commande with anaconda prompt in Windows ```benchopt run```, it can't work. After using ipdb for a slover, I found the problem is from here.
```
d:\anaconda\anaconda3\lib\site-packages\benchopt\runner.py(77)run_one_resolution()
info = get_sys_info()
```
Then this line.
```
d:\anaconda\anaconda3\lib\site-packages\benchopt\utils\sys_info.py(93)get_sys_info()
info["version-numpy"] = (np.__version__, _get_numpy_libs())
```
Then this one.
```
d:\anaconda\anaconda3\lib\site-packages\benchopt\utils\sys_info.py(44)_get_numpy_libs()
with SuppressStd() as capture:
```
And at last this function,
```
d:\anaconda\anaconda3\lib\tempfile.py(473)func_wrapper()
    472             def func_wrapper(*args, **kwargs):
--> 473                 return func(*args, **kwargs)
    474             # Avoid closing the file as long as the wrapper is alive,

ipdb> s
--Return--
6
d:\anaconda\anaconda3\lib\tempfile.py(473)func_wrapper()
    472             def func_wrapper(*args, **kwargs):
--> 473                 return func(*args, **kwargs)
    474             # Avoid closing the file as long as the wrapper is alive,

ipdb> func
<built-in method fileno of _io.BufferedRandom object at 0x000002017CFD7930>
ipdb> s
Traceback (most recent call last):
  File "d:\anaconda\anaconda3\lib\site-packages\benchopt\utils\pdb_helpers.py", line 25, in exception_handler
    yield
  File "d:\anaconda\anaconda3\lib\site-packages\benchopt\runner.py", line 303, in run_one_solver
    curve_one_rep, status = run_one_to_cvg_cached(
  File "d:\anaconda\anaconda3\lib\site-packages\joblib\memory.py", line 591, in __call__
    return self._cached_call(args, kwargs)[0]
  File "d:\anaconda\anaconda3\lib\site-packages\joblib\memory.py", line 534, in _cached_call
    out, metadata = self.call(*args, **kwargs)
  File "d:\anaconda\anaconda3\lib\site-packages\joblib\memory.py", line 761, in call
    output = self.func(*args, **kwargs)
  File "d:\anaconda\anaconda3\lib\site-packages\benchopt\runner.py", line 125, in run_one_to_cvg
[...]

PR #398, opened 2022-05-16
BUG - fix ANSI colors in Windows

Badr-MOUFAD:
Following PR #392


Badr-MOUFAD:
Actually, #310 was addressed thanks to PR #390. After that, I was able to run ``benchopt`` benchmarks without problems.

PR #401, opened 2022-05-17
FIX parameters of objective shouldn't be modified by set_data

bmalezieux:


benchopt/base.py
```diff
@@ -383,7 +383,18 @@ def set_dataset(self, dataset):
...
+        # Check if parameters are modified by set_data
+        parameters = {}
+        for key in self._parameters:
+            parameters[key] = getattr(self, key)
         self.set_data(**data)
+        for key in self._parameters:
+            if parameters[key] != getattr(self, key):
+                raise ValueError(
+                    "Parameters of Objective should not be "
+                    "modified by 'set_data'"
```
agramfort, line 395: can you say which parameter is the culprit and print the 2 different values? it helps debugging....

agramfort (approved):

tomMoral (approved):
LGTM! thx @bmalezieux

PR #406, opened 2022-05-22
[WIP] ENH - automatic generation of cookiecutter code

Badr-MOUFAD:
When writing benchmarks, I found it pretty repetitive to layout the structure of benchmarks (i.e. datasets, solvers) over and over again. Indeed, all the defined solvers and datasets follow the same logic.

Based on that, I suggest, through this PR, adding a command into the CLI to automatically generate cookie-cutter code for solvers, datasets, and why not the entire benchmark directory.

I made a small demo, you can generate the template for adding a new solver by running:
```shell
benchopt cookiecutter . --template solver --name some_given_solver 
```
The same thing can be achieved for a new dataset.

If this idea sounds good to you, here are the next steps:
- [ ] implement cookiecutter for the entire benchmark directory
- [ ] handle exceptions and add docs (namely the ``help``)
- [ ] Add/explain this new feature into documentation

### DIFF
--- a/bench_cookiecutter/datasets/dataset.py
+++ b/bench_cookiecutter/datasets/dataset.py
@@ -1,0 +1,16 @@
+from benchopt import BaseDataset
+from benchopt import safe_import_context
+
+with safe_import_context() as import_ctx:
+    # your dependencies here
+    pass
+
+
+class Dataset(BaseDataset):
+    name = "{name}"
+
+    def __init__(self):
+        pass
+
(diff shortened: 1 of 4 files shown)


mathurinm:
There is a template for benchmarks here : https://github.com/benchopt/template_benchmark 

Probably this means that it should be referenced better in the doc, maybe here : https://benchopt.github.io/#available-benchmarks

PR #421, opened 2022-06-07
ENH add tag in results csv file

mathurinm:
My concern is that it will break the loading of existing csv files if one does not use `comment='#'`

### DIFF
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -178,7 +178,7 @@ def get_results(fnames, kinds, root_html, benchmark_name, copy=False):
     for fname in fnames:
         print(f"Processing {fname}")
 
-        df, _ = read_results(fname)
+        df = pd.read_csv(fname)
         datasets = list(df['data_name'].unique())
         sysinfo = get_sysinfo(df)
         # Copy CSV if necessary and give a relative path for HTML page access
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -178,7 +178,7 @@ def get_results(fnames, kinds, root_html, benchmark_name, copy=False):
     for fname in fnames:
         print(f"Processing {fname}")
 
-        df = pd.read_csv(fname)
+        df = pd.read_csv(fname, comment='#')
         datasets = list(df['data_name'].unique())
         sysinfo = get_sysinfo(df)
         # Copy CSV if necessary and give a relative path for HTML page access
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -178,7 +178,7 @@ def get_results(fnames, kinds, root_html, benchmark_name, copy=False):
     for fname in fnames:
         print(f"Processing {fname}")
 
-        df = pd.read_csv(fname, comment='#')
+        df, _ = read_results(fname)
         datasets = list(df['data_name'].unique())
         sysinfo = get_sysinfo(df)
         # Copy CSV if necessary and give a relative path for HTML page access
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -4,14 +4,12 @@ from datetime import datetime
 
 from joblib import Parallel, delayed
 
-from .config import DEBUG
 from .callback import _Callback
-from .utils.shell_cmd import _run_shell
 from .benchmark import _check_name_lists
 from .utils.sys_info import get_sys_info
-from .utils.files import uniquify_results
 from .utils.pdb_helpers import exception_handler
 from .utils.terminal_output import TerminalOutput
+from .utils.files import uniquify_results, write_results
 
 # For compat with the lasso benchmark, expose INFINITY in this module.
 # Should be removed once benchopt/benchmark_lasso#55 is merged
@@ -325,12 +323,6 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
     for curve in results:
         run_statistics.extend(curve)
 
-    err, tag = _run_shell("git describe --tags --abbrev=0", return_output=True)
-    if err != 0:
-        if DEBUG:
-            print(err, tag)
-        tag = None
-
     import pandas as pd
     df = pd.DataFrame(run_statistics)
     if df.empty:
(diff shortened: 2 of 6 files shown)


benchopt/runner.py
```diff
@@ -338,7 +345,10 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
     else:
         save_file = output_dir / f"{output_name}.csv"
         save_file = uniquify_results(save_file)
-    df.to_csv(save_file)
+    with open(save_file, 'a') as f:
+        f.write(f'#tag: {tag}\n')
```
tomMoral, line 349: I think a more explicit name would be nice:
```suggestion
        f.write(f'# benchmark-git-tag: {tag}\n')
```

tomMoral (commented):
Seems nice. I agree with your concern but I think we don't have a large enough user base so it is a really an issue.

IMO, it is sufficient to add a note in the `whats_new.rst` in case someone stamble uppon this case, and update the plotting scripts in our benchmark.


Maybe to cope with future changes, we should have a `benchopt.read_results`, `benchopt.write_results`?

benchopt/runner.py
```diff
@@ -338,7 +345,10 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
     else:
         save_file = output_dir / f"{output_name}.csv"
         save_file = uniquify_results(save_file)
-    df.to_csv(save_file)
+    with open(save_file, 'a') as f:
```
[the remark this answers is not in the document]
mathurinm: I used `'a'` because I first write the tag, then the content of `df`.

In principle the file should not exist (using `output` we add numbers to save_file) unless two benchmarks finish at the exact same time. If that is the current with the current design we overwrite the first one

I can add a check that the file does not exist, but it's unclear to me what to do if it exists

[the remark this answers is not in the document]
tomMoral: `a` only change something at opening, not if you have multiple writes. (see https://docs.python.org/3/library/functions.html#open)

Basically, `w` opens the file with cursor at the beginning truncating the file content if it exists while `a` put the cursor at the end of the content.
But as long as you keep it open, the cursor is at the end of the previous write.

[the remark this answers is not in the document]
mathurinm: of course yes!

doc/whats_new.rst
```diff
@@ -21,11 +21,19 @@ CLI
...
   files for sharing with others or as supplementary materials for papers.
   By `Thomas Moreau`_ (:gh:`408`).
 
+
+API
+~~~
+
+- Add latest git tag via `benchmark-git-tag` comment in CSV results files,
+  together with ``read_results`` and ``write_results`` functions to handle the CSVs.
+  By `Mathurin Massias`_ (:gh:`421`).
```
tomMoral, line 35:
```suggestion
  Note that this change might break the compatiblity of newest result file
  with pre-existing plotting script, as the results need to be loaded with
  ``pd.read_csv(..., comment="#")``. For future compatibility, we advise using
  helper ``benchopt.utils.files.read_results`` which returns the ``pd.DataFrame``
  By `Mathurin Massias`_ (:gh:`421`).
```

benchopt/utils/files.py
```diff
@@ -25,7 +31,30 @@ def uniquify_results(file_path):
...
         return file_path
+
+
+def read_results(file_path):
+    """Read benchmark results as pandas DataFrame."""
+    with open(file_path, "r") as f:
+        metadata = {}
+        for line in takewhile(lambda s: s.startswith("#"), f):
+            k, v = line.split(':')
+            metadata[k.strip('#')] = v.strip('\n')
```
tomMoral, line 46: so thoughful :)

benchopt/utils/files.py
```diff
@@ -25,7 +31,30 @@ def uniquify_results(file_path):
...
             f"{file_path} already exists. Saving results to {alternative}"
-            )
+        )
         return alternative
     else:
         return file_path
+
+
+def read_results(file_path):
+    """Read benchmark results as pandas DataFrame."""
```
tomMoral, line 41: Need to explicit the return type as it is not only the dataframe.

benchopt/utils/files.py
```diff
@@ -25,7 +31,30 @@ def uniquify_results(file_path):
...
+    return df, metadata
+
+
+def write_results(df, file_path):
+    """Write benchmark results to CSV file."""
+    err, tag = _run_shell("git describe --tags --abbrev=0", return_output=True)
+    if err != 0:
+        if DEBUG:
+            print(err, tag)
+        tag = None
```
tomMoral, line 57: This should be a dedicated function.

tomMoral: Maybe in `sys_info.py`.

benchopt/utils/files.py
```diff
@@ -25,7 +31,30 @@ def uniquify_results(file_path):
...
+    with open(file_path, "r") as f:
+        metadata = {}
+        for line in takewhile(lambda s: s.startswith("#"), f):
+            k, v = line.split(':')
+            metadata[k.strip('#')] = v.strip('\n')
+        df = pd.read_csv(f, comment='#')
+    return df, metadata
+
+
+def write_results(df, file_path):
```
tomMoral, line 51: maybe give a dictionary of tags? that are going to be written here.

benchopt/utils/files.py
tomMoral: Thinking back, why didn't we put this directly in `sys-info`? this would avoid this change and be directly accessible as the other sys info?

benchopt/plotting/generate_html.py
```diff
@@ -178,7 +178,7 @@ def get_results(fnames, kinds, root_html, benchmark_name, copy=False):
     for fname in fnames:
         print(f"Processing {fname}")
 
-        df = pd.read_csv(fname)
+        df = pd.read_csv(fname, comment='#')
```
tomMoral, line 181: use read_results?
```suggestion
        df, _ = read_results(fname)
```

tomMoral (approved):
this is nice. Just having second though, why didn't we put this info in `sys_info`?

I think it is probably okay like this but wondering what is the best option.

benchopt/runner.py
```diff
@@ -334,7 +334,8 @@ def run_benchmark(benchmark, solver_names=None, forced_solvers=None,
     else:
         save_file = output_dir / f"{output_name}.csv"
         save_file = uniquify_results(save_file)
-    df.to_csv(save_file)
+
```
mathurinm, line 337:
```suggestion
```

benchopt/utils/files.py
mathurinm: yes, we can

tomMoral (approved):
LGTM!

PR #423, opened 2022-06-08
[WIP] FIX - Handle iterables parameters in ``Objective``

### LINKED ISSUE #422: BUG - Error when one of the ``Objective`` parameter is list
I was performing a benchmark of weighted group lasso, which in particular has an ``objective`` with some of the parameters as a list (e.g. the weights).

This breaks down the run of ``benchopt`` precisely here,

https://github.com/benchopt/benchopt/blob/a7faa13d298f7890e58963908b61a41e755557c2/benchopt/base.py#L395

with the following error log

```shell
  File "c:\users\hp\desktop\celer-repo\benchopt\benchopt\base.py", line 395, in set_dataset
    if parameters[key] != getattr(self, key):
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

A solution to remedy this problem is to handle the case of iterable separably.

Badr-MOUFAD:
fixes #422

### DIFF
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -392,7 +392,11 @@ class BaseObjective(ParametrizedNameMixin, DependenciesMixin):
             parameters[key] = getattr(self, key)
         self.set_data(**data)
         for key in self._parameters:
-            if parameters[key] != getattr(self, key):
+            has_changed = (any(parameters[key] != getattr(self, key))
+                           if hasattr(parameters[key], '__iter__')
+                           else parameters[key] != getattr(self, key))
+
+            if has_changed:
                 raise ValueError(
                     f"Parameter {key} has been changed from {parameters[key]} "
                     f"to {getattr(self, key)}. "


benchopt/base.py
```diff
@@ -392,7 +392,11 @@ def set_dataset(self, dataset):
             parameters[key] = getattr(self, key)
         self.set_data(**data)
         for key in self._parameters:
-            if parameters[key] != getattr(self, key):
+            has_changed = (any(parameters[key] != getattr(self, key))
+                           if hasattr(parameters[key], '__iter__')
+                           else parameters[key] != getattr(self, key))
```
agramfort, line 397: can you give an example of parameter value that creates a pb?

Badr-MOUFAD: yes sure

```python
# ...
    parameters = {
        'weights': [np.ones(n_groups), abs(np.random.randn(n_groups))],   
# ...
```

agramfort: I would use 

# ...
    parameters = {
        'weights': ["ones", "randn"],   
# ...

and do the set in the run function. Also it will allow to fix the seed of randn

Badr-MOUFAD: Agree!
but what about

```python
# ...
    parameters = {
        'groups': [[5, 7, 3], [5, 5, 2, 3]], # contiguous groups of 15 features
# ...
```

Anyways, I think the case of iterable params arises rarely. So no need to handle it at all.

agramfort:
I would say unless we have a very compelling argument let's keep it simple
for now

tomMoral:
Actually, other persons encounter the exact same issue not long ago.
While I agree that it is usually better to just use words, I think the fix is simple enough so we should merge it.

benchopt/base.py
tomMoral, line 397: A more comprehensive implem, which would also work for list.

```suggestion
            if hasattr(parameters[key], '__iter__'):
                has_changed = any(
                    p != a for p, a in zip(parameters[key], getattr(self, key)
                )
            else:
                has_changed = parameters[key] != getattr(self, key)
```

Badr-MOUFAD: This is a buggy implementation,

Indeed,
```python
[1, 2] != [1, 2, 3]
# output: True

any(i != j for i, j in zip([1, 2], [1, 2, 3]))
# output: False

# [(i, j) for i, j in zip([1, 2], [1, 2, 3])] gives [(1, 1), (2, 2)]
```

So your code won't produce the desired behavior.

tomMoral: Indeed

Badr-MOUFAD: I made a small change in the implementation.
Hope it's more comprehensible.

benchopt/base.py
```diff
@@ -392,7 +392,11 @@ def set_dataset(self, dataset):
             parameters[key] = getattr(self, key)
         self.set_data(**data)
         for key in self._parameters:
-            if parameters[key] != getattr(self, key):
+            has_changed = (parameters[key] != getattr(self, key)
+                           if not hasattr(parameters[key], '__iter__')
+                           else any(parameters[key] != getattr(self, key)))
```
tomMoral, line 397: Actually I think we can even simplify:
```suggestion
            # protect for cases where parameters[key] are numpy arrays.
            has_changed = any(parameters[key] != getattr(self, key))
```

Note that thinking back, this would not work for 2D ndarray.

Another way would be:
```python
has_changed = parameters[key] != getattr(self, key)
if not isinstance(has_changed, bool):
     has_changed = has_changed.any()
```

But not sure if it cover all potential cases.

Badr-MOUFAD: ### for the fist suggestion
>``has_changed = any(parameters[key] != getattr(self, key))``

This would fail in the basic case

```python
any(1 != 2)  # TypeError: 'bool' object is not iterable 
```

### for the second suggestion

There is nothing that prevents packages from creating their own variable types (e.g numpy has its own bool type ``np.bool_``)

Therefore the second suggestion would fail also

```python
parameters = {
    'a': np.bool_(1),
    'b': np.bool_(0),,
}

isinstance(parameters['a'] != parameters['b'], bool)  # Output: False

# Applying the second solution would raise an exception 
# TypeError: 'numpy.bool_' object is not iterable
```

### Another suggestion

I believe that checking ``__iter__`` is less prone to bugs. I suggest a mixture between your solution and mine

```python
has_changed = parameters[key] != getattr(self, key)
if hasattr(has_changed, '__iter__'):
     has_changed = any(has_changed)
```

tomMoral: Aha for the first one, I got owned by the use of `ipython` which imports symbols from `numpy`.
Indeed the basic `any` does not work with non iterable.

Let's go with the last suggestion then.

tomMoral (approved):
LGTM, we can always improve this if we have more use case comming our way.

PR #418, opened 2022-06-02
[DRAFT] Improve plot display

Melvin-klein:
THIS IS A DRAFT VERSION.

Instead of writing HTML plot for each combination of data, we use a unique plot and update it using JS.

- The constant `TEMPLATE_LOCAL_RESULT` seems to be unused.

- The link https://cdn.plot.ly/plotly-latest.min.js returns PlotlyJS V1 instead of V2.

PR #424, opened 2022-06-09
Update html template

Melvin-klein:
The aim of this pull request is to propose an HTML template that displays all the necessary information on the screen in order to avoid having to scroll through the page.

Some details are missing at the moment.

This PR is based on #418, that's why there are many modifed files.

The page adapts to the screen size but not yet for cell phones and tablets.

First screenshot is on a big screen :
big_screen

Second screeshot is on a 13 inches screen :
13inches


Melvin-klein:
@tomMoral

tomMoral (approved):
LGTM! thx @Melvin-klein

PR #428, opened 2022-06-17
Custom legend in result template

Melvin-klein:
Addition of a custom legend and removal of Plotly's legend to fix the chart size and the overflow of the Plotly legend.

Plotly only allows to set the size of the plot container (plot + legend) and don't provide any method to choose the plot and the legend size seperatly. So the solution is to remove Plotly's legend and create our own. Without the Plotly's legend, we can choose the size for the plotly container which now means the size of the plot himself and also the size of the legend as is it our own.

Size issue and legend overflow issue can be visible here : [Resnet](https://benchopt.github.io/results/benchmark_resnet_classif_benchmark_resnet_classif_benchopt_run_2022-05-03_10h31m54.html)


agramfort:
@Melvin-klein can we still click the legend to toggle the visibility of each curve?

Melvin-klein:
@agramfort Yes and also double click. I kept all the behaviours

agramfort:
great !

agramfort:
@Melvin-klein you have a conflict. When merge I will test and merge. Thx

agramfort (approved):
works like a charm for me !

🎉 👏

tomMoral (approved):
Works like a charm for me too!!! thanks a lot @Melvin-klein  :)

PR #429, opened 2022-06-17
Trim infinite values for HTML plotting

Melvin-klein:
This PR trims infinite values.

Some dataset have infinite value in their objective_columns. The points that have the coordinates like (x, infinite) are not displayed. Such points continue to be counted in the xlim. So if these points are far on the x-axis then they will enlarge the xlim without being seen. This has the effect of reducing the size of the other curves on the x-axis.

Example : [Logreg L2](https://benchopt.github.io/results/benchmark_logreg_l2_benchmark_logreg_l2_benchopt_run_2021-03-18_02h11m46.html)


tomMoral (approved):
LGTM

PR #431, opened 2022-06-20
Update problems

jolars:
This PR

* converts the list of optimization problems into a table
* adds a column with results for the problems
* updates the tables with links to more problems

This is what is looks like now:

image

This is what it would look like after this PR:

image


mathurinm (approved):

agramfort:
Thx @jolars

PR #432, opened 2022-06-21
Embed js, css, svg into Mako template

Melvin-klein:
Instead of loading JS file by the HTML <script> tag (and css/svg files), we write them directly into the template using the result dictionary.

As every things are written into the template, the HTML result file can be opened without any dependencies.


agramfort (approved):

tomMoral (approved):
Very neat! thx @Melvin-klein

PR #433, opened 2022-06-27
Compress data

Melvin-klein:
Parquet files are smaller than CSV files. So we change the CSV output by a Parquet file output.
Compatibility with previous CSV files is kept.

There is a new cli command to convert Parquet files to CSV file (or html or json) : 

`$ benchopt convert [parquet_file] --to [requested_format]`

TODO:
- [x] Choose between pyarrow and fastparquet packages
- [x] Test that everything works as expected with previous CSV files

### DIFF
--- a/benchopt/cli/process_results.py
+++ b/benchopt/cli/process_results.py
@@ -130,3 +130,30 @@ def generate_results(patterns=(), benchmarks=(), root=None, display=True):
     plot_benchmark_html_all(
         patterns=patterns, benchmarks=benchmarks, root=root, display=display
     )
+
+
+@process_results.command(
+    help="Convert a result file of type parquet into the requested type."
+)
+@click.argument('parquet_file',
+                default=Path.cwd(),
+                type=click.Path(exists=True))
+@click.option('--to', '-t', 'requested_formats',
+              metavar="<to>", multiple=True, type=str,
+              help="Choose the conversion format (CSV, JSON, Excel, etc.).")
+def convert(parquet_file, requested_formats):
+    import os
+    import pandas as pd
+    # remove parquet file extension
+    filename = os.path.splitext(parquet_file)[0]
+    accepted_types = ['csv', 'json', 'html']
+    methods = {}
+    for accepted in accepted_types:
+        methods[accepted] = "to_" + accepted
+
+    df = pd.read_parquet(parquet_file)
+    for requested_format in requested_formats:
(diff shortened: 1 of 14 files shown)


agramfort (commented):
I would not have bothered adding a convert tool. I would just use parquet from now on and read the old .csv if present

benchopt/cli/process_results.py
```diff
@@ -130,3 +130,30 @@ def generate_results(patterns=(), benchmarks=(), root=None, display=True):
...
+@process_results.command(
+    help="Convert a result file of type parquet into the requested type."
+)
+@click.argument('parquet_file',
+                default=Path.cwd(),
+                type=click.Path(exists=True))
+@click.option('--to', '-t', 'requested_formats',
+              metavar="<to>", multiple=True, type=str,
+              help="Choose the conversion format (CSV, JSON, Excel, etc.).")
+def convert(parquet_file, requested_formats):
```
tomMoral, line 144: Not sure we need this.
It seems quite easy and it will be hard to test so not sure we want to maintain this?

setup.cfg
```diff
@@ -39,6 +39,8 @@ install_requires =
     plotly>=4.12
     pyyaml
     line-profiler
+    pyarrow
+    fastparquet
```
tomMoral, line 43: we need to chose no?

tomMoral (commented):
A few comments but mostly this looks good.
for large files, what is the order of magnitude of gain?

Melvin-klein:
> I would not have bothered adding a convert tool. I would just use parquet from now on and read the old .csv if present

Ok I will remove this command

benchopt/tests/__init__.py
```diff
@@ -61,12 +61,12 @@ def __exit__(self, exc_class, value, traceback):
 
         # Make sure to delete all the result that created by the run command.
         self.result_files = re.findall(
-            r'Saving result in: (.*\.csv)', self.output
+            r'Saving result in: (.*\.csv|.*\.parquet)', self.output
```
tomMoral, line 64: We only save in `parquet` now, no? Not sure if we could have cases where both are produced/
```suggestion
            r'Saving result in: (.*\.parquet)', self.output
```

tomMoral (commented):
Just an extra comment bu otherwise, lgtm

tomMoral (approved):
LGTM, just 2 nitpicks

PR #435, opened 2022-07-04
FIX non-diverging negative losses

tomMoral:
This should fix the issue from benchopt/benchmark_jointdiag#4.

The issue is that we rescale the delta by `loss[0]` so for negative losses this reverses the dynamic. I added some absolute value to avoid this.
Let me know if this fixes the issue.


agramfort (approved):
I confirm it fixes my pb

🙏 @tomMoral

PR #438, opened 2022-07-05
Fix parquet file publishing

### LINKED ISSUE #437: migration to parquet broke `benchopt publish`
cc @Melvin-klein 

it complains about utf-8. As parquet files are binary files the code
needs to be adjusted.

Melvin-klein:
Fix #437 

@agramfort

### DIFF
--- a/benchopt/utils/github.py
+++ b/benchopt/utils/github.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 from github import Github
 from github import GithubException
+import pandas as pd
 
 
 BENCHOPT_RESULT_REPO = 'benchopt/results'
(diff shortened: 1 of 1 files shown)


benchopt/utils/github.py
```diff
@@ -26,7 +26,11 @@ def publish_result_file(benchmark_name, file_path, token):
             f"Could not upload file {file_to_upload}."
         )
 
-    file_content = pd.read_parquet(file_to_upload).to_csv()
+    if file_to_upload.suffix == ".csv":
+        with file_to_upload.open('r') as f:
+            file_content = f.read()
+    else:
+        file_content = pd.read_parquet(file_to_upload).to_csv()
```
tomMoral, line 33: Isn't is possible to keep parquet file here? you shoudl probably read it in binary but it would be more convenient to upload the results in a compress format.

Moreover, this would produce deprecation warning (we use the same code to generate result on the website and locally).

agramfort: agreed. We should upload the parquet files

benchopt/utils/github.py
```diff
@@ -24,8 +25,12 @@ def publish_result_file(benchmark_name, file_path, token):
         raise FileNotFoundError(
             f"Could not upload file {file_to_upload}."
         )
-    with file_to_upload.open('r') as f:
-        file_content = f.read()
+
+    if file_to_upload.suffix == ".csv":
+        with file_to_upload.open('r') as f:
```
agramfort, line 30: to have a binary file you would do:
```suggestion
        with file_to_upload.open('rb') as f:
```

Melvin-klein: Thanks

tomMoral (approved):
LGTM

PR #443, opened 2022-07-18
ENH add option for positive simulated data in make_correlated_data

cassiofragadantas:
This feature is necessary for the NNLS benchmark (see benchopt/benchmark_nnls#12)

### DIFF
--- a/benchopt/datasets/simulated.py
+++ b/benchopt/datasets/simulated.py
@@ -7,7 +7,8 @@ from ..utils.checkers import check_random_state
 
 def make_correlated_data(
         n_samples=100, n_features=50, n_tasks=1, rho=0.6, snr=3,
-        w_true=None, density=0.2, X_density=1, random_state=None):
+        w_true=None, density=0.2, X_density=1, random_state=None,
+        pos_data=False):
     r"""Generate a linear regression with decaying correlation for the design
     matrix :math:`\rho^{|i-j|}`.
 
@@ -51,6 +52,8 @@ def make_correlated_data(
         make the randomness deterministic.
     X_density: float in ]0, 1]
         Proportion of elements of X which are non-zero.
+    pos_data: boolean
+        Generate non-negative data (both design matrix and observation vector)
 
     Returns
     -------
@@ -101,6 +104,10 @@ def make_correlated_data(
     else:
         if w_true.ndim == 1:
             w_true = w_true[:, None]
+
+    if pos_data:
+        X = abs(X)
+        w_true = abs(w_true)
 
     Y = X @ w_true
     noise = rng.randn(n_samples, n_tasks)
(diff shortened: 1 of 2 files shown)


mathurinm:
I don't know if it's worth adding one parameter. Do you envision usecases apart from NNLS ? Otherwise this can be handled directly in `simulated.py` in the NNLS benchmark (use the function to generate X, w, then cast to positive and generate y manually) 

My 2 cents. It seems @agramfort disagrees here : https://github.com/benchopt/benchmark_nnls/pull/12#issuecomment-1181757270

cassiofragadantas:
I'm ok with both solutions. As you prefer.

benchopt/datasets/simulated.py
```diff
@@ -7,7 +7,8 @@
 
 def make_correlated_data(
         n_samples=100, n_features=50, n_tasks=1, rho=0.6, snr=3,
-        w_true=None, density=0.2, X_density=1, random_state=None):
+        w_true=None, density=0.2, X_density=1, random_state=None,
+        pos_data=False):
```
agramfort, line 11:
```suggestion
        positive=False):
```

benchopt/datasets/simulated.py
```diff
@@ -51,6 +52,8 @@ def make_correlated_data(
         make the randomness deterministic.
     X_density: float in ]0, 1]
         Proportion of elements of X which are non-zero.
+    pos_data: boolean
```
agramfort, line 55:
```suggestion
    positive: bool
```

benchopt/datasets/simulated.py
```diff
@@ -102,6 +105,10 @@ def make_correlated_data(
         if w_true.ndim == 1:
             w_true = w_true[:, None]
 
+    if pos_data:
+        X = abs(X)
+        w_true = abs(w_true)
```
agramfort, line 110:
```suggestion
    if positive:
        X = np.abs(X)
        w_true = np.abs(w_true)
```

agramfort (commented):
also can you add or update the test?

cassiofragadantas:
will do

agramfort:
I would not do y=abs(y) just to keep things simpler

mathurinm:
I am -1 on merging this one, IMO it's only needed in the NNLS benchmark and thus should be implemented over there rather than in the main codebase.

@tomMoral if you agree we can close

tomMoral:
Yes I agree this is very specific and having too many parameters make it harder to use/maintain so let's not add this as one can simply call `make_correlated`, take abs value and regenerate `Y`.

tomMoral:
Thanks a lot @cassiofragadantas anyway, sorry to have you made the code for not using it in the end.

cassiofragadantas:
No problem ;) I will make the modifications directly on the NNLS ` simulated.py`  as suggested by @mathurinm

PR #444, opened 2022-07-18
[Draft] Fix html artifact

Melvin-klein:
Create multiple workflow in single file.


.github/workflows/expose_link.yml
```diff
@@ -11,12 +11,16 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           artifact-path: 0/result.html
           circleci-jobs: build_docs
-
-  circleci_docs_artifacts_redirector_job:
+          job-title: benchmark result artifact
+---
+name: Expose docs artifact
+on: [status]
+jobs:
```
tomMoral, line 18: In the end, I don't think we actually need to have 2 workflows. Both jobs are running on master so I would say this is working. Note that we cannot see what is going on in this PR because it only changes when merge in master.
```suggestion
          job-title: benchmark result artifact
```

.github/workflows/expose_link.yml
```diff
@@ -11,12 +11,16 @@ jobs:
...
+  circleci_doc_artifact_redirector_job:
```
tomMoral, line 19:
```suggestion
  circleci_docs_artifact_redirector_job:
```

PR #474, opened 2022-11-08
ENH - Aggregate similar solvers in legend

Badr-MOUFAD:
This enhances the plotting part.

## main contributions

- Aggregate solvers in legend under the same name
- enable hiding/showing solvers in legend and plots

## check new features

Run ``snippets_gen_html.py`` to generate HTML.

## next step

- enhance styling
- more complete filtering tool

--------
P.S
- similar solvers: same solver run with different parameters
- ``pseud_bench`` folder is for checks

### DIFF
--- a/benchopt/plotting/html/static/hover_index.css
+++ b/benchopt/plotting/html/static/hover_index.css
@@ -317,3 +317,16 @@ input:checked+.slider:before {
 .space-r-2 > * {
   margin-right: 0.5rem;
 }
+
+.solver-accordion {
+  background-color: #eee;
+  color: #444;
+  cursor: pointer;
+  padding: 18px;
+  width: 100%;
+  border: none;
+  text-align: left;
+  outline: none;
+  font-size: 15px;
+  transition: 0.4s;
+}
--- a/benchopt/plotting/html/static/result.js
+++ b/benchopt/plotting/html/static/result.js
@@ -536,14 +536,64 @@ const makeLegend = () => {
 
   legend.innerHTML = '';
 
-  Object.keys(data().solvers).forEach(solver => {
+  var aggregateName = aggregateButton = aggregateDiv =  null;
+
+  for(solver of data().solvers) {
+    const solverName = solver.split("[")[0];
     const color = data().solvers[solver].color;
     const symbolNumber = data().solvers[solver].marker;
 
-    legend.appendChild(createLegendItem(solver, color, symbolNumber));
-  });
+    if(solverName === aggregateName) {
+      aggregateDiv.appendChild(createLegendItem(solver, color, symbolNumber));
+      continue;
+    }
+
+    // create button to toggle accordion
+    aggregateButton = document.createElement("button");
+    aggregateButton.innerText = solverName;
+    aggregateButton.className = 'solver-accordion';
+
+    // create div to gather the solvers with other bench params
+    aggregateDiv = document.createElement("div");
+    aggregateDiv.style.display = state().hidden_boxes.includes(current_solver) ? "none" : "block"; 
+
+    // add event handler for accordion
+    aggregateButton.addEventListener('click', function () {
+      solversInAggregate = getSolverInAggregate(aggregateDiv);
+
+      if (aggregateDiv.display === "block") {  
+        setState({
+          hidden_solvers: state().hidden_solvers.concat(solversInAggregate),
+          disabled_solvers_accordions: state().hidden_boxes.concat(solverName)
+        });
+      } 
+      else {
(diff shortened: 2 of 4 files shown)


benchopt/plotting/html/static/hover_index.css
```diff
@@ -317,3 +317,16 @@ input:checked+.slider:before {
...
+  color: #444;
+  cursor: pointer;
+  padding: 18px;
+  width: 100%;
+  border: none;
+  text-align: left;
+  outline: none;
+  font-size: 15px;
+  transition: 0.4s;
+}
```
tomMoral, line 332:
```suggestion
}

```

Badr-MOUFAD:
@tomMoral, I uploaded the feature preview instead.
Have a look at https://github.com/benchopt/benchopt/pull/474#issue-1440491550

Badr-MOUFAD:
@tomMoral, I consider your remarks in the last changes
pls have a look on the feature preview https://github.com/benchopt/benchopt/pull/474#issue-1440491550

bmalezieux:
> When only one is available, I would not create a group no?

I think it is better to keep the group structure to preserve homogeneity. Otherwise, the result looks a bit weird (see the Lasso benchmark in #474)

Badr-MOUFAD:
> I think it is better to keep the group structure to preserve homogeneity. Otherwise, the result looks a bit weird (see the Lasso benchmark in #474)

I managed to construct groups for solvers that are run with multiple configurations of parameters
@bmalezieux, could you check whether it renders well for you?

- [Resnet benchmark](https://badr-moufad.github.io/share-benchmarks/aggregate_plots/generated_df_resnet_bench.html)
- [Lasso Benchmark](https://badr-moufad.github.io/share-benchmarks/aggregate_plots/generated_df_lasso_bench.html)

bmalezieux:
> > I think it is better to keep the group structure to preserve homogeneity. Otherwise, the result looks a bit weird (see the Lasso benchmark in #474)
> 
> I managed to construct groups for solvers that are run with multiple configurations of parameters @bmalezieux, could you check whether it renders well for you?
> 
> * [Resnet benchmark](https://badr-moufad.github.io/share-benchmarks/aggregate_plots/generated_df_resnet_bench.html)
> * [Lasso Benchmark](https://badr-moufad.github.io/share-benchmarks/aggregate_plots/generated_df_lasso_bench.html)

Actually, I preferred the last version with one block per solver, even with a single configuration. What disturbs me is that the solvers should be represented in the same way in the legend, and this is not the case here in my opinion.

Badr-MOUFAD:
WDYT @tomMoral?

tomMoral:
I agree that we should find a consistent way to do this but I don't like the fact that we have to scroll to have the full legend, even when all legend labels could be put on 2/3 lines. This is something we changed 6 months ago so it feels that we are going back.
We need to find a better way to present this than big blocks that stack over each other.

Maybe simply make the parent block also a float, that do not take the full screen. That way, solvers with one config will jsut have an extra gray box above with the name of the solver and multiple solver will have a box around grouping them together?

Badr-MOUFAD:
WDYT @tomMoral ?

Screenshot from 2022-11-18 19-46-00

Badr-MOUFAD:
@bmalezieux any thoughts about https://github.com/benchopt/benchopt/pull/474#issuecomment-1320404474?

bmalezieux:
It looks good ! I think it would be even better to add the name of the solver above solo solvers (it is what Thomas meant if I understand the drawing correctly).

tomMoral:
Yes I like the style!! :)
As @bmalezieux  said, I would keep the box above the solo solvers.
Also, could you make it so that if we fall a box, it stay the same width?

bmalezieux:
By the way, I find it clearer to put the style in the css file and not in the class name.

Badr-MOUFAD:
> Also, could you make it so that if we fall a box, it stay the same width?

I'm afraid it won't be possible.

Instead, as discussed with @bmalezieux, I made the aggregate of the solver grayish when it's disabled.

Links to to preview:

- [bench resnet](https://badr-moufad.github.io/share-benchmarks/aggregate_plots/generated_df_resnet_bench.html)
- [bench lasso](https://badr-moufad.github.io/share-benchmarks/aggregate_plots/generated_df_lasso_bench.html)

mathurinm:
I'm late to the party, but instead of graying the unselected solvers, would it be possible to collapse the unselected block, displaying only its title ? If there are many solver sin a block and they're unselected, one still has to croll a lot to go to the next block

PR #477, opened 2022-11-09
FIX add the r conda channel (WIP)

jolars:
This should fix installation issues for r packages when using mamba. But since I seem to be unable to get my mamba setup up and running, I haven't been able to reproduce the original issue that was reported in https://github.com/benchopt/benchmark_slope/pull/3#issuecomment-1308452109, so I need someone to verify that this is all that we need. @mathurinm ?

### DIFF
--- a/benchopt/utils/conda_env_cmd.py
+++ b/benchopt/utils/conda_env_cmd.py
@@ -127,7 +127,7 @@ def create_conda_env(
         # see https://stackoverflow.com/questions/70098418/
         _run_shell_in_conda_env(
             f"{CONDA_CMD} config --env --prepend channels nodefaults " +
-            "--prepend channels conda-forge", env_name,
+            "--prepend channels conda-forge --prepend channels r", env_name,
             capture_stdout=quiet)
         if empty:
             return
--- a/benchopt/utils/conda_env_cmd.py
+++ b/benchopt/utils/conda_env_cmd.py
@@ -21,6 +21,7 @@ BENCHOPT_ENV = """
 channels:
   - conda-forge
   - nodefaults
+  - r
 dependencies:
   - python=3.9
   - numpy
@@ -35,6 +36,7 @@ EMPTY_ENV = """
 channels:
   - conda-forge
   - nodefaults
+  - r
 """
 
 


mathurinm:
I tried, and it doesn't work so far: 

When I try to do it in my current env, which has rpy2 and r-base installed:

```
Looking for: ['r-slope', 'rpy2', 'r-base', 'scipy', 'r-matrix']

conda-forge/linux-64                                        Using cache
conda-forge/noarch                                          Using cache

Pinned packages:
  - python 3.10.*


Encountered problems while solving:
  - nothing provides requested r-slope

  File "/home/mathurin/mambaforge/envs/rpy2/bin/benchopt", line 8, in <module>
    sys.exit(benchopt())
  File "/home/mathurin/mambaforge/envs/rpy2/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/mathurin/mambaforge/envs/rpy2/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/mathurin/mambaforge/envs/rpy2/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mathurin/mambaforge/envs/rpy2/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mathurin/mambaforge/envs/rpy2/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/mathurin/workspace/benchopt/benchopt/cli/main.py", line 447, in install
    benchmark.install_all_requirements(
  File "/home/mathurin/workspace/benchopt/benchopt/benchmark.py", line 382, in install_all_requirements
    install_in_conda_env(
  File "/home/mathurin/workspace/benchopt/benchopt/utils/conda_env_cmd.py", line 201, in install_in_conda_env
    _run_shell_in_conda_env(
  File "/home/mathurin/workspace/benchopt/benchopt/utils/shell_cmd.py", line 129, in _run_shell_in_conda_env
    return _run_shell(
  File "/home/mathurin/workspace/benchopt/benchopt/utils/shell_cmd.py", line 67, in _run_shell
    raise RuntimeError(raise_on_error.format(output=output))
RuntimeError: Failed to conda install packages ('r-slope', 'rpy2', 'r-base', 'scipy', 'r-matrix')
```
(note also the quite confusing stacktrace at the beginning, where it says it's installed then no) 


When I try to create an env (`benchopt install --env -s rSLOPE`):

```
Preparing transaction: done
Verifying transaction: done
Executing transaction: done
Installing pip dependencies: | Ran pip subprocess with arguments:
['/home/mathurin/mambaforge/envs/benchopt_slope/bin/python', '-m', 'pip', 'install', '-U', '-r', '/tmp/condaenv.ugpwmxea.requirements.txt']
Pip subprocess output:
Obtaining file:///home/mathurin/mambaforge/envs/rpy2/lib/python3.10/site-packages (from -r /tmp/condaenv.ugpwmxea.requirements.txt (line 1))

Pip subprocess error:
ERROR: file:///home/mathurin/mambaforge/envs/rpy2/lib/python3.10/site-packages (from -r /tmp/condaenv.ugpwmxea.requirements.txt (line 1)) does not appear to be a Python project: neither 'setup.py' nor 'pyproject.toml' found.

failed

CondaEnvException: Pip failed

failed to create the environment.
  File "/home/mathurin/mambaforge/envs/rpy2/bin/benchopt", line 8, in <module>
    sys.exit(benchopt())
  File "/home/mathurin/mambaforge/envs/rpy2/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/mathurin/mambaforge/envs/rpy2/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/mathurin/mambaforge/envs/rpy2/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mathurin/mambaforge/envs/rpy2/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mathurin/mambaforge/envs/rpy2/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/mathurin/workspace/benchopt/benchopt/cli/main.py", line 443, in install
    create_conda_env(env_name, recreate=recreate, quiet=quiet)
  File "/home/mathurin/workspace/benchopt/benchopt/utils/conda_env_cmd.py", line 121, in create_conda_env
    _run_shell(
  File "/home/mathurin/workspace/benchopt/benchopt/utils/shell_cmd.py", line 67, in _run_shell
    raise RuntimeError(raise_on_error.format(output=output))
RuntimeError
```

mathurinm:
Ah, I forgot. It's because prending in the yaml file does not configure the channels for future installs. It needs to be done in command line as I did.
Now it works in the dedicated env (eg `benchopt_benchmark_slope`), but not when installing in the current env because we cannot/do not/should not add new channels. What's your take on this @tomMoral ?

tomMoral:
I am -.5 on this one because it means that all env will need to update the r channel while most won't use it. And as @mathurinm said, it won't work for pre existing env.

To me, the way to deal with this is to specify the channel in the command.
I think adding an extra entry in dependencies with `'-c r'` does the trick.
Else, I thought I had implemented something like `'chan#package'` in the reps to allow specifying a channel.

WDYT?

jolars:
I agree that we probably shouldn't add the channel i the majority of the packages don't need it. Also note that I can install r-slope fine just using conda-forge, so this seems to be somehow related only to mamba if I understand correctly.

But adding `-c r` to dependencies, even if it works, seems a little hacky to me. I don't know what you mean with `chan#package`, but otherwise I think just adding a new entry such as `conda_channels` would be a clear, solid solution. Maybe that would also allow us to avoid using conda-forge sometimes when it's not needed.

mathurinm:
I agree with @tomMoral now, given that many R packages can be installed from `conda-forge` I don't think it's worth adding the `r` channel as default (+ to try and add it in existing env if there is no `--env` flag, which seems bad practice).

I tried `    requirements = ["r-base", "rpy2", "-c r r-slope", "r-matrix", "scipy"]` and it does not work, so we'd need a more advanced solution. `    requirements = ["r-base", "rpy2", "r#r-slope", "r-matrix", "scipy"]` also fails

tomMoral:
I think you need to do `requirements = ["-c r", "r-slope", ...]` for it to work no?

mathurinm:
I think ` requirements = ["-c r", "r-slope", ...]` would work but is very hacky, as I'd expect the order of requirements to not matter.

On the other hand I like the `"c#p"` syntax very much

tomMoral:
should we put the channel before or after? I would day after but not sure

PR #479, opened 2022-11-09
PLT - add visualization of objective as function of ``stop_val``

Badr-MOUFAD:
This adds a toggle to enable visualizing the objectives as function `stop_val`, namely ``n_iter``.

[Link to a dummy benchmark ](https://badr-moufad.github.io/share-benchmarks/stop_val_plot/validate-plotting-stop-val_benchopt_run_2022-11-09_20h57m17.html) to preview the feature. (credit to @mathurinm for the idea of the benchmark).
A link to the [benchmark repo ](https://github.com/Badr-MOUFAD/validate-plotting-stop-val) to reproduce.

-----------------------

PR related to #287

### DIFF
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -182,12 +182,12 @@ def shape_solvers_for_html(df, objective_column):
         df_filtered = df_filtered.dropna(subset=[objective_column])
 
         q1, q9 = compute_quantiles(df_filtered)
+        groupby_stop_val_median = df_filtered.groupby('stop_val').median()
         solver_data[solver] = {
             'scatter': {
-                'x': df_filtered.groupby('stop_val')['time']
-                                .median().tolist(),
-                'y': df_filtered.groupby('stop_val')[objective_column]
-                                .median().tolist(),
+                'x': groupby_stop_val_median['time'].tolist(),
+                'y': groupby_stop_val_median[objective_column].tolist(),
+                'stop_val': groupby_stop_val_median.index.tolist(),
                 'q1': q1.tolist(),
                 'q9': q9.tolist(),
             },
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -185,6 +185,17 @@ def shape_solvers_for_html(df, objective_column):
         color, marker = get_solver_style(solver)
 
         stopping_strategy = df_filtered['stopping_strategy'].unique()
+
+        if len(stopping_strategy) != 1:
+            found_stopping_strategies = ', '.join(
+                f"`{item}`" for item in stopping_strategy)
+
+            raise Exception(
+                "Solver can be run using only one stopping strategy. "
+                f"Expected one stopping strategy "
+                f"but found {found_stopping_strategies}"
+            )
+
         stopping_strategy = stopping_strategy[0]
 
         solver_data[solver] = {
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -183,7 +183,10 @@ def shape_solvers_for_html(df, objective_column):
         q1, q9 = compute_quantiles(df_filtered)
         groupby_stop_val_median = df_filtered.groupby('stop_val').median()
         color, marker = get_solver_style(solver)
-        stopping_strategy = df_filtered['stopping_strategy'][0]
+
+        stopping_strategy = df_filtered['stopping_strategy'].unique()
+        stopping_strategy = stopping_strategy[0]
+
         solver_data[solver] = {
             'scatter': {
                 'x': groupby_stop_val_median['time'].tolist(),
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -183,6 +183,7 @@ def shape_solvers_for_html(df, objective_column):
         q1, q9 = compute_quantiles(df_filtered)
         groupby_stop_val_median = df_filtered.groupby('stop_val').median()
         color, marker = get_solver_style(solver)
+        stopping_strategy = df_filtered['stopping_strategy'][0]
         solver_data[solver] = {
             'scatter': {
                 'x': groupby_stop_val_median['time'].tolist(),
@@ -195,7 +196,8 @@ def shape_solvers_for_html(df, objective_column):
                 **computeBarChartData(df, objective_column, solver)
             },
             'color': color,
-            'marker': marker
+            'marker': marker,
+            'stopping_strategy': stopping_strategy
         }
 
     return solver_data
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -188,7 +188,8 @@ def shape_solvers_for_html(df, objective_column):
 
         if len(stopping_strategy) != 1:
             found_stopping_strategies = ', '.join(
-                f"`{item}`" for item in stopping_strategy)
+                f"`{item}`" for item in stopping_strategy
+            )
 
             raise Exception(
                 "Solver can be run using only one stopping strategy. "
--- a/benchopt/plotting/html/static/result.js
+++ b/benchopt/plotting/html/static/result.js
@@ -136,7 +136,7 @@ const getScatterCurves = () => {
       legendgroup: solver,
       hovertemplate: solver + ' <br> (%{x:.1e},%{y:.1e}) <extra></extra>',
       visible: isVisible(solver) ? true : 'legendonly',
-      x: data(solver).scatter.x,
+      x: (!state().xaxis_with_iteration) ? data(solver).scatter.x :  [...Array(data(solver).scatter.x.length).keys()],
       y: useTransformer(data(solver).scatter.y, 'y', data().transformers),
     });
 
--- a/benchopt/plotting/html/static/result.js
+++ b/benchopt/plotting/html/static/result.js
@@ -153,18 +153,14 @@ const getScatterCurves = () => {
       y: useTransformer(data(solver).scatter.y, 'y', data().transformers),
     });
 
+    // skip plotting quantiles if xaxis is not time
+    // as stop_val are predefined and hence deterministic 
+    if(xaxisType !== "time") {
+      return
+    }
+
     if (state().with_quantiles) {
       // Add shaded area for each solver, with proper style and visibility.
-
-      solverStoppingStrategy = data(solver)['stopping_strategy'];
-
-      // plot only solvers that were stopped using xaxis type
-      // plot all solver if xaxis type is `time`
-      if(xaxisType !== "time" && solverStoppingStrategy !== xaxisType) {
-        return
-      }
-
-      scatterXaxisProperties = xaxisType === "time" ? ['q1', 'q9'] : ['stop_val', 'stop_val'];
   
       curves.push({
         type: 'scatter',
@@ -177,7 +173,7 @@ const getScatterCurves = () => {
         legendgroup: solver,
         hovertemplate: '(%{x:.1e},%{y:.1e}) <extra></extra>',
         visible: isVisible(solver) ? true : 'legendonly',
-        x: data(solver).scatter[scatterXaxisProperties[0]],
+        x: data(solver).scatter['q1'],
         y: useTransformer(data(solver).scatter.y, 'y', data().transformers),
       }, {
         type: 'scatter',
@@ -191,7 +187,7 @@ const getScatterCurves = () => {
         legendgroup: solver,
         hovertemplate: '(%{x:.1e},%{y:.1e}) <extra></extra>',
         visible: isVisible(solver) ? true : 'legendonly',
-        x: data(solver).scatter[scatterXaxisProperties[1]],
+        x: data(solver).scatter['q9'],
         y: useTransformer(data(solver).scatter.y, 'y', data().transformers),
       });
     }
--- a/benchopt/plotting/html/static/result.js
+++ b/benchopt/plotting/html/static/result.js
@@ -121,7 +121,19 @@ const getScatterCurves = () => {
   const curves = [];
 
   // For each solver, add the median curve with proper style and visibility.
+  xaxisType = state().xaxis_type
+
   getSolvers().forEach(solver => {
+    solverStoppingStrategy = data(solver)['stopping_strategy'];
+
+    // plot only solvers that were stopped using xaxis type
+    // plot all solver if xaxis type is `time`
+    if(xaxisType !== "time" && solverStoppingStrategy !== xaxisType) {
+      return
+    }
+
+    ScatterXaxisProperty = xaxisType === "time" ? 'x' : 'stop_val';
+
     curves.push({
       type: 'scatter',
       name: solver,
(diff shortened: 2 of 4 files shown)


Badr-MOUFAD:
any thoughts @mathurinm?

mathurinm:
Thanks for having a look @Badr-MOUFAD ! I can't see anything for the other objective, is it normal ?
https://hal.archives-ouvertes.fr/hal-01941152v2/document

image


How does this handle combination of solvers that use iteration stop_val and tol stop val ?

Badr-MOUFAD:
> Thanks for having a look @Badr-MOUFAD ! I can't see anything for the other objective, is it normal ?

@mathurinm, the values of the objective are negative and hence can't be visualized using a log-y scale.
Try linear scale or in the sidebar set the objective to `Affine[with_exp=True]`

jeandut:
Some minor comments after @mathurinm's invitation (to take with a grain of salt): 
- `Stopping Criterion` is not intuitive for the name of such a toggle, an alternative name would at least include "X-axis" somewhere
- `Stop_value` is also unintuitive #iterations is better
-  I would avoid scientific notation for the x-axis in the case of iterations and use the integer prior over the indices to have a better visualization (as in matplotlib `MaxNLocator(integer=True)`)
- The default scale for a plot function of iterations should be linear

(Not linked to this PR, but with benchopt the kind of ZigZag pattern that can be observed in the current result plot is not rare and a little bit puzzling)

tomMoral:
Actually, I am not sure we want to add a switch.
I would be more in favor of a solution to have a selector that allows to put any metric as `x` and any as `y`? (keeping the default to `time` and first objective value).
This would make the UI very explicit that one is choosing the axis value.
And it would also solve extra use cases, where people want to look for pareto front with multiple objectives.

Also, this would allow to separate `iterations/tolerance`, which I think is a pain to handle right now (and probably why it is called `stop_value` in the PR).

WDYT?

jeandut:
I would like to reiterate that the completion of this PR would be extremely valuable for a lot of us in the community.
I side with @tomMoral on the selector for `x` instead of a toggle. It would be very easy to add more resampling options in the long-term.

I also remain adamant that at least on the display html, names should reflect what the user is interested in and not variable names from the codebase i.e. iterations/tolerance and not StoppingCriterion or stop_val.

It would be really great as well if we can have integer ticks if x == iterations.

Cheers.

tomMoral:
As discussed with @mathurinm IRL, I propose the following solution to move on with this PR:

- We add a drop-down menu with `['iteration', 'tolerance']` on the page.
- In the `solver_data[solver]['scatter']` dictionary, we add a value `'strategy'` in `iteration, tolerance`.
- Depending on the drop-down menu, we display only the solver with the strategy that matches the one selected in the drop-down menu.

I would say this looks like a good first step, without adding too much complexity? WDYT @Badr-MOUFAD ?

Badr-MOUFAD:
Yes, it sounds good to me.

Badr-MOUFAD:
@tomMoral, I have a little concern about this point

> In the solver_data[solver]['scatter'] dictionary, we add a value 'strategy' in iteration, tolerance.

In fact, we don't store the strategy in the data frame of results and hence we cannot determine the stopping strategy at the plotting stage: [`generate_html.py`](https://github.com/benchopt/benchopt/blob/main/benchopt/plotting/generate_html.py).

One workaround is to add the stopping strategy attribute in ``meta`` when running the benchmark. That way the ``stopping_startegy`` would be available in the final data frame

https://github.com/benchopt/benchopt/blob/1487639e156a9acec809cb3a86fd2d2e7960115f/benchopt/runner.py#L187-L192

WDYT @tomMoral? Do you see another workaround?

tomMoral:
> WDYT @tomMoral? Do you see another workaround?

I would have suggested the same so perfect :) As the dataframe is stored as a parquet (thus compressed), the impact on the output file's size should be minim so works for me.

Badr-MOUFAD:
With the latter remarks being taken into account here is a [preview of the features](https://badr-moufad.github.io/share-benchmarks/stop_val_plot/validate-plotting-stop-val_benchopt_run_2023-03-02_17h04m58.html).

The xaxis type drop menu acts also as a filter: only the solver with a stopping strategy matching xaxis type will be shown.

@tomMoral, @jeandut I am open to suggestions.

benchopt/plotting/generate_html.py
```diff
@@ -181,21 +181,26 @@ def shape_solvers_for_html(df, objective_column):
         df_filtered = df_filtered.dropna(subset=[objective_column])
 
         q1, q9 = compute_quantiles(df_filtered)
+        groupby_stop_val_median = df_filtered.groupby('stop_val').median()
         color, marker = get_solver_style(solver)
+
+        stopping_strategy = df_filtered['stopping_strategy'].unique()
```
Badr-MOUFAD, line 187: Shall we consider a sanity check here: if run, a solver can have only one stopping strategy?

tomMoral: You can always add the sanity check, it is better to fail fast than get weird behaviors :)

benchopt/plotting/html/templates/result.mako.html
```diff
@@ -92,6 +92,16 @@ <h1 class="ml11">
                                 <span class="slider round"></span>
                             </label>
                         </div>
+                        <div id="xaxis-type-form-group" class="mt-8 px-8">
+                            <label for="change_scaling" class="block text-sm font-medium text-white">X axis type</label>
+                            <div class="mt-1 rounded-md shadow-sm">
+                                <select id="change_xaxis_type" class="focus:ring-blue-500 focus:border-blue-500 block w-full h-8 px-4 sm:text-sm border-gray-300 rounded-md" onchange="setState({xaxis_type: this.value})">
+                                    <option value="time">Time</option>
+                                    <option value="iteration">Iteration</option>
+                                    <option value="tolerance">Tolerance</option>
```
tomMoral, line 101: can this be parametrized to avoid putting `iteration` or `tolerance` if it does not exists in the data?
Actually, all these values should come from the `['time'] + list(df['stopping_strategy'].unique())`

benchopt/runner.py
```diff
@@ -189,6 +189,7 @@ def run_one_solver(benchmark, dataset, objective, solver, n_repetitions,
             solver_name=str(solver),
             data_name=str(dataset),
             idx_rep=rep,
+            stopping_strategy=solver._solver_strategy
```
tomMoral, line 192: for `stopping_strategy = 'callback'`, we should replace it with `'iteration'`.

benchopt/plotting/html/static/result.js
```diff
@@ -362,7 +386,7 @@ const getScatterChartLayout = () => {
     },
     xaxis: {
       type: getScale().xaxis,
-      title: 'Time [sec]',
+      title: withStopVal ? 'Stop value': 'Time [sec]',
```
tomMoral, line 389: can you change this to the value of `x_axis_type`?

benchopt/plotting/html/static/result.js
```diff
@@ -167,7 +190,7 @@ const getScatterCurves = () => {
         legendgroup: solver,
         hovertemplate: '(%{x:.1e},%{y:.1e}) <extra></extra>',
         visible: isVisible(solver) ? true : 'legendonly',
-        x: data(solver).scatter.q9,
+        x: data(solver).scatter[scatterXaxisProperties[1]],
         y: useTransformer(data(solver).scatter.y, 'y', data().transformers),
       });
```
tomMoral, line 195: Maybe skip this all together if `xaxisType` is not `time`?

benchopt/plotting/html/static/result.js
```diff
@@ -136,12 +148,23 @@ const getScatterCurves = () => {
...
       visible: isVisible(solver) ? true : 'legendonly',
-      x: data(solver).scatter.x,
+      x: data(solver).scatter[ScatterXaxisProperty],
       y: useTransformer(data(solver).scatter.y, 'y', data().transformers),
     });
 
     if (state().with_quantiles) {
       // Add shaded area for each solver, with proper style and visibility.
+
+      solverStoppingStrategy = data(solver)['stopping_strategy'];
```
tomMoral, line 158: Already defined in line 127 no?

benchopt/plotting/html/static/result.js
```diff
@@ -136,12 +148,23 @@ const getScatterCurves = () => {
...
+
+      // plot only solvers that were stopped using xaxis type
+      // plot all solver if xaxis type is `time`
+      if(xaxisType !== "time" && solverStoppingStrategy !== xaxisType) {
+        return
+      }
```
tomMoral, line 164: already done too no? or am I missing a statement?

tomMoral (commented):
LGTM! a few comments but I really like the result, it looks very smooth! thanks a lot @Badr-MOUFAD 

the two things to fix:
- [ ] `callback` strategy should be mapped to `iteration`
- [ ] the drop down menu should contain only values present in the dataframe

jeandut:
> With the latter remarks being taken into account here is a [preview of the feature](https://badr-moufad.github.io/share-benchmarks/stop_val_plot/validate-plotting-stop-val_benchopt_run_2023-03-02_17h04m58.html).
> 
> The xaxis type drop menu acts also as a filter: only the solver with a stopping strategy matching xaxis type will be shown.
> 
> @tomMoral, @jeandut I am open to suggestions.

Very cool work it looks very good !
A few disorganised comments on the preview:
- I would remove the word type after X-axis.
- Tolerance axis does not seem to work in this example
- would love to have integer tick labels when switching to iterations but this is a detail
Overall it's very nice !

benchopt/plotting/html/static/result.js
Badr-MOUFAD: Thanks for pointing this out, quantiles lose sense as ``stop_val`` is deterministic hence won't change throughout the reps.

benchopt/plotting/html/static/result.js
Badr-MOUFAD: You didn't, thanks for the catch!

benchopt/plotting/html/templates/result.mako.html
Badr-MOUFAD: I opted for dynamically creating the options depending on the selected dataset and objective columns. In fact, some solvers might be skipped in some cases.

Any thoughts?

Badr-MOUFAD:
Thanks, @jeandut for the feedback! Here is the [updated preview of the feature](https://badr-moufad.github.io/share-benchmarks/stop_val_plot/validate-plotting-stop-val_benchopt_run_2023-03-03_15h06m33.html).

tomMoral:
Could not check on my computer right now but any idea why the x-axis is not visible on mobile?


I will review it tomorrow but the changes look good and good idea to create the drop down menu dynamically, it is better to only have non empty option 😁

Badr-MOUFAD:
> Could not check on my computer right now but any idea why the x-axis is not visible on mobile?

I have just become aware that the responsiveness (mobile support) was implemented by duplicating the sidebar's content and styling it separately.

https://github.com/benchopt/benchopt/blob/d520e74b09eed7b3bf6c87bc2a4ef90e5d36655b/benchopt/plotting/html/templates/result.mako.html#L22-L25

https://github.com/benchopt/benchopt/blob/d520e74b09eed7b3bf6c87bc2a4ef90e5d36655b/benchopt/plotting/html/templates/result.mako.html#L97-L99

Hence to fix https://github.com/benchopt/benchopt/pull/479#issuecomment-1454283066, we should make the same changes we did for the sidebar of the big screen to the one on the mobile screen. 

Responsiveness is usually implemented [through breakpoints](https://tailwindcss.com/docs/responsive-design) that specify the style to be applied according to the screen size. 
A follow-up TODO of this PR is to remove the content duplication and implement properly the responsiveness.

jeandut:
> Thanks, @jeandut for the feedback! Here is the [updated preview of the feature](https://badr-moufad.github.io/share-benchmarks/stop_val_plot/validate-plotting-stop-val_benchopt_run_2023-03-04_13h58m29.html).

Awesome !!!

jeandut:
Although there might be a bug see values in iteration axis when converting to log-log ?:
Capture d’écran 2023-03-04 à 16 12 52

benchopt/plotting/generate_html.py
```diff
@@ -181,21 +181,37 @@ def shape_solvers_for_html(df, objective_column):
...
+
+        if len(stopping_strategy) != 1:
+            found_stopping_strategies = ', '.join(
+                f"`{item}`" for item in stopping_strategy)
```
tomMoral, line 191:
```suggestion
                f"`{item}`" for item in stopping_strategy
            )
```

benchopt/plotting/html/static/result.js
```diff
@@ -362,8 +390,8 @@ const getScatterChartLayout = () => {
     },
     xaxis: {
       type: getScale().xaxis,
-      title: 'Time [sec]',
-      tickformat: '.1e',
+      title: xaxisType === "time" ? "Time [sec]": xaxisType,
+      tickformat: xaxisType === "time" ? '.1e': '',
```
tomMoral, line 394: I think the format should still be `.1e` for `tolerance` so only change it to integer for `iterations`?

benchopt/runner.py
```diff
@@ -189,6 +196,7 @@ def run_one_solver(benchmark, dataset, objective, solver, n_repetitions,
             solver_name=str(solver),
             data_name=str(dataset),
             idx_rep=rep,
+            stopping_strategy=stopping_strategy
```
tomMoral, line 199: To make it prettier?
```suggestion
            stopping_strategy=stopping_strategy.capitalize()
```

tomMoral (approved):
Ok, it looks good! Just a small fix for the xticks and 2 nitpicks and we are good to go!
Thx @Badr-MOUFAD !

tomMoral:
> > Could not check on my computer right now but any idea why the x-axis is not visible on mobile?
> 
> I have just become aware that the responsiveness (mobile support) was implemented by duplicating the sidebar's content and styling it separately.
> 
> https://github.com/benchopt/benchopt/blob/d520e74b09eed7b3bf6c87bc2a4ef90e5d36655b/benchopt/plotting/html/templates/result.mako.html#L22-L25
> 
> https://github.com/benchopt/benchopt/blob/d520e74b09eed7b3bf6c87bc2a4ef90e5d36655b/benchopt/plotting/html/templates/result.mako.html#L97-L99
> 
> Hence to fix [#479 (comment)](https://github.com/benchopt/benchopt/pull/479#issuecomment-1454283066), we should make the same changes we did for the sidebar of the big screen to the one on the mobile screen.
> 
> Responsiveness is usually implemented [through breakpoints](https://tailwindcss.com/docs/responsive-design) that specify the style to be applied according to the screen size. A follow-up TODO of this PR is to remove the content duplication and implement properly the responsiveness.
> 
> P.S. I updated the feature preview

Thanks a lot for the explanation, indeed it would be much better to implement it with the right synthax.

Badr-MOUFAD:
> Although there might be a bug see values in iteration axis when converting to log-log ?: 1-2-5-10-2-5-100-2-5-1000 etc.

Looking at [Plotly log scale doc](https://plotly.com/python/log-plot/), I don't think it is a bug.
Looks fine to me. Do you think it might be misleading @jeandut?

benchopt/plotting/html/static/result.js
Badr-MOUFAD: Didn't pay attention to that. Thanks for the catch!

jeandut:
> > Although there might be a bug see values in iteration axis when converting to log-log ?: 1-2-5-10-2-5-100-2-5-1000 etc.
> 
> Looking at [Plotly log scale doc](https://plotly.com/python/log-plot/), I don't think it is a bug. Looks fine to me. Do you think it might be misleading @jeandut?

You are totally right. The tick values do make sense somehow on log-scale. Looks like it's ready to be merged ! Great work !

tomMoral:
merged, thanks a lot for making this feature @Badr-MOUFAD !

PR #483, opened 2022-11-10
ENH specify channel for conda reqs

tomMoral:
Introduce a new synthax to add a channel for conda requirements with `channel:package`.

closes #477

### DIFF
--- a/benchopt/utils/conda_env_cmd.py
+++ b/benchopt/utils/conda_env_cmd.py
@@ -188,7 +188,7 @@ def get_cmd_from_requirements(packages):
         ))
         packages = ' '.join(pkg.split(':')[-1] for pkg in conda_packages)
         cmd.append(
-            f"{CONDA_CMD} install jsdg --update-all -y {channels}  {packages}"
+            f"{CONDA_CMD} install --update-all -y {channels} {packages}"
         )
 
     if pip_packages:
--- a/benchopt/utils/conda_env_cmd.py
+++ b/benchopt/utils/conda_env_cmd.py
@@ -172,30 +172,44 @@ def delete_conda_env(env_name):
                capture_stdout=True)
 
 
+def get_cmd_from_requirements(packages):
+    """Process the packages from requirements and create the install cmd.
+
+    This detect the packages that needs to be installed with pip and also
+    the additional channels for conda packages.
+    """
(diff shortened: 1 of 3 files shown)


benchopt/utils/conda_env_cmd.py
```diff
@@ -172,30 +172,44 @@ def delete_conda_env(env_name):
...
     conda_packages = [pkg for pkg in packages if not pkg.startswith('pip:')]
 
-    error_msg = ("Failed to conda install packages "
-                 f"{packages if len(packages) > 1 else packages[0]}\n"
-                 "Error:{output}")
     cmd = []
     if conda_packages:
-        packages = ' '.join(conda_packages)
-        cmd.append(f"{CONDA_CMD} install --update-all -y {packages}")
+        channels = ' '.join(set(
```
mathurinm, line 186: Does this use common channels for all packages? And do we want to use some channels only for some packages (I'd say yes) ? So we'd need several "conda install" lines, each with its channel

tomMoral: We want to have all installed packages in a same env, so usually what is done is to let the deps solver figure out which are compatible together. so I would not mess with multiple install as this tends to make the install unstable.
Moreover, this would make as many dependency resolution as you have lines so it would be slow.

tomMoral: This is quite rare but it could happen. In this case, the most recent one will be installed.
But I think this is very marginal, we should not over engineer this. Most packages are in one channel (hopefully `conda-forge`) and we should only add a channel if the package is really not avail in the other channel IMO.

mathurinm: then fine by me!

benchopt/utils/conda_env_cmd.py
```diff
@@ -172,30 +172,44 @@ def delete_conda_env(env_name):
...
 
 
-def install_in_conda_env(*packages, env_name=None, force=False, quiet=False):
-    """Install the packages with conda in the given environment"""
-    if len(packages) == 0:
-        return
+def get_cmd_from_requirements(packages):
+    """Process the packages from requirements and create the install cmd.
 
+    This detect the packages that needs to be installed with pip and also
```
mathurinm, line 178:
```suggestion
    This detects the packages that need to be installed with pip and also
```

mathurinm (approved):

PR #485, opened 2022-11-11
FIX detection of csv files when plotting with option --all

MatDag:
When running `benchopt plot --all`, the variable `result_filename` is a list and so hasn't a `suffix` attribute. This PR handle this case.

### DIFF
--- a/benchopt/benchmark.py
+++ b/benchopt/benchmark.py
@@ -220,12 +220,20 @@ class Benchmark:
             if filename == 'all':
                 result_filename = all_result_files
 
-        if result_filename.suffix == ".csv":
-            print(colorify(
-                "WARNING: CSV files are deprecated."
-                "Please use Parquet files instead.",
-                YELLOW
-            ))
+        if isinstance(result_filename, list):
+            if any(fname.suffix == ".csv" for fname in result_filename):
+                print(colorify(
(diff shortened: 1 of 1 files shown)


benchopt/benchmark.py
```diff
@@ -220,12 +220,20 @@ def get_result_file(self, filename=None):
...
+                    "Please use Parquet files instead.",
+                    YELLOW
+                ))
+        else:
+            if result_filename.suffix == ".csv":
+                print(colorify(
+                    "WARNING: CSV files are deprecated."
+                    "Please use Parquet files instead.",
+                    YELLOW
+                ))
```
Badr-MOUFAD, line 236: That's a nice catch indeed @MatDag!

For conciseness I suggest this reformulation:
```python
if isinstance(result_filename, list):
    is_csv_file = any(fname.suffix == ".csv"
                      for fname in result_filename)
else:
    is_csv_file = result_filename.suffix == ".csv"

if is_csv_file:
    print(colorify(
        "WARNING: CSV files are deprecated."
        "Please use Parquet files instead.",
        YELLOW
    ))

mathurinm (approved):

mathurinm:
@tomMoral merge?

tomMoral (approved):

PR #486, opened 2022-11-12
FIX test command relative import

tomMoral:
As reported by @mathurinm , [this change](https://github.com/benchopt/benchopt/commit/4733a8f5a9f278353bfd4c98ba2de2a8d253ead6#diff-b02f52b735585c6cb5cf0d64f4a4031b464e42b9ddd51519faaa0edbba5310c4R9) broke the tests on all benchmark repos.

This PR fixes it.

### DIFF
--- a/benchopt/tests/__init__.py
+++ b/benchopt/tests/__init__.py
@@ -6,9 +6,8 @@ from joblib.executor import get_memmapping_executor
 from benchopt.benchmark import Benchmark
 from benchopt.utils.stream_redirection import SuppressStd
 
-from .fixtures import TEST_BENCHMARK_DIR
-
 # Default benchmark
+TEST_BENCHMARK_DIR = Path(__file__).parent / 'test_benchmarks'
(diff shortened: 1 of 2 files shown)


tomMoral:
I was actually surprise that we do not test the `benchopt test` command... We should fix this to avoid such situation :sweat_smile:

mathurinm:
I couldn't understand why `pytest` was missing: it's in the `[options.extras_require]` section of the `setup.cfg` ?

mathurinm (approved):

tomMoral:
I did not see where purest was missing?

The issue fixed here was just the one of relative import that should be avoided to make `benchopt test` work. If there is an issue with missing pytest, can you print me to it?

mathurinm:
It's what I saw here for example : https://github.com/benchopt/benchmark_logreg_l1/actions/runs/3450096350/jobs/5758467949#step:7:3

tomMoral:
Yes this is normal. We do not have to install `pytest` in the main env, it is installed in the env where the test are run, `bench_test_env` here.

PR #487, opened 2022-11-14
change colormap + clean plotting functions

### LINKED ISSUE #467: PLT loop over dark colors first, then light ones
Currently if there are only two solvers, they are dark blue and light blue. Looping over dark colors first will maximize diversity of colors when there are less than 10 solvers

Currently our cmap is a callable, but we can transform it into a list with `ListedColormap()` from matplotlib.colors and then use `colors = colors[::2] + colors[1::2]`

AmelieVernay:
fix #467

### DIFF
--- a/benchopt/plotting/plot_objective_curve.py
+++ b/benchopt/plotting/plot_objective_curve.py
@@ -68,7 +68,7 @@ def plot_objective_curve(df, obj_col='objective_value',
 
     for i, solver_name in enumerate(solver_names):
         df_ = df[df['solver_name'] == solver_name]
-        curve = df_.groupby('stop_val').median()
+        curve = df_.groupby('stop_val').median(numeric_only=True)
 
         q1 = df_.groupby('stop_val')['time'].quantile(.1)
         q9 = df_.groupby('stop_val')['time'].quantile(.9)
--- a/benchopt/plotting/plot_objective_curve.py
+++ b/benchopt/plotting/plot_objective_curve.py
@@ -149,7 +149,8 @@ def get_solver_style(solver, plotly=True):
     global html_solver_styles
     if (len(html_solver_styles) != 0 and
             not plotly and
-            'rgb' in html_solver_styles[next(iter(html_solver_styles))]['color']):
+            'rgb' in html_solver_styles[
+                next(iter(html_solver_styles))]['color']):
         html_solver_styles = {}
 
     if solver in html_solver_styles:
--- a/benchopt/plotting/plot_objective_curve.py
+++ b/benchopt/plotting/plot_objective_curve.py
@@ -145,52 +145,16 @@ def compute_quantiles(df_filtered):
     return q1, q9
 
 
-#def get_solver_color(solver, plotly=True):
-#    print(f'{plotly=}')
-#    if solver in html_solver_styles and 'color' in html_solver_styles[solver]:
-#        return html_solver_styles[solver]['color']
-#
-#    idx = len(html_solver_styles)
-#    color = colors[idx % len(colors)]
-#
-#    if plotly:
-#        color = tuple(255*x if i != 3 else x for i, x in enumerate(color))
-#        color = f'rgba{color}'
-#
-#    if solver in html_solver_styles:
-#        html_solver_styles[solver]['color'] = color
-#    else:
-#        html_solver_styles[solver] = {
-#            'color': color
-#        }
-#
-#    return color
-
-
-#def get_solver_marker(solver, plotly=True):
-#    if solver in html_solver_styles and 'marker' in html_solver_styles[solver]:
-#        return html_solver_styles[solver]['marker']
-#
-#    if plotly:
-#        markers = {i: i for i, v in enumerate(plt.Line2D.markers)}
-#    else:
-#        markers = {i: v for i, v in enumerate(plt.Line2D.markers)}
-#
-#    idx = len(html_solver_styles)
-#    marker = markers[idx % len(markers)]
-#
-#    if solver in html_solver_styles:
-#        html_solver_styles[solver]['marker'] = marker
-#    else:
-#        html_solver_styles[solver] = {
-#            'marker': marker
-#        }
-#
-#    return marker
-
 def get_solver_style(solver, plotly=True):
+    global html_solver_styles
+    if (len(html_solver_styles) != 0 and
+        not plotly and
+        'rgb' in html_solver_styles[next(iter(html_solver_styles))]['color']):
+        html_solver_styles = {}
+    
     if solver in html_solver_styles:
-        return html_solver_styles[solver]['color'], html_solver_styles[solver]['marker']
+        return html_solver_styles[solver]['color'], \
+                html_solver_styles[solver]['marker']
 
     idx = len(html_solver_styles)
     color = colors[idx % len(colors)]
@@ -200,7 +164,7 @@ def get_solver_style(solver, plotly=True):
         color = tuple(255*x if i != 3 else x for i, x in enumerate(color))
         color = f'rgba{color}'
         marker = list(markers.values()).index(marker)
-    
+
     if solver in html_solver_styles:
         html_solver_styles[solver]['color'] = color
         html_solver_styles[solver]['marker'] = marker
@@ -210,5 +174,3 @@ def get_solver_style(solver, plotly=True):
             'marker': marker
         }
     return color, marker
-
-
--- a/benchopt/plotting/plot_objective_curve.py
+++ b/benchopt/plotting/plot_objective_curve.py
@@ -5,6 +5,9 @@ from .helpers_compat import add_h_line
 from .helpers_compat import fill_between_x
 
 CMAP = plt.get_cmap('tab20')
+colors = [CMAP(i) for i in range(CMAP.N)]
+colors = colors[::2] + colors[1::2]
+
 
 html_solver_styles = {}
 
@@ -13,7 +16,7 @@ def _remove_prefix(text, prefix):
     return text[len(prefix):] if text.startswith(prefix) else text
 
 
-def plot_objective_curve(df, obj_col='objective_value', plotly=False,
+def plot_objective_curve(df, obj_col='objective_value',
                          suboptimality=False, relative=False):
     """Plot objective curve for a given benchmark and dataset.
 
@@ -25,8 +28,6 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
         The benchmark results.
     obj_col : str
         Column to select in the DataFrame for the plot.
-    plotly : bool
-        If set to True, output a plotly figure for HTML display.
     suboptimality : bool
         If set to True, remove the optimal objective value F(x^*). Here the
         value of F(x^*) is taken as the smallest value reached across all
@@ -37,13 +38,10 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
 
     Returns
     -------
-    fig : matplotlib.Figure or pyplot.Figure
+    fig : matplotlib.Figure
         The rendered figure, used to create HTML reports.
     """
-    if plotly:
-        markers = {i: i for i, v in enumerate(plt.Line2D.markers)}
-    else:
-        markers = {i: v for i, v in enumerate(plt.Line2D.markers)}
+    markers = {i: v for i, v in enumerate(plt.Line2D.markers)}
 
     df = df.copy()
     solver_names = df['solver_name'].unique()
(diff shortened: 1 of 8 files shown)


benchopt/plotting/plot_objective_curve.py
```diff
@@ -84,42 +76,28 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
 
         q1 = df_.groupby('stop_val')['time'].quantile(.1)
         q9 = df_.groupby('stop_val')['time'].quantile(.9)
-
+        
         fill_between_x(
-            fig, curve['time'], q1, q9, curve[obj_col], color=CMAP(i % CMAP.N),
-            marker=markers[i % len(markers)], label=solver_name, plotly=plotly
+            fig, curve['time'], q1, q9, curve[obj_col], color=colors[i % CMAP.N],
```
mathurinm, line 81: could you use `get_solver_color` here in order to make all the color dispatching mechanism happen in this function ? Otherwise we have to maintain it here, and in `get_solver_function`

tomMoral: agreed

benchopt/plotting/plot_objective_curve.py
```diff
@@ -66,16 +64,10 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
         max_f_0 = df[df['stop_val'] == 1][obj_col].max()
         df.loc[:, obj_col] /= max_f_0
 
-    fig = get_figure(plotly)
+    fig = plt.figure()
```
tomMoral, line 67: also remove all import from helper compat, that are only here to make it work with plotly.

benchopt/plotting/plot_objective_curve.py
```diff
@@ -84,42 +76,28 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
 
         q1 = df_.groupby('stop_val')['time'].quantile(.1)
         q9 = df_.groupby('stop_val')['time'].quantile(.9)
-
+        
         fill_between_x(
-            fig, curve['time'], q1, q9, curve[obj_col], color=CMAP(i % CMAP.N),
-            marker=markers[i % len(markers)], label=solver_name, plotly=plotly
+            fig, curve['time'], q1, q9, curve[obj_col], color=colors[i % CMAP.N],
+            marker=markers[i % len(markers)], label=solver_name, plotly=False
```
tomMoral, line 82: remove argument `plotly` from `fill_between_x` and possibly simplify. if this is two lines, consider removing the finc as I believe it is not reused anywhere?

benchopt/plotting/plot_objective_curve.py
```diff
@@ -84,42 +76,28 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
...
         )
 
     if suboptimality and not relative:
         add_h_line(fig, eps, [df['time'].min(), df['time'].max()],
```
tomMoral, line 86: just use matplotlib hlines.

benchopt/plotting/plot_objective_curve.py
tomMoral: also same for markers?

tomMoral (commented):
This is great clean up! thx @AmelieVernay 

I think if we can reuse `get_color_solver` and marker, that would be great.
also remove the plotly/matplotlib compat layer if possible

AmelieVernay:
Sorry @tomMoral, @mathurinm  I might have accidentally resolve the conversation about `get_solver_color()` and `get_solver_marker()` so I catch up here.

The thing with `get_solver_color()`is that it outputs a non normalized rgba tuple prepended with 'rgba', all of this **as a string** (e.g. `'rgba(255.0, 127.3, 14.0, 1.0)'`, and matplotlib needs a normalized rgba tuple.

The simplest way I found to make it work with `get_solver_color()` is 
```{py}
str_col = get_solver_color(solver_name)[4:]
col = tuple(map(float, str_col.replace('(','').replace(')','').split(',')))
norm_col = tuple(x/255. if i != 3 else x for i, x in enumerate(col))
fill_between_x(
    fig, curve['time'], q1, q9, curve[obj_col], color=norm_col,
    marker=markers[i % len(markers)], label=solver_name, plotly=False
)
```
which is rather verbose... do you have a better idea ?

By the way, maybe it is better to only use regexp instead of indexing in case the 'rgba' was changed into 'rgb' for some reason. IDK

As for `get_solver_marker()`, I would need another trick because it outputs integers for plotly markers and matplotlib takes symbols as str, e.g. `'o'` (it can also take integers from 1 to 11 [but these do not encode the same symbols](https://matplotlib.org/stable/api/markers_api.html)).

I thought `get_solver_color()` and `get_solver_marker()` were written for plotly purpose.

WDYT ?

benchopt/plotting/plot_objective_curve.py
mathurinm: In fact, as Amélie pointed out I find it weird to use `get_solver_color` here, as we would just undo the work that is done by `get_solver_color` (which takes colors[i] and maps it to a string). I'd be in favor of leaving the call to `colors`

benchopt/plotting/plot_objective_curve.py
AmelieVernay: No it is not, neither is `add_h_line()`so I will remove these two functions here and remove the `if not plotly` part in their definition.

tomMoral:
I would say do not reverse engineer `get_solver_plot` but add a `plotly` kwargs?

benchopt/plotting/plot_objective_curve.py
```diff
@@ -85,41 +73,29 @@ def plot_objective_curve(df, obj_col='objective_value', plotly=False,
         q1 = df_.groupby('stop_val')['time'].quantile(.1)
         q9 = df_.groupby('stop_val')['time'].quantile(.9)
 
-        fill_between_x(
-            fig, curve['time'], q1, q9, curve[obj_col], color=CMAP(i % CMAP.N),
-            marker=markers[i % len(markers)], label=solver_name, plotly=plotly
-        )
+        col = get_solver_color(solver_name, plotly=False)
+        plt.loglog(curve['time'], curve[obj_col], color=col,
+                   marker=markers[i % len(markers)],
```
mathurinm, line 78: I suggest to harmonize here: use `get_solver_marker`

benchopt/plotting/plot_objective_curve.py
```diff
@@ -37,7 +37,7 @@ def plot_objective_curve(df, obj_col='objective_value',
     fig : matplotlib.Figure
         The rendered figure, used to create HTML reports.
     """
-    markers = {i: v for i, v in enumerate(plt.Line2D.markers)}
+#    markers = {i: v for i, v in enumerate(plt.Line2D.markers)}
```
tomMoral, line 40: Avoid having trailing comments in a PR.

benchopt/plotting/plot_objective_curve.py
```diff
@@ -167,14 +167,17 @@ def get_solver_color(solver, plotly=True):
...
-def get_solver_marker(solver):
+def get_solver_marker(solver, plotly=True):
     if solver in html_solver_styles and 'marker' in html_solver_styles[solver]:
         return html_solver_styles[solver]['marker']
 
-    markers = {i: i for i, v in enumerate(plt.Line2D.markers)}
+    if plotly:
+        markers = {i: i for i, v in enumerate(plt.Line2D.markers)}
+    else:
+        markers = {i: v for i, v in enumerate(plt.Line2D.markers)}
```
tomMoral, line 177: Something like this seems a bit easier to follow, with `MARKERS` a constant fo rthe module.

```suggestion
    idx = len(html_solver_styles)
    if plotly:
        marker = idx % len(MARKERS)
    else:
        marker = MARKERS[idx % MARKERS]
```

benchopt/plotting/plot_objective_curve.py
```diff
@@ -1,10 +1,9 @@
 import matplotlib.pyplot as plt
 
-from .helpers_compat import get_figure
-from .helpers_compat import add_h_line
-from .helpers_compat import fill_between_x
-
 CMAP = plt.get_cmap('tab20')
+colors = [CMAP(i) for i in range(CMAP.N)]
+colors = colors[::2] + colors[1::2]
```
tomMoral, line 5: These are constants, so they should be in upper case.

benchopt/plotting/plot_objective_curve.py
```diff
@@ -195,14 +167,17 @@ def get_solver_color(solver):
     return color
 
 
-def get_solver_marker(solver):
+def get_solver_marker(solver, plotly=True):
     if solver in html_solver_styles and 'marker' in html_solver_styles[solver]:
         return html_solver_styles[solver]['marker']
```
tomMoral, line 172: There is something wrong with the logic... here, the marker `0` will never be used, and the color `1` neither ^^.
This is due to the original design of this function but could we merge the two into a `get_solver_style` that generate both color and marker at the same time? and then select wathever we need.

tomMoral (commented):
A few comments to make this more robust

AmelieVernay:
I converted to draft so I can test with the CI (cannot reproduce the bug locally). I will convert back when succeeding

AmelieVernay:
For the tests to pass, I had to add a condition in `get_solver_style()` to re-initialized `html_solver_style` if it was already filled with ploty-like values but called with `plotly=False` (there might be a better way).
Perhaps it was to avoid having to do this that `get_solver_color()` was not used in the first place with `--no-html`.

AmelieVernay:
When we run benchopt with `--no-html`, the barchart and the linechart do not use the same color palette (unlike in the html version). Could we change this ?
screen_benchopt_nohtml

tomMoral:
Yes we can harmonize the colors.

For the `get_solver_color`, maybe a better way would be to store an index for each `solver` and then compute the `color` or  `marker` only on the index value? that way, the stored info is indep of `plotly=True/False`?

AmelieVernay:
You mean so that e.g. solver `celer` with plotly and `celer` without plotly would have two different indices (and thus two different color values?

tomMoral:
No I mean that they would have the same indices.
when you call `get_color_solver(solver)`, first you generate an index and store it. then you generate the style only based no this index: compute a color and if plotly you create a string:

```python
    idx = solvers_idx.get(solver, len(solvers_idx)
    solvers_idx[solver] = idx

    c = COLORS[idx % len(COLORS)]
    if plotly:
        c = ... # generate str
    return c
```

AmelieVernay:
Ok so the color/marker would be computed each time instead of being stored once and for all, right?

benchopt/plotting/plot_objective_curve.py
```diff
@@ -1,19 +1,19 @@
 import matplotlib.pyplot as plt
 
-from .helpers_compat import get_figure
-from .helpers_compat import add_h_line
-from .helpers_compat import fill_between_x
-
 CMAP = plt.get_cmap('tab20')
+colors = [CMAP(i) for i in range(CMAP.N)]
+colors = colors[::2] + colors[1::2]
+markers = {i: v for i, v in enumerate(plt.Line2D.markers)}
```
tomMoral, line 6:
```suggestion
COLORS = [CMAP(i) for i in range(CMAP.N)]
COLORS = COLORS[::2] + COLORS[1::2]
MARKERS = {i: v for i, v in enumerate(plt.Line2D.markers)}
```

benchopt/plotting/plot_objective_curve.py
```diff
@@ -176,39 +145,16 @@ def compute_quantiles(df_filtered):
...
+    solvers_idx[solver] = idx
 
-    if solver in html_solver_styles:
-        html_solver_styles[solver]['color'] = color
-    else:
-        html_solver_styles[solver] = {
-            'color': color
-        }
+    color = colors[idx % len(colors)]
+    marker = markers[idx % len(markers)]
```
tomMoral, line 153:
```suggestion
    color = COLORS[idx % len(COLORS)]
    marker = MARKERS[idx % len(MARKERS)]
```

benchopt/plotting/plot_objective_curve.py
```diff
@@ -176,39 +145,16 @@ def compute_quantiles(df_filtered):
...
-    if solver in html_solver_styles:
-        html_solver_styles[solver]['marker'] = marker
-    else:
-        html_solver_styles[solver] = {
-            'marker': marker
-        }
+    if plotly:
+        color = tuple(255*x if i != 3 else x for i, x in enumerate(color))
+        color = f'rgba{color}'
+        marker = list(markers.values()).index(marker)
```
tomMoral, line 158: I think we can imply do:
```suggestion
        marker = idx
```

AmelieVernay:
I added a `numeric_only=True` in pandas median to avoid numerous FutureWarning when using `--no-html`.

tomMoral:
> When we run benchopt with `--no-html`, the barchart and the linechart do not use the same color palette (unlike in the html version). Could we change this ? screen_benchopt_nohtml

I think we did not do this one? I merge this as I want to do a release this afternoon, but feel free to do this in a follow up PR.

PR #489, opened 2022-11-15
API deprecate `Objective.to_dict` in favor of `Objective.get_objective`

### LINKED ISSUE #460: API deprecate Objective.to_dict in favor of Objective.get_objective
It will be more coherent with `Dataset.get_data`/`Objective.set_data` and `BaseSolver.set_objective`

mathurinm:
closes #460

### DIFF
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -21,7 +21,7 @@ class BaseSolver(ParametrizedNameMixin, DependenciesMixin, ABC):
 
     - ``set_objective(self, **objective_parameters)``: prepares the solver to
       be called on a given problem. ``**objective_parameters`` is the output of
-      the method ``set_objective`` from the benchmark objective. In particular,
+      the method ``get_objective`` from the benchmark objective. In particular,
       this method should dumps the parameter to compute the objective function
       in a file for command line solvers to reduce the impact of dumping the
       data to the disk in the benchmark.
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -20,8 +20,8 @@ class BaseSolver(ParametrizedNameMixin, DependenciesMixin, ABC):
     Solvers that derive from this class should implement three methods:
 
     - ``set_objective(self, **objective_parameters)``: prepares the solver to
-      be called on a given problem. ``**objective_parameters`` are the output
-      of the method ``to_dict`` from the benchmark objective. In particular,
+      be called on a given problem. ``**objective_parameters`` is the output of
+      the method ``set_objective`` from the benchmark objective. In particular,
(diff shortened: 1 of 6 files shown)


benchopt/base.py
```diff
@@ -20,8 +20,8 @@ class BaseSolver(ParametrizedNameMixin, DependenciesMixin, ABC):
     Solvers that derive from this class should implement three methods:
 
     - ``set_objective(self, **objective_parameters)``: prepares the solver to
-      be called on a given problem. ``**objective_parameters`` are the output
-      of the method ``to_dict`` from the benchmark objective. In particular,
+      be called on a given problem. ``**objective_parameters`` is the output of
+      the method ``set_objective`` from the benchmark objective. In particular,
```
tomMoral, line 24:
```suggestion
      the method ``get_objective`` from the benchmark objective. In particular,
```

tomMoral (commented):
Thx @mathurinm , a few comments as for now, the deprecation part is not working.

benchopt/base.py
```diff
@@ -86,7 +86,15 @@ def _set_objective(self, objective):
...
-        objective_dict = objective.to_dict()
+        # XXX remove in version 1.4
+        if hasattr(objective, "get_objective"):
+            objective_dict = objective.get_objective()
+        else:
+            warnings.warn(
+                "The method ``Objective.to_dict`` has been deprecated in "
+                "favor of ``Objective.get_objective``. Using it will no "
+                "longer work in version 1.4", FutureWarning)
+            objective_dict = objective.to_dict()
```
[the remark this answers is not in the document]
mathurinm: I missed that fact, thanks for coming up with a better solution !

benchopt/base.py
```diff
@@ -317,7 +329,7 @@ def set_data(self, **data):
         ...
 
     @abstractmethod
```
tomMoral, line 331:
```suggestion
```

tomMoral, line 331:
```suggestion
    # to uncomment in version 1.4, once `to_dict` has been deprecated.
    # @abstractmethod
```

mathurinm:
I tested it and it works as expected. However if we update a benchmark by replacing `to_dict` with `get_objective`, and the benchmark is run with a previous version of benchopt, it will no longer work as Objective does not have a `to_dict` objective... 

Is there a way to not break things? Keep the `to_dict` in the benchmark, which calls get_objective and warns to update benchopt?

tomMoral:
This is indeed an issue... I think we should add a `min_benchopt_version` in the benchmarks, that we  bump once we break compat and if you run a benchmark with older version of benchopt, it warns that you should update

PR #492, opened 2022-11-17
ENH min benchopt version

tomMoral:
Define `min_benchopt_version` attribute in a benchmark to raise an error if the version of benchopt that is used it too old.

### DIFF
--- a/benchopt/cli/main.py
+++ b/benchopt/cli/main.py
@@ -200,6 +200,16 @@ def run(config_file=None, **kwargs):
     # Create the Benchmark object
     benchmark = Benchmark(benchmark)
 
+    if benchmark.min_version is not None:
+        from packaging.version import parse
+        from benchopt import __version__
+        if parse(__version__) < parse(benchmark.min_version):
+            raise RuntimeError(
+                "benchopt version is too old to run this benchmark. "
+                "Please update benchopt with `pip install -U benchopt` "
+                "for this benchmark."
+            )
+
     # If env_name is False, the flag `--local` has been used (default) so
     # run in the current environment.
     if env_name == 'False':
@@ -391,7 +401,6 @@ def install(
 
     # Check that the dataset/solver patterns match actual dataset
     benchmark = Benchmark(benchmark)
-    print(f"Installing '{benchmark.name}' requirements")
     benchmark.validate_dataset_patterns(dataset_names)
     benchmark.validate_solver_patterns(solver_names)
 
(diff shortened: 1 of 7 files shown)


benchopt/cli/main.py
```diff
@@ -200,6 +200,16 @@ def run(config_file=None, **kwargs):
     # Create the Benchmark object
     benchmark = Benchmark(benchmark)
 
+    if benchmark.min_version is not None:
+        from packaging.version import parse
+        from benchopt import __version__
+        if parse(__version__) < parse(benchmark.min_version):
+            raise RuntimeError(
+                "benchopt version is too old to run this benchmark. "
```
mathurinm, line 208:
```suggestion
                f"benchopt version {__version__} is too old to run this  "
                f"benchmark, version {benchmark.min_version} is required. "
```

benchopt/cli/main.py
```diff
@@ -200,6 +200,16 @@ def run(config_file=None, **kwargs):
...
+                "Please update benchopt with `pip install -U benchopt` "
```
mathurinm, line 209: This may not work if we require the dev version as in the deprecate `to_dict` issue?

benchopt/tests/test_benchmark_features.py
```diff
@@ -32,3 +39,18 @@ def test_template_solver():
     # Make sure that this error is not raised when listing all solvers from
     # the benchmark.
     DUMMY_BENCHMARK.get_solvers()
+
+
+def test_benchopt_min_version():
+    # Make sure that importing template_dataset raises an error.
```
mathurinm, line 45:
```suggestion
```

benchopt/tests/test_benchmark_features.py
```diff
@@ -32,3 +39,18 @@ def test_template_solver():
...
+    with pytest.raises(RuntimeError, match="pip install -U"):
+        run([str(FUTURE_BENCHMARK_PATH)], 'benchopt', standalone_mode=False)
+
+    with CaptureRunOutput() as out:
```
mathurinm, line 49: I don't understand why this is needed in this test, can you please explain to me?

mathurinm: Ok got it

benchopt/cli/main.py
tomMoral: Yes, not sure what to do... We could say: `please update benchopt.` mais c'est peu actionable.

PR #502, opened 2022-11-22
PLT - add responsiveness to benchmarks figures

### LINKED ISSUE #488: PLT make figures sizes in HTML report responsive
The figures in the HTML report are not responsive to the width of the window, but I think they should be.

image

Badr-MOUFAD:
In response to @jolars request in #488,

This adds responsiveness to the benchmarks figures.

closes #488 

---------------
[link to preview new features](https://badr-moufad.github.io/share-benchmarks/responsive-plot/generated_df_lasso_bench.html)

### DIFF
--- a/benchopt/plotting/html/static/hover_index.css
+++ b/benchopt/plotting/html/static/hover_index.css
@@ -277,19 +277,6 @@ input:checked+.slider:before {
 /***** Plot *****/
 /*----------------------------------------*/
 
-@media (min-width: 768px) {
-  #unique_plot {
-    flex: 1 1 0%;
-  }
-}
-
-@media (max-width: 767px) {
-  #unique_plot {
-    flex: initial;
-    flex-shrink: 0;
-  }
-}
-
 .js-plotly-plot .plotly .modebar {
   transform: translateX(-30px);
 }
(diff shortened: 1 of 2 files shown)


tomMoral:
Neat! but I would say it would be good to give a min height and width if possible, other wise it can go wrong.



Is it possible to impose a min aspect ratio? or maybe hard constants for both height and width

Badr-MOUFAD:
I used the [chrome dev tool](https://developer.chrome.com/docs/devtools/device-mode/) to test the feature on several screen sizes but I worked. 
 
As mentioned in the article [limitation section](https://developer.chrome.com/docs/devtools/device-mode/#limitations), I think it has to do more with the mobile device itself. Indeed, on mine, I don't have even the responsiveness.

</img>

I will check your suggestion on ``min-aspect-ratio`` and let you know

Badr-MOUFAD:
The ``min-aspect-ratio`` doesn't seem to fix the problem. When enabling responsiveness, the plot manages its own width and height. Besides, when handled by the plot, the responsiveness depends on the browser (it works on my browsers, but not on my phone).

Closing this PR as it doesn't address the responsiveness properly.

tomMoral:
It seems to me it was not fixing everything but still improved compare to the current status no?

Badr-MOUFAD:
Yes absolutely, but in some cases (e.g. https://github.com/benchopt/benchopt/pull/502#issuecomment-1324718739), the plot was completely squeezed and hence one wouldn't be able to visualize the benchmark results.

So we add responsiveness for some users but perhaps deprive others of visualization.

tomMoral:
Ok so let's keep it that way then. Thanks @Badr-MOUFAD

PR #503, opened 2022-11-23
PLT use same color palette as curves for barchart

AmelieVernay:
As mentioned in #487, when we run benchopt with the flag `--no-html`, the barchart and the linechart do not use the same color palette (unlike in the html version). This fixes it.


AmelieVernay:
The function `plot_bar_chart()` has a `plotly` argument but I believe this is never used as the function is not called without the flag `--no-html`. Maybe we could remove it? If so, some modifications would be required in `plotting/__init__.py` as well.

mathurinm (approved):

tomMoral (approved):
LGTM

PR #504, opened 2022-11-24
FIX AttributeError: module 'warnings' has no attribute 'FutureWarning'

mathurinm:
kudos to @AmelieVernay for spotting this


tomMoral (approved):
Thx for catching this.

We should add some test for this at some point.

PR #510, opened 2022-11-26
PLT - enable style reset with multiple examples in doc

AmelieVernay:
Since #487, the solvers styles are not reset between examples. This has effect for instance when building the doc because it runs multiple examples. These modifications might fix that.

### DIFF
--- a/benchopt/plotting/plot_bar_chart.py
+++ b/benchopt/plotting/plot_bar_chart.py
@@ -1,13 +1,12 @@
 import numpy as np
 
 from .helpers_compat import get_figure, _make_bars
-from .plot_objective_curve import get_solver_style, reset_solvers_idx
+from .plot_objective_curve import get_solver_style
 
 PLOTLY_GRAY = (.8627, .8627, .8627)
 
 
-def plot_bar_chart(df, obj_col='objective_value', plotly=False,
-                   reset_style=False):
+def plot_bar_chart(df, obj_col='objective_value', plotly=False):
     """Plot bar chart for a given benchmark and dataset.
 
     Parameters
@@ -18,9 +17,6 @@ def plot_bar_chart(df, obj_col='objective_value', plotly=False,
         Column to select in the DataFrame for the plot.
     plotly : bool
         If set to True, creates a figure with plotly instead of matplotlib.
-    reset_style : bool
-        If set to true, reset the colors and markers associated with each
-        solver before generating the plots.
 
     Returns
     -------
@@ -42,8 +38,6 @@ def plot_bar_chart(df, obj_col='objective_value', plotly=False,
     fig = get_figure(plotly)
     c_star = df[obj_col].min() + eps
     for i, solver_name in enumerate(solver_names):
-        if reset_style and i == 0:
-            reset_solvers_idx()
         col, _ = get_solver_style(solver_name, plotly=False)
         colors.append(col)
         xi = (i + 1.5) * width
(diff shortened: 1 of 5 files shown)


benchopt/plotting/plot_bar_chart.py
```diff
@@ -38,6 +42,8 @@ def plot_bar_chart(df, obj_col='objective_value', plotly=False):
     fig = get_figure(plotly)
     c_star = df[obj_col].min() + eps
     for i, solver_name in enumerate(solver_names):
+        if reset_style and i == 0:
+            reset_solvers_idx()
```
tomMoral, line 46: Maybe better to call before the loop no?

benchopt/plotting/plot_objective_curve.py
```diff
@@ -73,6 +77,8 @@ def plot_objective_curve(df, obj_col='objective_value',
         q1 = df_.groupby('stop_val')['time'].quantile(.1)
         q9 = df_.groupby('stop_val')['time'].quantile(.9)
 
+        if reset_style and i == 0:
+            reset_solvers_idx()
```
tomMoral, line 81: ditto

benchopt/plotting/plot_objective_curve.py
```diff
@@ -145,6 +160,13 @@ def compute_quantiles(df_filtered):
     return q1, q9
 
 
+def reset_solvers_idx():
+    keys = list(solvers_idx.keys())
+    for key in keys:
+        del solvers_idx[key]
```
tomMoral, line 166:
```suggestion
    solvers_idx.clear()
```

benchopt/plotting/plot_objective_curve.py
```diff
@@ -145,6 +160,13 @@ def compute_quantiles(df_filtered):
...
+    return None
```
tomMoral, line 167: Don't put explicit return to `None` as this is the default behavior
```suggestion
```

benchopt/plotting/plot_objective_curve.py
```diff
@@ -145,6 +160,13 @@ def compute_quantiles(df_filtered):
     return q1, q9
 
 
+def reset_solvers_idx():
```
tomMoral, line 163: Could you add a doc string.

For the name, I would say: `reset_solver_styles_idx` to make it clear what the idx represent?

examples/plot_run_benchmark_python_R.py
```diff
@@ -34,5 +34,5 @@
 
 kinds = list(PLOT_KINDS.keys())
 figs = plot_benchmark(save_file, benchmark=Benchmark(BENCHMARK_PATH),
-                      kinds=kinds, html=False)
+                      kinds=kinds, html=False, reset_style=True)
```
tomMoral, line 37: Actually, I am not sure I am in favor of adding such argument to the `plot_benchmark` function. Why not call explicitely `reset_solver_idx` here?

tomMoral (commented):
Looks good but still have a few comments. Feel free to tell me if you think it is better the way you propose.

benchopt/plotting/plot_bar_chart.py
AmelieVernay: It would have been better indeed, but anyway I removed those lines as I removed the `reset_style` argument from `plot_benchmark()`

examples/plot_run_benchmark_python_R.py
AmelieVernay: Done. It seems way better indeed

tomMoral (approved):
This it neat! :)

LGTM! thx @AmelieVernay

PR #511, opened 2022-12-05
ENH single run stopping criterion + improve API doc

tomMoral:


doc/api.rst
```diff
@@ -4,26 +4,18 @@
 Python API
 ==========
 
-Here is a list of Python function available to construct a new benchmark.
-
+Here is a list of Python function available to construct a new benchmark with
```
mathurinm, line 7:
```suggestion
Here is a list of Python functions available to construct a new benchmark with
```

doc/whats_new.rst
```diff
@@ -16,6 +16,13 @@ CLI
 API
 ---
 
+- Add :class:`~benchopt.stopping_criterion.SingleRunCriterion` for solver where
```
mathurinm, line 19: L22 we say "Solvers"

doc/api.rst
```diff
@@ -32,3 +24,26 @@ List of base classes:
    BaseDataset
    BaseSolver
 
+Benchopt run hooks
+~~~~~~~~~~~~~~~~~~
+
+:func:`~benchopt.BaseSolver.skip`: hook to allow skipping some configurations.
```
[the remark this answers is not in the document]
tomMoral: It is a classical term in CS I think: https://en.wikipedia.org/wiki/Hooking
Do you see q clearer one?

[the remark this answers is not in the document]
mathurinm: nope, thanks for the link

benchopt/runner.py
```diff
@@ -90,17 +90,10 @@ def run_one_to_cvg(benchmark, objective, solver, meta, stopping_criterion,
 
     curve = []
     with exception_handler(output, pdb=pdb) as ctx:
+        output.progress('empty run for compilation')
+        solver.warmup_solver()
```
mathurinm, line 94: should we warmup all solvers? I'm thinking of solvers which are very slow even for one iteration (need to compute an SVD for example), this will double their running time

tomMoral: I don't really know what is the best option. Note that I merely harmonized: we used to do it for all `callback` solver, but not for the others...

I like that it is automatic and easy to override by implementing the method.
But I agree it might be costly.

tomMoral: We should decide this point soon. maybe tqlking together?

tomMoral: Actually, warmup does not make sense for `SingleRunCriterion` for instance...
The default should probably depends on many things... I think trying to be smart will shoot us in the foot.

IMO, the solvers should be responsible to make the pre-compilation call in `set_objective` instead of using `warmup` function. To make it easy, we could provide a `run_once` function that take the solver and run it once but the user need to explicitely call it? WDYT?

benchopt/base.py
```diff
@@ -183,24 +187,43 @@ def skip(self, **objective_dict):
 
         return False, None
 
+    def run_once(self):
```
mathurinm, line 190: can we expose stop_val here and pass it L206 for more flexibility? Some solvers need to be run with more than 1 iteration, solvers based on tolerance could use another value too

benchopt/stopping_criterion.py
```diff
@@ -132,7 +136,7 @@ def get_runner_instance(self, max_runs=1, timeout=None, output=None,
                 # and type(solver.get_next) == staticmethod
             ), "if defined, get_next should be a static method of the solver."
             try:
-                solver.get_next(0)
+                solver.get_next(1)
             except TypeError:
                 raise ValueError(
                     "get_next(0) throw a TypeError. Verify that `get_next` "
```
mathurinm, line 142:
```suggestion
                    "get_next(1) throw a TypeError. Verify that `get_next` "
```

doc/advanced.rst
```diff
@@ -159,6 +159,28 @@ Then, these modules and packages can be imported as a regular package, i.e.,
         from benchmark_utils.helper_module.submodule1 import func2
 
 
+
+.. _precompilation:
+
+Caching pre-compilation and warmup effects
+------------------------------------------
+
+For some solvers, such as solver relying on just-in-time compilation with ``numba``, the first iteration might get a longer time due to "warmup" effects. To avoid having such effect in the benchmark results, it is usually advised to call the solver once before running the benchmark, in the ``Solver.set_objective`` method. For solvers with ``stopping_strategy in {'tolerance',  'iteration'}``, simply calling the ``Solver.run`` with a simple enough value is usually enough. For solvers with ``stopping_strategy = 'callback'``, it is possible to call ``Solver.run_once``, which will call the ``run`` method with a simple callback that do not compute the objective value and stop after one iteration.
```
mathurinm, line 168:
```suggestion
For some solvers, such as solver relying on just-in-time compilation with ``numba``, the first iteration might get a longer time due to "warmup" effects. To avoid having such effect in the benchmark results, it is usually advised to call the solver once before running the benchmark, in the ``Solver.set_objective`` method. For solvers with ``stopping_strategy in {'tolerance',  'iteration'}``, simply calling the ``Solver.run`` with a simple enough value is usually enough. For solvers with ``stopping_strategy = 'callback'``, it is possible to call ``Solver.run_once``, which will call the ``run`` method with a simple callback that does not compute the objective value and stops at the first call to callback.
```

doc/whats_new.rst
```diff
@@ -16,6 +16,13 @@ CLI
 API
 ---
 
+- Add :class:`~benchopt.stopping_criterion.SingleRunCriterion` to run a solver
+  only once. This can be used for benchmarking methods where we are interested
+  in objective value at convergence. By `Thomas Moreau`_ (:gh:`511`)
+- Add :func:`~benchopt.BaseSolver.warmup_solver` hook for to control solvers
```
mathurinm, line 22:
```suggestion
- Add :func:`~benchopt.BaseSolver.warmup_solver` hook to control solvers
```

mathurinm (approved):

doc/whats_new.rst
tomMoral: This should actually be removed as we do not have the `warmup_solver` in the chosen solution.

benchopt/base.py
tomMoral: Good point

mathurinm (approved):

PR #514, opened 2022-12-06
API rename cost_curve in objective_list

### LINKED ISSUE #463: API rename `cost_curve` in stopping criterion ?
It's not a curve but a list, and it contains objectives (dicts) that are broader than plain cost. WDYT of `objective_list` ?
 
A priori this part of the code is not exposed so no need for a deprecation cycle.

AmelieVernay:
fix #463


tomMoral (approved):
LGTMm thx @AmelieVernay

PR #523, opened 2022-12-20
WIP Document compute returns

### LINKED ISSUE #522: Add description to objective columns
It would be nice to have a textual description of objective columns displayed in benchopt results I think!
These descriptions could take the form of strings (or docstrings?) written where the objective columns are defined.
Thank you for all the effort and the nice librairie anyways :blush:

mathurinm:
thanks to numpydoc it can be done quite easily.
We need to have valid variable names as keys i nthe dict which is currently not the case (eg in Ridge, "Test loss" cannot be parsed properly)

Once we have retrieved this where do we store it and in which fashion do we display it in the HTML ?

closes #522

### DIFF
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -6,6 +6,7 @@ from datetime import datetime
 import numpy as np
 import pandas as pd
 from mako.template import Template
+from numpydoc import NumpyDocString
 
 from benchopt.benchmark import Benchmark
 from ..constants import PLOT_KINDS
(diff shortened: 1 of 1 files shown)


mathurinm:
@tomMoral is it not redundant to store these as a column of the results dataframe? It seems very redundant as it's the same for each row, but we already do this for the results of `get_sys_info`.

And where would we display it on the HTML?

tomMoral:
I would put is in the benchmark metadata, stored in `benchmark_meta.json`? Not sure it is worth putting in the result file (eventhough the compression in parquet should minimize the impact).
For the display, I would say we could be displayed here?

image

PR #529, opened 2022-12-29
DOC add page on curve sampling and stopping criterion

### LINKED ISSUE #528: Understanding the sampling rate of the objective values in the plots
[User story]
When writing a new solver's run method **it is unclear how often and when exactly are the values of the underlying model gonna be sampled to create the final curves (computing objective values)**.
Going through the code files it seems when using, for instance, `iterations` strategy they might be sampled "logarithmically" until `n_iter`, restarting the `run` method from scratch each time (same for `tolerance` strategy).
When using a callback however they seem to be sampled each time the callback is called, or are they ? It's not clear from the documentation. Is the callback used only to know when to stop the solver or does it have an influence on the sampling rate ?  

Following this absence of documentation on this specific point a lot of questions remain unanswered:
For instance when writing a naïve run method as below with `n_iter` empty, it seems 3-4 points are created by default. Why is that the case ? Is it because there is a default `n_iter` that would create 3-4 points when sampled logarithmically ?
```python
def run(self, n_iter):
    self.my_solver(self.X, self.y, 100)
    return self.my_solver._model
```
Is there a way to control the frequency of the sampling (see #526 for instance) ? What happens when the solver exposes the intermediate values of the model ? Can we take advantage of that without restarting the solver each time ? 

It seems BenchOpt already has answers for all those questions but it is unclear where to find them.

I really think some documentation improvement on this specific part including precise walk-through examples would go a long way into better onboarding users. Even though existing documentation is already great and templates are super cool !

mathurinm:
Closes #528

### DIFF
--- a/doc/convergence_curves.rst
+++ b/doc/convergence_curves.rst
@@ -3,23 +3,28 @@
 Sampling of convergence curves: ``stopping_strategy``
 =====================================================
 
-There are two ways to create a convergence curve for a given solver, controlled by the ``stopping_strategy`` attribute of the solver.
+For each solver, there are two ways to create a convergence curve.
+They are chosen by the ``stopping_strategy`` attribute of the solver.
 
 1. Using iterations or tolerance
 --------------------------------
 
-The first way, used for black box solvers, is to create curves by calling the ``Solver.run`` method several times for each solver with different values for the ``stop_val`` parameter.
-The solver is restarted from scratch at each call to ``run``.
-- if the solver's ``stopping_strategy`` is ``"iteration"``, the number of iterations passed to ``run`` increases geometrically from 1 with a factor XXX, eventually switching to XXX if the curves becomes flat., and the
-- if the solver's ``stopping_strategy`` is ``"tolerance"``, the ``stop_val`` parameter corresponds the the tolerance, and decreases geometrically by a factor of XXX between each call to ``run``, starting from XXX.
 
+The first way is to use ``Solver.stopping_strategy = "iteration"`` or ``Solver.stopping_strategy = "tolerance"``.
+This is used for black box solvers, where one can only get the result of the solver for a given number of iterations or numerical tolerance.
+This stopping strategy creates curves by calling ``Solver.run(stop_val)`` several times with different values for the ``stop_val`` parameter:
+- if the solver's ``stopping_strategy`` is ``"iteration"``, the number of iterations passed to ``run`` increases geometrically by at least 1, starting from 1 with a factor 1.5.
+If the objective curve becomes too flat, the geometric factor is multiplied by 1.2.
+- if the solver's ``stopping_strategy`` is ``"tolerance"``, the ``stop_val`` parameter corresponds to the tolerance, and decreases geometrically by a factor of 1.5 between each call to ``run``, starting from 1. The first call uses a tolerance of 1e38.
 
-The evolution of ``stop_val`` can be controlled on a solver level, by implementing a  ``Solver.get_next`` method, which receives the current value for tolerance/number of iterations, and should return the next one.
+Note that the solver is started from scratch at each call to ``solver.run``.
+For more advanced configurations, the evolution of ``stop_val`` can be controlled on a per solver basis, by implementing a static  ``Solver.get_next`` method, which receives the current value for tolerance/number of iterations, and returns the next one.
 
 2. Using a callback
 -------------------
 
-The solution above can be costly. When the solvers expose the intermediate values of the iterates, it is possible to create the curve in a single solver run, by using a callback method.
+The solution above can be costly.
+When a solver exposes the intermediate values of the iterates, it is possible to create the curve in a single solver run, by using a callback method.
 XXX link to solver code.
 
 When are the convergence curves considered complete: ``StoppingCriterion``
--- a/doc/convergence_curves.rst
+++ b/doc/convergence_curves.rst
@@ -1,7 +1,7 @@
 .. _convergence_curves:
 
-Choose the sampling of convergence curves
-=========================================
+Sampling of the convergence curves
+==================================
 
 For each solver, there are two ways to create a convergence curve.
 They are chosen by the ``stopping_strategy`` attribute of the solver.
@@ -14,9 +14,11 @@ The first way is to use ``Solver.stopping_strategy = "iteration"`` or ``Solver.s
 This is used for black box solvers, where one can only get the result of the solver for a given number of iterations or numerical tolerance.
 This stopping strategy creates curves by calling ``Solver.run(stop_val)`` several times with different values for the ``stop_val`` parameter:
 
-- if the solver's ``stopping_strategy`` is ``"iteration"``, the number of iterations passed to ``run`` increases geometrically by at least 1, starting from 1 with a factor 1.5. Note that the first call uses a number of iterations of 0.
+- if the solver's ``stopping_strategy`` is ``"iteration"``, the number of iterations passed to ``run`` increases geometrically by at least 1, starting from 1 with a factor 1.5.
+  Note that the first call uses a number of iterations of 0.
 
-- if the solver's ``stopping_strategy`` is ``"tolerance"``, the ``stop_val`` parameter corresponds to the tolerance, and decreases geometrically by a factor of 1.5 between each call to ``run``, starting from 1 at the second call. Note that the first call uses a tolerance of 1e38.
+- if the solver's ``stopping_strategy`` is ``"tolerance"``, the ``stop_val`` parameter corresponds to the tolerance, and decreases geometrically by a factor of 1.5 between each call to ``run``, starting from 1 at the second call.
+  Note that the first call uses a tolerance of 1e38.
 
 In both cases, if the objective curve is flat (i.e., the variation of the objective between to points is numerically 0), the geometric rate is multiplied by 1.2.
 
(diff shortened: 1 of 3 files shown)


doc/convergence_curves.rst
```diff
@@ -0,0 +1,38 @@
...
+=====================================================
+
+There are two ways to create a convergence curve for a given solver, controlled by the ``stopping_strategy`` attribute of the solver.
+
+1. Using iterations or tolerance
+--------------------------------
+
+The first way, used for black box solvers, is to create curves by calling the ``Solver.run`` method several times for each solver with different values for the ``stop_val`` parameter.
+The solver is restarted from scratch at each call to ``run``.
+- if the solver's ``stopping_strategy`` is ``"iteration"``, the number of iterations passed to ``run`` increases geometrically from 1 with a factor XXX, eventually switching to XXX if the curves becomes flat., and the
```
jeandut, line 13: Thanks for the great PR ! Could it be explained how exactly is the curve deemed flat with pointers to the lines of code that take care of that ?
I am assuming it uses heuristics similar to this dynamic learning rate scheduler (http://blog.dlib.net/2018/02/automatic-learning-rate-scheduling-that.html) ?

mathurinm: @jeandut I did another pass on the PR, could you point anything that remains unclear? Thanks for the great feedback :pray:

doc/convergence_curves.rst
```diff
@@ -0,0 +1,51 @@
...
+1. Using iterations or tolerance
+--------------------------------
+
+
+The first way is to use ``Solver.stopping_strategy = "iteration"`` or ``Solver.stopping_strategy = "tolerance"``.
+This is used for black box solvers, where one can only get the result of the solver for a given number of iterations or numerical tolerance.
+This stopping strategy creates curves by calling ``Solver.run(stop_val)`` several times with different values for the ``stop_val`` parameter:
+
+- if the solver's ``stopping_strategy`` is ``"iteration"``, ``stop_val`` is  the number of iterations passed to ``run``.
+  It increases geometrically by at least 1, starting from 1 with a factor 1.5.
```
jeandut, line 18: Maybe a small markdown equation would help ? But this is already super helpful.

doc/convergence_curves.rst
```diff
@@ -0,0 +1,51 @@
...
+  Note that the first call uses a number of iterations of 0.
+
+- if the solver's ``stopping_strategy`` is ``"tolerance"``, the ``stop_val`` parameter corresponds to the numerical tolerance.
+  It decreases geometrically by a factor of 1.5 between each call to ``run``, starting from 1 at the second call.
+  Note that the first call uses a tolerance of 1e38.
+
+In both cases, if the objective curve is flat (i.e., the variation of the objective between two points is numerically 0), the geometric rate is multiplied by 1.2.
```
jeandut, line 25: Thank you ! Very helpful !

doc/convergence_curves.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Note that the solver is started from scratch at each call to ``solver.run``.
```
jeandut, line 27:
```suggestion
Note that the solver is restarted from scratch at each call to ``solver.run``.
```

doc/convergence_curves.rst
```diff
@@ -0,0 +1,51 @@
...
+For more advanced configurations, the evolution of ``stop_val`` can be controlled on a per solver basis, by implementing a static  ``Solver.get_next`` method, which receives the current value for tolerance/number of iterations, and returns the next one.
+
+2. Using a callback
+-------------------
+
+restarting the solver from scratch, though inevitable to handle black box solvers, can be costly.
```
jeandut, line 33:
```suggestion
Restarting the solver from scratch, though inevitable to handle black box solvers, can be costly.
```

jeandut (approved):
Thank you very much for this PR. It is super helpful I made minor comments regarding typos on the doc but it's already great !
I have also a few general comments regarding this PR but none of them is blocking:
- Step-by-step examples of using the callback strategy as in ResNet classification benchmark would be nice as it is still not immediately obvious how one would go about implementing a callback in the general case.
- Overall in my opinion there is a confusion between sampling rate and stopping criterion. I.e. one could want to have a uniform sampling rate (every iteration) but a tolerance stopping criterion. Therefore ideally one would like to separate those two concepts in the code as well for more modularity.

doc/convergence_curves.rst
```diff
@@ -0,0 +1,51 @@
...
+
+When a solver exposes the intermediate values of the iterates, it is possible to create the curve in a single solver run, by using ``stopping_strategy = "callback"``.
+In that case, the argument passed to ``Solver.run`` will be a callable object, ``callback``.
+Every time it is called inside ``Solver.run`` with the current iterate ``x`` as argument, this object will store the objective value at the current iterate, and the current running time.
```
tomMoral, line 37: It is not this.
Actually, `stopping_strategy == "callback"` follows the same principle as `"iteration"`, except that `n_iter` is replaced by the number of call to the callback.

The loss is computed only when reaching this `stop_val` iterations and `shoud_stop` is also only called at this moment.

I think I agree with @jeandut that this is a bit a "melange des genres". I am all up to improve this, but the best would be to take time to discuss what we wanna do, to have a proper solution. Happy to chat IRL with you two.

doc/convergence_curves.rst
```diff
@@ -0,0 +1,51 @@
...
+
+For each solver, there are two ways to create a convergence curve.
+They are chosen by the ``stopping_strategy`` attribute of the solver.
+
+1. Using iterations or tolerance
+--------------------------------
+
+
+The first way is to use ``Solver.stopping_strategy = "iteration"`` or ``Solver.stopping_strategy = "tolerance"``.
+This is used for black box solvers, where one can only get the result of the solver for a given number of iterations or numerical tolerance.
```
tomMoral, line 14:
```suggestion
This is used for black box solvers, where one can only get the result of the solver for a given number of iterations or for a given numerical tolerance.
```

tomMoral, line 14:
```suggestion
This is used for black box solvers, where one can only get the result of the solver for a given number of iterations or for a given numerical tolerance.
```

doc/convergence_curves.rst
tomMoral: I would move this comment up, before describing the two strqtegy, to avoid people missing it, considering that it is part of the second strategy. Or it could be highlighted as a `note`.

doc/convergence_curves.rst
mathurinm: I don't get it, there is a single loss computed per run, which is the final one?

doc/convergence_curves.rst
```diff
@@ -0,0 +1,51 @@
...
+
+
+When are the solvers stopped?
+=============================
+
+For each of the sampling strategies above, the solvers continue running (i.e. the callback returns ``True``, the number of iterations passed to ``Solver.run`` increases or the tolerance passed to ``Solver.run`` decreases) until ``Solver.StoppingCriterion.should_stop()`` returns ``True``.
+
+This value is determined by ``Solver.StoppingCriterion.check_convergence()``, based on the objective curve so far.
+For example, the ``check_convergence`` method of a ``SufficientDescentCriterion(eps, patience)`` returns ``True`` when the relative decrease of the objective was less than a tolerance ``eps`` for more than ``patience`` calls to ``check_convergence``.
+
```
tomMoral, line 60:
```suggestion
For each of the sampling strategies above, the solvers continue running (i.e. the callback returns ``True``, the number of iterations passed to ``Solver.run`` increases or the tolerance passed to ``Solver.run`` decreases) until it the ``StoppingCriterion.should_stop()`` associated to the solver ``Solver.stopping_criterion`` returns ``True``.

This method takes into account the maximal number of runs given as ``--max-runs``, the timeout given by ``--timeout`` and also tries to stop the solver if it has converged.
The convergence of a solver is determined by  the ``StoppingCriterion.check_convergence()`` method, based on the objective curve so far.
There are three ``StoppingCriterion`` implemented in ``benchopt``:
- ``SufficientDescentCriterion(eps, patience)`` considers that the solver has converged when the relative decrease of the objective was less than a tolerance ``eps`` for more than ``patience`` calls to ``check_convergence``.
- ``SufficientProgressCriterion(eps, patience)`` considers that the solver has converged when the objective has not decreased by more than a tolerance ``eps`` for more than ``patience`` calls to ``check_convergence``.
- ``SingleRunCriterion(stop_val)`` only call the solver once with the given stop_val. This criterion designed for methods that converge to a given value, when one aim to benchmark final performance of multiple solvers.

```

tomMoral (approved):
Nice one, did a couple suggestion.

Not sure how to link this page on the index. Maybe a better way would be to have a FAQ and put this as a link, with other link to some of the topics in https://benchopt.github.io/advanced.html?

doc/convergence_curves.rst
tomMoral: No, the loss is computed at the same point where it would have been with `stopping_strategy == "iteration"`, and the `should_stop` is also evaluated at the same points.
Basically, `callback` is a mirror for `iteration`, where the number of iterations is equal to the number of call to the callback.

doc/convergence_curves.rst
jeandut: Done !

mathurinm:
@tomMoral merge if new link location is OK

doc/convergence_curves.rst
```diff
@@ -0,0 +1,66 @@
...
+  It increases geometrically by at least 1, starting from 1 with a factor 1.5.
+  Note that the first call uses a number of iterations of 0.
+  The value from one call to the other follows:
+
+  .. math::
+
+    \text{stop_val} = \max(\text{stop_val} + 1, \text{int}(\rho * \text{stop_val}))
+
+- if the solver's ``stopping_strategy`` is ``"tolerance"``, the ``stop_val`` parameter corresponds to the numerical tolerance.
+  It decreases geometrically by a factor of 1.5 between each call to ``run``, starting from 1 at the second call.
```
tomMoral, line 27:
```suggestion
  It decreases geometrically by a factor :math:`\rho=1.5` between each call to ``run``, starting from 1 at the second call.
```

doc/convergence_curves.rst
```diff
@@ -0,0 +1,66 @@
...
+1. Using iterations or tolerance
+--------------------------------
+
+
+The first way is to use ``Solver.stopping_strategy = "iteration"`` or ``Solver.stopping_strategy = "tolerance"``.
+This is used for black box solvers, where one can only get the result of the solver for a given number of iterations or for a given numerical tolerance.
+This stopping strategy creates curves by calling ``Solver.run(stop_val)`` several times with different values for the ``stop_val`` parameter:
+
+- if the solver's ``stopping_strategy`` is ``"iteration"``, ``stop_val`` is the number of iterations passed to ``run``.
+  It increases geometrically by at least 1, starting from 1 with a factor 1.5.
```
tomMoral, line 18:
```suggestion
  It increases geometrically by at least 1, starting from 1 with a factor :math:`\rho=1.5`.
```

doc/convergence_curves.rst
tomMoral: A small pass on this text.
```suggestion
Like with ``stopping_strategy == "iteration"``, the loss is computed after a number of callback's calls that grows geometrically. If the loss was computed after :math:`n` calls, the loss and timing will be computed again when reaching :math:`\max(n+1, \rho * n)` calls to the callback.
The callback makes sure we do not account for loss computation time and also check for convergence every time the loss is computed (as described in the next section).
It returns ``False`` when the solver should be stopped. A classical usage pattern is:

.. code:: python

    def run(self, callback):
        x = ... # Initialize iterate

        while callback(x):
            x = ...  # Update iterate
        self.x = x

```

tomMoral (approved):
Great like this.
Just spotted a couple of typos and did a pass on the callback strategy.

doc/convergence_curves.rst
```diff
@@ -0,0 +1,78 @@
...
+
+    def run(self, callback):
+        x = ... # Initialize iterate
+
+        while callback(x):
+            x = ...  # Update iterate
+        self.x = x
+
+Each call to ``callback`` also calls  ``Solver.StoppingCriterion.should_stop`` and returns a boolean indicating whether or not the ``StoppingCriterion`` deems the solver should be stoppped.
+
```
tomMoral, line 62:
```suggestion
```

tomMoral:
Merging! thanks a lot @mathurinm

Issue #531, opened 2023-01-19
Change default value for timeout


jeandut:
Hi !
Currently the default timeout is 100 seconds. For inattentive users the abrupt stop of the computation is surprising and hard to debug. Maybe like 10 minutes (or 600 seconds) would be more sensible.
Also, not sure how to best implement that but it would be great if users could pass hours or minutes directly. One way to do that would be to overload how click CLI loads this argument and have a regexp parser converting the string to an int in seconds allowing syntax like:
```python
--timeout 3h #understood as 3 hours
--timeout 60 #understood as 60 seconds
--timeout 10m #understood as 10 minutes
```
Again thanks for the great tool !

tomMoral:
Hello,

I actually like the default timeout to `100`as this avoids launching very large computations without thinking about how long it will take and this makes running a first example when you discover `benchopt ` or debugging runs fast to run. 

We could raise it to a larger timeout like `600` (which would still be compatible with the first goal) but then we loose the default "fast to run" property. Not sure what others think on this one? @agramfort @mathurinm ?

For the "hard to debug" part, I am not sure why you say it is surprising? The output is supposed to clearly state that the solver reached a timeout, showing a yellow `done (timeout)` . Is this not explicit when you run this (*e.g.* because we should display the timeout value) or did it not report this properly?

For passing the timeout with `m`, `h`, I am all in for this (this is tiresome to do the conversion). Maybe we can make use `pd.to_timedelta(x).total_seconds()` which should make the right conversion I think, except for integer which would need to be special cased.

jeandut:
Hi !
Thanks for the answer maybe the term "hard to debug" was a bit of an overstatement I apologize. It's just that coming from a "mini-batch SGD DL" background, an ML training almost never takes less than a couple of minutes even for very very lightweight architectures and few data. Therefore I was not expecting such a low timeout. But I feel this is a fundamental difference between the theoretical optimization community that was the intended target audience of benchopt and the DL crowd. 

It does show up on the log, but since it does not fail or interrupt the flow of logs I did not pay attention and was careless and tried to find a bug in the code before I re-checked the logs to find that out. This is why I used the word "inattentive user".

Therefore my very biased viewpoint would be to have indeed a larger timeout.

Regardless of the value of the default passing a certain number of hours instead of seconds would be great !
Regarding the implementation details maybe pandas is indeed the way to go !


Thanks for the feedback anyways !

agramfort:
[...]

PR #535, opened 2023-01-30
FEAT syntactic sugar for timeout

mathurinm:
related to @jeandut suggestion in #531

### DIFF
--- a/benchopt/cli/main.py
+++ b/benchopt/cli/main.py
@@ -188,7 +188,7 @@ def run(config_file=None, **kwargs):
 
     try:
         timeout = int(timeout)
-        timeout = f'{timeout}s'
+        timeout = f'{int(1000 * timeout)}ms'
     except ValueError:
         pass
     timeout = pd.to_timedelta(timeout).total_seconds()
--- a/benchopt/cli/main.py
+++ b/benchopt/cli/main.py
@@ -186,7 +186,7 @@ def run(config_file=None, **kwargs):
     ) = _get_run_args(kwargs, config)
 
     try:
-        timeout = int(timeout)
+        timeout = int(float(timeout))
     except ValueError:  # already under string format
         import pandas as pd
         timeout = pd.to_timedelta(timeout).total_seconds()
--- a/benchopt/cli/main.py
+++ b/benchopt/cli/main.py
@@ -1,7 +1,6 @@
 import yaml
 import click
 import warnings
-import pandas as pd
 from pathlib import Path
 
 from benchopt.benchmark import Benchmark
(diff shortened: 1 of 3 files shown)


tomMoral (approved):
Nice, thx @mathurinm 

just a nitpick on import

benchopt/cli/main.py
```diff
@@ -1,6 +1,7 @@
 import yaml
 import click
 import warnings
+import pandas as pd
```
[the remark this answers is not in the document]
mathurinm: nice catch, I have included the import in the if clause where in which it's needed

benchopt/cli/main.py
```diff
@@ -183,6 +185,12 @@ def run(config_file=None, **kwargs):
         deprecated_objective_filters, old_objective_filters
     ) = _get_run_args(kwargs, config)
 
+    try:
+        timeout = int(timeout)
+    except ValueError:  # already under string format
+        import pandas as pd
+        timeout = pd.to_timedelta(timeout).total_seconds()
```
tomMoral, line 192: This won't work in an environment, as this function returns a float, which gets casted as a string that is not properly process here. In order to make this more robust, I would say either we try to cast floats in line 189 `int(float(timeout))`, we cast the output of `192` to `int`, or we move this processing in the `run_benchmark` (but maybe it is better to check validity before launching computations in another env).

tomMoral (changes requested):
Can you add a test for the proper processing of timeout in a run which is called in an env? I think somehting like this should fail right now:

```python

    @pytest.mark.parametrize('timeout', ['10', '1m', '0.03h', '100s'])
    def test_benchopt_run_in_env_timeout(self, test_env_name, timeout):
        with CaptureRunOutput() as out:
            with pytest.raises(SystemExit, match='False'):
                run([str(DUMMY_BENCHMARK_PATH), '--env-name', test_env_name,
                     '-d', SELECT_ONE_SIMULATED, '-f', SELECT_ONE_PGD,
                     '-n', '1', '-r', '1', '-o', SELECT_ONE_OBJECTIVE,
                     '--no-plot', '--timeout', timeout], 'benchopt', standalone_mode=False)

        out.check_output(f'conda activate {test_env_name}')
        out.check_output('Simulated', repetition=1)
        out.check_output('Dummy Sparse Regression', repetition=1)
        out.check_output(r'Python-PGD\[step_size=1\]:', repetition=6)
        out.check_output(r'Python-PGD\[step_size=1.5\]:', repetition=0)

        # Make sure the results were saved in a result file
        assert len(out.result_files) == 1, out.output

```

benchopt/cli/main.py
mathurinm: int(float(timeout)) seems the best to me!

mathurinm: (because it also supports the edge case `--timeout 100.` passed from command line)

Thanks for the feedback, all addressed

tomMoral: agreed, it seemed the best.

tomMoral (approved):
LGTM, thx @mathurinm

tomMoral:
Could you simply add a what's new entry?

PR #543, opened 2023-03-08
PLT - Show solver description when hovering over it

Badr-MOUFAD:
This adds a way of showing additional information about solvers when placing the cursor over them in the legend.

Here is a [preview of the features](https://badr-moufad.github.io/share-benchmarks/solver_description/validate-plotting-stop-val_benchopt_run_2023-03-08_17h46m44.html) that shows how the solvers' metadata would be displayed.

I see one limitation of this approach is the amount of content we can display.

-------------------
Also related to https://github.com/benchopt/benchopt/pull/523.

### DIFF
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -1,6 +1,5 @@
 import json
 import shutil
-import warnings
 import webbrowser
 from pathlib import Path
 from datetime import datetime
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -125,7 +125,10 @@ def get_results(fnames, kinds, root_html, benchmark_name, copy=False):
 
 
 def get_metadata(df):
-    """Get the benchmark metadata such Objective and solver description.
+    """Get the benchmark metadata.
+
+    Metadata are already available among the columns of `df`.
+    It might be Objective and/or Solvers description.
 
     Returns
     -------
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -109,6 +109,18 @@ def get_results(fnames, kinds, root_html, benchmark_name, copy=False):
         # JSON
         result['json'] = json.dumps(shape_datasets_for_html(df))
 
+        # get solver description
+        solvers_description = df.groupby("solver_name")["solver_description"]
+        solvers_description = solvers_description.unique()
+
+        # convert description to str
+        solvers_description = solvers_description.apply(lambda desc: desc[0])
+
+        # save in result
+        result["solvers_description"] = json.dumps(
+            solvers_description.to_dict()
+        )
+
         results.append(result)
 
     for result in results:
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -181,13 +181,29 @@ def shape_solvers_for_html(df, objective_column):
         df_filtered = df_filtered.dropna(subset=[objective_column])
 
         q1, q9 = compute_quantiles(df_filtered)
+        groupby_stop_val_median = df_filtered.groupby('stop_val').median()
         color, marker = get_solver_style(solver)
+
+        stopping_strategy = df_filtered['stopping_strategy'].unique()
+
+        if len(stopping_strategy) != 1:
+            found_stopping_strategies = ', '.join(
+                f"`{item}`" for item in stopping_strategy
+            )
+
+            raise Exception(
+                "Solver can be run using only one stopping strategy. "
+                f"Expected one stopping strategy "
+                f"but found {found_stopping_strategies}"
+            )
+
+        stopping_strategy = stopping_strategy[0]
+
         solver_data[solver] = {
             'scatter': {
-                'x': df_filtered.groupby('stop_val')['time']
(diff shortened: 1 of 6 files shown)


benchopt/runner.py
```diff
@@ -196,7 +196,8 @@ def run_one_solver(benchmark, dataset, objective, solver, n_repetitions,
             solver_name=str(solver),
             data_name=str(dataset),
             idx_rep=rep,
-            stopping_strategy=stopping_strategy.capitalize()
+            stopping_strategy=stopping_strategy.capitalize(),
+            solver_description=str(solver.__doc__) or ""
```
mathurinm, line 200: 100% cosmetic: if you add a comma at the end of the line, you'll make future diffs easier (notice here the diff is two lines because you had to add a comma to the previously last line, whereas essentially it should be a single one)

mathurinm:
That's a really cool feature! When running it I see a None displayed for skglm : 
image

  in benchmark lasso, with the latest upstream version, and with the command `benchopt run . -f skglm -d leukemia -r 1`
  
  
  Also can you edit the what's new to mention this new feature (in particular, to track when `solver_description` was added to the parquet results)

benchopt/runner.py
mathurinm: I believe str(solver.__doc__) evaluates to `"None"` when there is no docstring so it's True and we never enter the part after the `or` when the solver has no docstring

You can achieve whet you wanted with `str(solver.__doc__) if solver.__doc__ is not None else ""`

Badr-MOUFAD: yes indeed, thanks!

mathurinm (approved):
I can't comment on the JS part, but the rendering is as expected, thanks @Badr-MOUFAD !

benchopt/plotting/generate_html.py
```diff
@@ -109,6 +109,27 @@ def get_results(fnames, kinds, root_html, benchmark_name, copy=False):
...
+        try:
+            solvers_description = df.groupby(
+                by=["solver_name"]
+            )["solver_description"]
+            solvers_description = solvers_description.unique()
+
+            # convert description to str
+            solvers_description = solvers_description.apply(
+                lambda desc: desc[0]
+            )
```
tomMoral, line 124: You can simply do:
```suggestion
            solvers_description = df.groupby(
                by=["solver_name"]
            )["solver_description"].first()
```

benchopt/plotting/html/static/hover_index.css
```diff
@@ -317,3 +317,57 @@ input:checked+.slider:before {
...
+
+.solver-description-title {
+  margin-bottom: 1em;
+  font-weight: bolder;
+}
+
+.solver-description-body {
+  max-height: 450px;
+  overflow-y: auto;
+}
```
tomMoral, line 373: missing endline at the end of the file

benchopt/plotting/html/static/result.js
```diff
@@ -651,6 +666,27 @@ const createLegendItem = (solver, color, symbolNumber) => {
 }
 
 
+function createSolverDescription(legendItem, {title, description}) {
+  // skip if no description was provided
+  if (description === null || description === undefined || description === "")
+    return legendItem;
```
tomMoral, line 672: I think I would prefer to have a legend but with `no description provided` in this case, otherwise, this is weird that there is no hover effect in some cases but a popup in others.

tomMoral:
```suggestion
  if (description === null || description === undefined || description === "")
    description = "No description provided";
```

benchopt/runner.py
```diff
@@ -196,7 +196,9 @@ def run_one_solver(benchmark, dataset, objective, solver, n_repetitions,
             solver_name=str(solver),
             data_name=str(dataset),
             idx_rep=rep,
-            stopping_strategy=stopping_strategy.capitalize()
+            stopping_strategy=stopping_strategy.capitalize(),
+            solver_description=(solver.__doc__
+                                if solver.__doc__ is not None else ""),
```
tomMoral, line 201:
```suggestion
            solver_description=(
                solver.__doc__ if solver.__doc__ is not None else ""
            ),
```

tomMoral (approved):
I like the style very much!
A bunch of comments but otherwise it looks pretty :)

benchopt/plotting/generate_html.py
Badr-MOUFAD: It looks neat and clean :heart_eyes: !

benchopt/plotting/generate_html.py
```diff
@@ -109,6 +109,27 @@ def get_results(fnames, kinds, root_html, benchmark_name, copy=False):
...
+
+            # save in result
+            result["solvers_description"] = json.dumps(
+                solvers_description.to_dict()
+            )
```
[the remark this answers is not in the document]
Badr-MOUFAD: Agree!

Just a small concern, I'm afraid that naming it ``meta`` would conflict with the *real* [``meta`` in ``runner.py``](https://github.com/benchopt/benchopt/blob/bdb21cbfe1279412ccd1e3aa4f62aa16707ed4da/benchopt/runner.py#L194).
Perhaps ``descriptions``?

PR #551, opened 2023-04-18
MTN add circle_ci api token for status report

tomMoral:


mathurinm:
This needs to be merged in order to test its effect because otherwise the workflow from `main` is used : 
https://github.com/benchopt/benchopt/actions/runs/4733211357/workflow

PR #552, opened 2023-05-05
PLT add saved views and config files for plots

AmelieVernay:
Allow plot configuration in `yml` or `ini` files 

- [x] `yml` support for config
- [x] axes limits through config file
- [x] ~set default value for `scale`~
- [x] ~filter datasets through config file~
- [x] clean up

Here is an example `config.yml` to put in the benchmark repo
```{yml}
benchmark_lasso:
  plots:
    - suboptimality_curve
    - bar_chart
    - objective_curve

suboptimality_curve:
  xaxis: None
  yaxis: None
  scale: loglog

objective_curve:
  xaxis: None
  yaxis: [0.0, 1.0]
  scale: linear
```

(poke @tomMoral @tanglef)


benchopt/benchmark.py
```diff
@@ -304,6 +304,9 @@ def _func_cached(**kwargs):
 
     def get_config_file(self):
         "Get the location for the config file of the benchmark."
+        yml_path = self.benchmark_dir / "config.yml"
+        if yml_path.exists():
+            return yml_path
         return self.benchmark_dir / 'config.ini'
```
tomMoral, line 310: I would say in this case, convert the `config.ini` to `yml`, so we can deprecate it, with a warning?

benchopt/plotting/__init__.py
```diff
@@ -39,13 +39,22 @@ def plot_benchmark(fname, benchmark, kinds=None, display=True, plotly=False,
...
+            dict_config[param] = []
+    config_kinds = dict_config["plots"]
     if kinds is None or len(kinds) == 0:
         kinds = config_kinds
 
     if html:
-        plot_benchmark_html(fname, benchmark, kinds, display)
-        return None
+        plot_benchmark_html(fname, benchmark, dict_config["plots"], display,
+                            plot_config=dict_config)
```
tomMoral, line 57: Why do you need to pass both?

tomMoral (commented):
This seems a super nice solution! :) thank you both!

I would simply try to deprecate the `.ini`, by converting to `yml` when it exists with a warning.
Also, if you could add a config in the dummy benchmark so we can see if this works in the exposed artifact, this would be great.

tomMoral:
This starts being a very large PR ^^
To summarize what it does:
- It moves the config files to `yaml`.
- It allows to have a list of "plot configurations" that can be applied by clicking on a button in the interface.
- The first configuration is automatically applied in the HTML.
- There is an export button that allows retrieving a configuration that reproduces the current view of the plot.

You can test the behavior here: https://output.circle-artifacts.com/output/job/e5d35c00-a20d-484f-9257-dd47ffce5d0a/artifacts/0/result.html

Now the remaining problem is how to export this configuration to the result website. I started #637 to have a way to embed the configs in the `parquet` files so we can retrieve them in the result website. But we need to decide when we embed the config in the parquet:
- At the end of `benchopt run`: this makes sense as this is the current config. But as we have not looked at the result yet, the config is probably less relevant.
- Each time we call `benchopt plot`: this is a possibility but might be bothersome.
- When using `benchopt publish`: this is really the time where we need to export the file so it makes sense to export it at this point but this makes having a different config for a bundle upload impossible.
All three together?

Happy to have your thoughts on this one @mathurinm @AmelieVernay @agramfort @jeandut @tanglef @Badr-MOUFAD

Melvin-klein:
Hello !

I've changed the web layout of the feature. You can see them here : https://output.circle-artifacts.com/output/job/0e9d56d6-a36d-479b-9d83-53b189eb1c0a/artifacts/0/result.html

@tomMoral asked me to review the PR. I tested :

- run without any config file,
- with `.ini` global config file,
- with `.ini` benchmark config file,
- `.ini` is converted to yml
- with `.yml` global config file,
- with `.yml` benchmark config file,
- responsive design,
- first view is displayed by default,
- HTML export,
- config export,

All this things are OK for me.

Tests failed on linux but it works on my computer, I don't understand why. I will investigate.

tomMoral:
Thank a lot @Melvin-klein , I really like the new interface.
For the failure, this should have been resolved by #649 , it was unrelated to the PR.

Thanks @jeandut for your input. I just changed the behavior for the config export to download the yaml file. and I let you tell us if you thnk the new UI is better.

For the export, I really like your idea of having a command to process result files, with an option to update the metadata and config. This would also be used to merge two result files, which is something missing. And I agree that we could update the metadata with each call to `plot/run`. I will try to implement this quickly so we can ship this feature in the upcoming `benchopt 1.5` release.

tomMoral:
I think this PR is ready for review/merge :tada: 

For adding a new command to process the results files, I think we can do this in a follow-up PR as this one is already very (too) large.

I will review it Monday again and check if we can add more tests. It would be nice if we can review it by the end of next week and merge it, before releasing benchopt 1.5.

jeandut:
Awesome new look UI wise !

jeandut:
Only minor comments on copy to clipboard instead of actual download of a yaml file but this is just for lazy users like me and could be the topic of another PR !

jeandut:
Also sorry I am confused by the Save As a View button what does it do ?

tanglef:
This PR looks great indeed, and neat new UI ! Just two notes maybe:
- the download config button: I'm switching a lot between (mostly two) benchmarks. Each has a `config.yml` config file. Downloading each ones generates `config.yml` and `config(1).yml`, could we -- if we keep going with the download button road -- have a way to differentiate config files between benchmarks (with their names surely)
- an "unexpected" behavior with views: they are not aware when a trace is hidden, thus axis do not rescale when switching views. I put a small example to illustrate it

| after switching views | what is expected |
| ---- | ---- |
| Capture d’écran du 2023-09-14 17-03-26 | Capture d’écran du 2023-09-14 17-04-33  |

But this is mostly nitpicking, we can deal with these things in future PRs.

tomMoral:
> Only minor comments on copy to clipboard instead of actual download of a yaml file but this is just for lazy users like me and could be the topic of another PR !

I am confused, this is now the behavior to download the config file. But maybe you did not look at the latest version of the PR?
https://output.circle-artifacts.com/output/job/60e945cd-5fac-42b6-90aa-0b1c24d19fec/artifacts/0/result.html


> Also sorry I am confused by the Save As a View button what does it do ? It seems it's trying to download something but it's not working on my end it seems.

This button creates an extra view in the list of views, so that you can come back to it. This is used to create the config.
So when you click on it, you are supposed to see a prompt for a name and then have a new "view" in the view's list.

@tanglef for the download, I think I'd rather keep `config.yml` as this is the name of the config file that we will look for in the benchmark for now, and we don't work with different name for now (we can do this in follow up PR).
For saving the hidden/non hidden solvers, this needs to be done but I'd rather merge this one as is, and add these features in follow up PR.

Thanks a lot to both of you (@tanglef, @jeandut , @Melvin-klein ) for the reviews, I think we can merge this first version and then move on from here.

jeandut:
> > Only minor comments on copy to clipboard instead of actual download of a yaml file but this is just for lazy users like me and could be the topic of another PR !
> 
Yes I was not looking at the last one ! It's great thanks so much !

jeandut:
It's really awesome work you guys did !

PR #556, opened 2023-05-16
PLT - Add objective description

Badr-MOUFAD:
This adds the possibility to provide details about the objective metrics, for instance, value and support size.

It exposes an icon beside ``objective_column`` which when hovered shows a description of the objective metrics.
[Live preview of the feature](https://badr-moufad.github.io/share-benchmarks/prev_objective_desc/generated_df_lasso_bench.html).

------
revival of #523

### DIFF
--- a/benchopt/plotting/html/templates/result.mako.html
+++ b/benchopt/plotting/html/templates/result.mako.html
@@ -59,9 +59,11 @@
                                 <label for="objective_column" class="block">Objective column</label>
                                 <!--don't show description unless it is provided -->
                                 % if result['obj_description'] != "":
-                                    <span class="objective-description-icon" style="cursor: pointer;">
-                                        <svg xmlns="http://www.w3.org/2000/svg" fill="currentColor" class="w-4 h-4" viewBox="0 0 512 512">
-                                            <path d="M464 256A208 208 0 1 0 48 256a208 208 0 1 0 416 0zM0 256a256 256 0 1 1 512 0A256 256 0 1 1 0 256zm169.8-90.7c7.9-22.3 29.1-37.3 52.8-37.3h58.3c34.9 0 63.1 28.3 63.1 63.1c0 22.6-12.1 43.5-31.7 54.8L280 264.4c-.2 13-10.9 23.6-24 23.6c-13.3 0-24-10.7-24-24V250.5c0-8.6 4.6-16.5 12.1-20.8l44.3-25.4c4.7-2.7 7.6-7.7 7.6-13.1c0-8.4-6.8-15.1-15.1-15.1H222.6c-3.4 0-6.4 2.1-7.5 5.3l-.4 1.2c-4.4 12.5-18.2 19-30.6 14.6s-19-18.2-14.6-30.6l.4-1.2zM224 352a32 32 0 1 1 64 0 32 32 0 1 1 -64 0z"/></svg>
+                                    <span class="objective-description-trigger" style="cursor: pointer;">
+                                        <span style="margin-right: 0.5em;">(hover over)</span>
+                                        <svg xmlns="http://www.w3.org/2000/svg" fill="currentColor" class="objective-description-icon" viewBox="0 0 512 512">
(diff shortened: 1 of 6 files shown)


Badr-MOUFAD:
Referring to #523, it was suggested to place the description at the bottom of the sidebar.

WDYT about this way of tackling that? I believe that the sidebar is more valuable for adding visualization capabilities than showing details.

mathurinm:
I like the current design! Would you display information about all objectives possible, or only the selected one?

Badr-MOUFAD:
Here is [a preview of the new feature](https://badr-moufad.github.io/share-benchmarks/prev_objective_desc/dummy_bench_benchopt_run_2023-05-22_13h33m23.html).
Also, I ensured it won't break the older versions.

mathurinm (approved):

mathurinm:
Following discussion with @tomMoral can you try an option where the description is more visible, ie:
below the selector box, showing the first line of the description, and hovering would expand the box?

tomMoral:
maybe not expanding on hovering but showing the tool tip?
I just find it hard to discover that hovering on the question mark shows the description?

Badr-MOUFAD:
My apologies @tomMoral, I'm not sure if I understand well what you mean.

tomMoral:
It took me a while to understand where the description where because the `?` button is very small and not super intuitive IMO.
image

I would rather have the first line of the docstring displayed bellow the selectbox, and when hovering on this line, the current popup that displays the full description.
WDYT?

mathurinm:
I'm +1 on @Badr-MOUFAD's original suggestion here (the plain question mark without effect nor "hover over"), the question mark is explicit enough in my opinion

tomMoral:
Sorry for the late reply.
I understand the argument that it might clug the display if we do this for every drop down menu. For the current suggestion:
- I don't like the pulse effect, it takes too much attention while we just want the feature to be discoverable.
- Instead of `(hover over)`, It would simply put `description`, it is explicit enough?

mathurinm:
@Badr-MOUFAD nice! 
can you add a line in the whatsnew and a small example documentation ?

mathurinm:
Just tested it: 
- the question mark button needs clicking. Hovering seems more natural, WDYT? Don't know if people use it on mobile though
- can you configure the box size so that it has a length of 80 characters? in its current state it's a bit short, while we allow the docstring to go up to 80 characters: 
image

Badr-MOUFAD:
> the question mark button needs clicking. Hovering seems more natural, WDYT? Don't know if people use it on mobile though

Clicking would work for mobile. Do you suggest switching completely to click instead of hover?

> can you configure the box size so that it has a length of 80 characters? in its current state it's a bit short, while we allow the docstring to go up to 80 characters:

Yes indeed, it makes sense to set 80 characters long for the description box

tomMoral:
Do you understand why there is no description in this example:
https://output.circle-artifacts.com/output/job/0f815dbb-0e04-4aee-8d97-1107c2f3afcf/artifacts/0/result.html

should you add a description in the example result to see it?

tomMoral:
the exposed artifact is generated with this call:
`benchopt run benchopt/tests/test_benchmarks/dummy_benchmark -d Simulated[n_features=100,n_samples=100,rho=0]  -o "*[reg=0.1]" -s python-pgd -s cd -s Solver-Test -r 5 -n 15 --output benchmark-result`

Badr-MOUFAD:
> Do you understand why there is no description in this example: https://output.circle-artifacts.com/output/job/0f815dbb-0e04-4aee-8d97-1107c2f3afcf/artifacts/0/result.html
> 
> should you add a description in the example result to see it?


Nothing will be shown if the description was not provided. [code ref](https://github.com/Badr-MOUFAD/benchopt/blob/c01e80c7025a379d6171d7e5a3635adfdb74bc2b/benchopt/plotting/html/templates/result.mako.html#L61-L69)

Do you prefer to show it anyways, for instance "Description not provided"?

tomMoral:
No it is fine like that but can you update the example in the dummy dataset so it shows and we can check the feature o. the webpage?

benchopt/plotting/html/templates/result.mako.html
```diff
@@ -55,7 +55,19 @@ <h1 class="ml11">
                             </div>
                         </div>
                         <div class="mt-8 px-8">
-                            <label for="objective_column" class="block text-sm font-medium text-white">Objective column</label>
+                            <div class="flex flex-row items-center justify-between text-sm font-medium text-white">
+                                <label for="objective_column" class="block">Objective column</label>
+                                <!--don't show description unless it is provided -->
+                                % if result['metadata']['obj_description'] != "":
+                                    <span class="objective-description-trigger" style="cursor: pointer;">
+                                        <span style="margin-right: 0.5em;">(description)</span>
```
tomMoral, line 63: nitpick: remove the parenthesis:
```suggestion
                                        <span style="margin-right: 0.5em;">description</span>
```

benchopt/runner.py
```diff
@@ -187,6 +187,13 @@ def run_one_solver(benchmark, dataset, objective, solver, n_repetitions,
     if stopping_strategy == 'callback':
         stopping_strategy = 'iteration'
 
+    # get objective description
+    # use `obj_` instead of `objective_` to avoid conflicts with
+    # the name of metrics in Objective.compute
+    obj_description = objective.__doc__
+    if obj_description is None:
+        obj_description = ""
```
tomMoral, line 195:
```suggestion
    obj_description = objective.__doc__ or ""
```

tomMoral (approved):
A few nitpicks but otherwise LGTM

tomMoral:
Thanks a lot @Badr-MOUFAD  :)

Issue #561, opened 2023-06-06
AttributeError: type object 'Solver' has no attribute 'requirements'


UR4NVS:
## How to reproduce the error
I got *AttributeError: type object 'Solver' has no attribute 'requirements'* when I tried to launch a benchmark with XGBClassifier solver without installing in my conda environnement the package xgboost.

## Fixing the error
To fix it, I added a variable called `requirements` in the file of the solver causing this error (i.e `xgboost.py`). This variable must contain a list with the name of the solver's conda package (in this case it'll be `requirements = ["py-xgboost"]`).

mathurinm:
Hi @UR4NVS, thanks for your feedback ! 
If the solver is not installed, you indeed need to tell benchopt which packages needs to be installed in order for the solver to work, by adding a `requirements` attribute as you did.
 
 However I am surprised that you got a bug, because if the solver is not installed it should be skipped. Can you share the command that generated the error? And if your benchmark is public, can you share a link to the repo?

tomMoral:
I think the issue is with the `benchopt install` command, when collecting uninstalled solvers. If there is an import error in the `safe_import_context`, it expect to find a `requirements` attribute. If there is none, I would suggest we raise a more comprehensive error :)

@UR4NVS , do you think you could look into  this? this involves changing the line [here](https://github.com/benchopt/benchopt/blob/main/benchopt/utils/dependencies_mixin.py#L102)
to check if the requirements exists and raise an error asking to add a `requirements` attribute in the solver class if not.

Badr-MOUFAD:
Should it be closed after the merge of #562?

mathurinm:
fixed by #562

PR #562, opened 2023-06-08
Created error message for requirements error

UR4NVS:
Added an error message in `install` method of `DependenciesMixin` class to make error `AttributeError: type object 'Solver' has no attribute 'requirements'` understandable.

### DIFF
--- a/benchopt/utils/dependencies_mixin.py
+++ b/benchopt/utils/dependencies_mixin.py
@@ -99,19 +99,18 @@ class DependenciesMixin:
             try:
                 cls._pre_install_hook(env_name=env_name)
                 if cls.install_cmd == 'conda':
-                    try:
+                    if hasattr(cls, "requirements"):
                         install_in_conda_env(*cls.requirements,
                                              env_name=env_name,
                                              force=force)
-                    except AttributeError:
-                        print(
-                            "Missing 'requirements' variable in Solver",
-                            "class.\nRequirements must be a list containing",
-                            "the name of the package that you are",
-                            "using for the solver (except scikit-learn).\n",
-                            "\rExamples :\n",
-                            "\rrequirements = ['package'] (conda package)\n",
-                            "\rrequirements = ['pip:package'] (Pypi package)"
+                    else:
+                        raise AttributeError(
+                            """Missing 'requirements' variable in Solver class.
+                            \rRequirements must be a list containing the name of the package that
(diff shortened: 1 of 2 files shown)


benchopt/utils/dependencies_mixin.py
```diff
@@ -99,8 +99,20 @@ def install(cls, env_name=None, force=False):
             try:
                 cls._pre_install_hook(env_name=env_name)
                 if cls.install_cmd == 'conda':
-                    install_in_conda_env(*cls.requirements, env_name=env_name,
-                                         force=force)
+                    try:
+                        install_in_conda_env(*cls.requirements,
+                                             env_name=env_name,
+                                             force=force)
+                    except AttributeError:
```
mathurinm, line 106: there could be an AttributeError due to another attribute than `requirements` missing. Instead of try except, should we use `if hasattr(cls, "requirement") / else` ?

tomMoral: Yes this would be better

UR4NVS: I pushed a new commit. Tell me if I need to modify again the file

Badr-MOUFAD:
Happy to take over this PR!

I have one concern regarding making a unit test the fix.
In fact, I'm unable to reproduce the bug in #561 when running a benchmark with a solver not having a ``requirements`` attribute.

<details>
<summary>(click to expend solver code)</summary>

```python
from benchopt import BaseSolver
from benchopt import safe_import_context


with safe_import_context() as import_ctx:
    import i_m_not_a_package


class Solver(BaseSolver):
    name = 'test-requirement'

    def set_objective(self, X):
        i_m_not_a_package.break_this_code()
        pass

    def run(self, n_iter):
        pass

    def get_result(self):
        return
```
</details>

As expected the solver is skipped
```bash
Benchopt is running
Simulated                                                                                                                                                                                           
  |--Test objective requirements                                                                                                                                                                    
  File "~/solvers/solver_requirements.py", line 6, in <module>
    import i_m_not_a_package
ModuleNotFoundError: No module named 'i_m_not_a_package'
    |--test-requirement: not installed                                                                                                                                                              
No output produced.  
```

I'm a bit confused,
am I lacking context about the issue, @tomMoral, @mathurinm?

mathurinm:
The problem happens if you try to *install* the solver, no ?

PR #566, opened 2023-06-27
ENH - make ``get_next`` a regular solver method

### LINKED ISSUE #564: FEAT - change ``get_next`` to a regular method instead of ``staticmethod`
## Context

I'm in a situation where I would like to set the ``n_iter`` growth rate depending on
- the actual solver parameters
- the pass-in dataset

For instance, I have a Solver that can be run using either coordinate descent or L-BFGS. However,
L-BFGS requires fewer iterations (~50) compared to coordinate descent (~n_features) to converge.

Therefore, I get either a figure with many juxtaposed points or a figure that drops to the suboptimality in one iteration.

The current design provides the static method ``get_next`` to define how ``n_iter`` is updated.
However, it sets it regardless of the parameters and the dataset.

## Considered alternatives

Currently, I tweak it to get nicely sampled figures. But, it's (almost) impossible when the range size of the dataset is large or the solvers require an excentric number of iterations.

## Suggestion

Could we opt for a design similar to ``Solver.skip``? It is a regular method that has access to all Solver members, and that gets constantly called before ``set_objective``?

@tomMoral, any thoughts?

Badr-MOUFAD:
This enables accessing solvers parameters and attributes by making `get_next` a regular method instead of `staticmethod`

closes #564

### DIFF
--- a/benchopt/tests/test_benchmarks/dummy_benchmark/solvers/solver_test.py
+++ b/benchopt/tests/test_benchmarks/dummy_benchmark/solvers/solver_test.py
@@ -52,6 +52,8 @@ class Solver(BaseSolver):
     def get_result(self):
         return self.w
 
-    @staticmethod
-    def get_next(stop_val):
+    def get_next(self, stop_val):
+        # check that we access instance parameters
+        self.X.shape[0] == self.y.shape[0]
+
         return stop_val + 1
--- a/benchopt/tests/test_benchmarks/dummy_benchmark/solvers/solver_test.py
+++ b/benchopt/tests/test_benchmarks/dummy_benchmark/solvers/solver_test.py
@@ -54,6 +54,6 @@ class Solver(BaseSolver):
 
     def get_next(self, stop_val):
         # check that we access instance parameters
-        self.X.shape[0] == self.y.shape[0]
+        self.raise_error
 
         return stop_val + 1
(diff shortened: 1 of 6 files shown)


Badr-MOUFAD:
Do you think it is worth going through a deprecation cycle for this as it is used extensively in benchmarks?

Any thoughts @tomMoral @mathurinm?

benchopt/tests/test_benchmarks/dummy_benchmark/solvers/solver_test.py
```diff
@@ -52,6 +52,8 @@ def st(self, w, mu):
     def get_result(self):
         return self.w
 
-    @staticmethod
-    def get_next(stop_val):
+    def get_next(self, stop_val):
+        # check that we access instance parameters
+        self.raise_error
```
tomMoral, line 57: I would not bother to do this here, this will already fails if it is only for static as `self` would not be defined in the signature..
```suggestion
```

tomMoral (approved):
LGTM up to a nitpick

mathurinm (approved):

doc/whats_new.rst
```diff
@@ -20,6 +20,9 @@ CLI
 API
 ~~~
 
+- Enable accessing :class:`~benchopt.BaseSolver` attributes in ``get_next``.
```
mathurinm, line 23:
```suggestion
- The ``get_next`` method of :class:`~benchopt.BaseSolver` is no longer static.
```

PR #575, opened 2023-07-03
FIX - set width of solver description to 80 characters

### LINKED ISSUE #572: ENH change solver preview windows size to 80 characters
It's what we did for the Objective details and would fit nicely with the required format for the docstring.

Current state:
image

Badr-MOUFAD:
This  

- [x] homogenize both solver and objective description*
- [x] set a max-width of 80 characters of solver description

also fixes #572


mathurinm:
Tested, works fine:
image
image



@tomMoral with @Badr-MOUFAD we think that we can omit the solver name in the displayed box for concision, do you agree?

tomMoral:
agreed!

mathurinm:
Now looks like this:
image

benchopt/runner.py
```diff
@@ -203,7 +203,7 @@ def run_one_solver(benchmark, dataset, objective, solver, n_repetitions,
             idx_rep=rep,
             stopping_strategy=stopping_strategy.capitalize(),
             obj_description=obj_description,
-            solver_description=solver.__doc__ or "",
+            solver_description=(solver.__doc__ or "").replace("    ", ""),
```
Badr-MOUFAD, line 206: looks :ok_hand: @mathurinm !

benchopt/plotting/html/static/hover_index.css
```diff
@@ -439,4 +442,4 @@ input:checked+.slider:before {
 .solver-description-body {
   max-height: 450px;
   overflow-y: auto;
-}
+}
```
tomMoral, line 445: missing end line

benchopt/runner.py
tomMoral: this will mess up with list for instance no?
why not use the [dedent function](https://www.oreilly.com/library/view/mastering-python-scripting/9781789133226/38c16e80-ef5e-44fe-8b12-241c9d6a0e58.xhtml#:~:text=of%20O'Reilly.-,The%20dedent()%20function,every%20line%20of%20your%20text.&text=text%20is%20the%20text%20to%20dedent%20.)?

tomMoral (approved):

mathurinm:
textwrap.dedent only remove **common** whitespaces to all lines and first line is usually not indented in docstrings. cleandoc.inspect does the same but starting from second line onwards, which is what we want. If one line is more indented than others, it stays indented:
image

PR #576, opened 2023-07-03
API rename `objective.compute` to `objective.evaluate_result` and use dictionary as input

### LINKED ISSUE #592: FEAT access to n_iter in compute
It would be nice to have access to `n_iter` in the `compute` function. This way we could specify at which iteration a given metric should be computed. For example we want to optimize, i.e. define the `stopping_strategy` w.r.t. `value=metric_1`, but compute `metric_2` only every 10^th iteration (as its computation is more costly).

mathurinm:
closes #592 and #604


Badr-MOUFAD:
Where should we put the unittest @mathurinm ?

tomMoral (approved):
LGTM, but I think we need to update the dummy benchmark?

benchopt/tests/test_benchmarks/dummy_benchmark/doc_objective.py
```diff
@@ -36,7 +36,7 @@ def set_data(self, X, y):
         self.X, self.y = X, y
         self.lmbd = self.reg * self._get_lambda_max()
 
-    def compute(self, beta):
+    def evaluate_result(self, beta):
         "Compute the objective value given the output x of a solver."
```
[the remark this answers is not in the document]
tomMoral: We also need to update the solvers.

tomMoral (commented):
Some comments

benchopt/base.py
```diff
@@ -385,16 +384,32 @@ def compute(self, beta):
...
+        # XXX remove in version 1.5
+        if hasattr(self, "compute"):
+            warnings.warn(
+                "objective.compute was renamed `objective.evaluate_result` in "
+                "v 1.4", FutureWarning,
+            )
+            self.evaluate_result = self.compute
+        # XXX remove in version 1.5
+        try:
+            objective_dict = self.evaluate_result(**solver_result)
```
Badr-MOUFAD, line 405: We better write

```python
try:
    kwargs = {**solver_result}
except TypeError:
    warnings.warn("...")
```
to avoid catching potential `TypeErrors` that might occur in `self.evaluate_result`.

WDTY @mathurinm?

benchopt/tests/test_benchmarks.py
```diff
@@ -205,16 +205,16 @@ def _test_solver_one_objective(solver, objective):
             stop_val = 1e-10 if is_convex else 1e-2
         solver.run(stop_val)
 
-    # Check that beta_hat is compatible to compute the objective function
-    beta_hat = solver.get_result()
-    objective(beta_hat)
+    # Check that returned results are compatible with objective
```
tomMoral, line 208:
```suggestion
    # Check that returned results are compatible with the objective
```

tomMoral (approved):
LGTM! I think we can merge this one rather quickly and help everyone move on from `compute` ^^

tomMoral (approved):
LGTM

PR #586, opened 2023-07-04
Implement a tag field for BaseSolver

svaiter:
- [X] Implement a tag field for BaseSolver (e.g. `tags = ['proximal_coordinate_descent', 'python']`)
- [X] Serialize it in the html output
- [ ] Write the necessary JS logic to filter according to this list of tag

### DIFF
--- a/benchopt/runner.py
+++ b/benchopt/runner.py
@@ -181,6 +181,12 @@ def run_one_solver(benchmark, dataset, objective, solver, n_repetitions,
     states = []
     run_statistics = []
 
+    # get tags
+    if hasattr(solver, 'tags'):
+        solver_tags = solver.tags
+    else:
+        solver_tags = []
+
     # get stopping strategy
     # for plotting purpose consider 'callback' as 'iteration'
     stopping_strategy = solver._solver_strategy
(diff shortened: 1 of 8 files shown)


Badr-MOUFAD:
Here is [the preview](https://badr-moufad.github.io/share-benchmarks/preview_tags/benchmark_lasso_benchopt_run_2023-07-10_14h09m55.html) of the visualization feature, and the [benchmark repository](https://github.com/Badr-MOUFAD/benchmark_lasso/pull/4) to reproduce it.
Feedback is very welcome, @svaiter.

Also, how should we handle visualization conflicts between tags and solvers?
In fact, tags are ineffective for solvers hidden by clicking on the legend. I'm thinking about re-initializing the filtering whenever the tags are used. any thoughts?

svaiter:
Excellent ! Thank you.

In my opinion,
- Clicking a dedicated solver should deactivate the tags.
- Reciprocally, clicking a tag should reinitialise the state of the hidden solvers.
- I would prefer I think if the tag cloud is either at the end of the column, or below the solver list.
- I don't know how we should handle if there is a lot of different tag. Maybe a search area allowing to select the tag ? (maybe this is premature optimization...)

benchopt/runner.py
```diff
@@ -182,6 +182,12 @@ def run_one_solver(benchmark, dataset, objective, solver, n_repetitions,
     states = []
     run_statistics = []
 
+    # get tags
+    if hasattr(solver, 'tags'):
+        solver_tags = solver.tags
+    else:
+        solver_tags = []
```
tomMoral, line 189:
```suggestion
    solver_tags = getattr(solver, 'tags', [])
```

tomMoral (commented):
This is a nice start, would be interesting to finish this one during the sprint?

PR #588, opened 2023-07-04
API deprecate solver.stopping_strategy in favor of solver.sampling_strategy

### LINKED ISSUE #577: API rename stopping_strategy to sampling_strategy
It would make the way benchopt samples convergence curves easier to understand, whereas now there can be a confusion between stopping_strategy (how is the curved sampled) and stopping_criterion (where does the curve stop)

mathurinm:
closes #577


benchopt/plotting/generate_html.py
```diff
@@ -230,24 +230,27 @@ def shape_solvers_for_html(df, objective_column):
...
-            stopping_strategy = df_filtered['stopping_strategy'].unique()
+            sampling_strategy = df_filtered['sampling_strategy'].unique()
         except KeyError:
-            stopping_strategy = ["Time"]
-
-        if len(stopping_strategy) != 1:
-            found_stopping_strategies = ', '.join(
-                f"`{item}`" for item in stopping_strategy
+            try:
+                sampling_strategy = df_filtered['stopping_strategy'].unique()
```
mathurinm, line 238: @tomMoral should we preserve backwards compatibility with all parquet/csv? we didn't when we changed from `stop_strategy` to `stopping_strategy`

tomMoral: Yes we should but it will be bothersome...

mathurinm:
@tomMoral how to prevent the html to be opened in firefox when running the tests? It does not seem to happen on other tests.

benchopt/plotting/generate_html.py
```diff
@@ -230,24 +230,27 @@ def shape_solvers_for_html(df, objective_column):
 
         color, marker = get_solver_style(solver)
         # to preserve support of previous benchopt version
-        # where 'stopping_strategy' wasn't saved in solver meta
+        # where 'sampling_strategy' wasn't saved in solver meta
         try:
```
tomMoral, line 234: I am not super fan of using nested try except as this makes the code hard to read.

I would be in favor of using a `if` checking if the column name exists.

benchopt/tests/test_benchmarks.py
```diff
@@ -215,3 +217,37 @@ def _test_solver_one_objective(solver, objective):
...
+    """
+
+    solver2 = solver1.replace("stopping_strategy", "sampling_strategy")
+
+    with temp_benchmark(solvers=[solver1, solver2]) as benchmark:
+        with pytest.warns(
+                FutureWarning,
+                match="'stopping_strategy' attribute is deprecated"):
+            run([str(benchmark.benchmark_dir),
+                 *'-s solver1 -d test-dataset -n 1 -r 1'.split()],
```
tomMoral, line 248: to avoid the opening of Firefox, you can use the `--no-plot` flag

benchopt/plotting/generate_html.py
```diff
@@ -230,24 +230,26 @@ def shape_solvers_for_html(df, objective_column):
...
+                f"`{item}`" for item in sampling_strategy
             )
 
             raise Exception(
-                "Solver can be run using only one stopping strategy. "
-                f"Expected one stopping strategy "
-                f"but found {found_stopping_strategies}"
+                "Solver can be run using only one sampling strategy. "
+                f"Expected one sampling strategy "
+                f"but found {found_sampling_strategies}"
```
tomMoral, line 249: Much simpler to read, thanks :pray:

benchopt/tests/test_benchmarks.py
```diff
@@ -215,3 +217,38 @@ def _test_solver_one_objective(solver, objective):
 
             diff = val_eps - val_star
             assert diff >= 0
+
+
+def test_deprecated_stopping_strategy():
+    # XXX remove in 1.5
```
tomMoral, line 223:
```suggestion
    # XXX remove in 1.5
    assert benchopt.__version__ < '1.5'
```

benchopt/tests/test_benchmarks.py
```diff
@@ -1,9 +1,11 @@
 import pytest
 import numpy as np
 
+from benchopt.cli.main import run
```
tomMoral, line 4:
```suggestion
import benchopt
from benchopt.cli.main import run
```

tomMoral (approved):
Perfect, just adding am `assert` to make sure we remove this in `1.5` :)

PR #590, opened 2023-07-04
ENH: track energy consumption

bmalezieux:
Add an option to visualize performance with respect to energy consumption instead of time.


bmalezieux:
The type of tracker in the code can only be used in online mode, with access to internet. There is an option to use it in offline mode, should we switch to this instead ?

Issue #120, opened 2020-12-21
ENH : adding bibtex / references to credit solvers


josephsalmon:
That would be great to have a uniform way of crediting the papers / solvers added to the various benchmarks.
For the moment, no credit is given (except may be the github where some sources might be coming from).

Bibtex of the papers might be reasonable start.

Any thought on the way to proceed @agramfort ?

agramfort:
I would go with either the dumb solution :

- add a bibtex class attribute to each solver

or (maybe) use a fancy solution like:

- https://github.com/duecredit/duecredit

not sure it's worth it though as it adds quite some dependencies

mathurinm:
@josephsalmon this can go in the solver's docstring, which is displayed on the HTML page:
 
![image](https://github.com/benchopt/benchopt/assets/8993218/a261eb4f-8f2c-46ad-81a0-933afba8916d)

if you find it satisfying I'll let you close the issue

josephsalmon:
That's lovely. Thanks for this solution.

PR #594, opened 2023-07-04
PLT - append references in Solvers and Objective HTML descriptions

### LINKED ISSUE #570: Feature Request: Add Solver reference preview in Benchopt results
It would be nice to have access to reference of each solver in the benchopt resuts page: e.g. a hover on the solver name displays a DOI link to the paper, or the reference implementation.

Badr-MOUFAD:
Follow up of #556 #543,
to append solver and objective references to the shown description. 

fixes #570 and accordingly addresses #120


mathurinm:
Does it fix 570 ? I think it works already, no?

Badr-MOUFAD:
Right @mathurinm!
It is just because we have a dedicated [parameter `references`](https://benchopt.github.io/how.html?highlight=reference) to put academic references?

Should we drop it in favor of docstrings to provide descriptions?

mathurinm:
Yes I'd go this way : it's the chosen way to display the information, and so we don't want to store it in two places. I have started it on the Lasso benchmark for example. https://github.com/benchopt/benchmark_lasso/pull/133

PR #602, opened 2023-07-05
ENH: add a warm_up function in the solvers

pierreablin:
This changes how to warm up solvers that are, e.g., jitted.

Previously, the warm-up was done in `set_objective`, so the warm-up is done even if the algorithm has already been run and the results are cached; warm-up was the bottleneck when results were already cached.

Now, the solvers have a `warm_up` method specified by the user,  and the warm_up method is called on the first `run_one_to_cvg` call. Hence, the warm_up is also in the cached function, so not run when results are already cached.

Not sure about the best way to code this is:

So far I have a `_warm_up` method that calls warm up if the flag `Solver.warmup_done` is False/ does not exist, but it would maybe be more explicit to play with the flag in the `run_one_to_cvg` function, WDYT?

### DIFF
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -200,7 +200,7 @@ class BaseSolver(ParametrizedNameMixin, DependenciesMixin, ABC):
     def run_once(self, stop_val=1):
         """Run the solver once, to cache warmup times (e.g. pre-compilations).
 
-        This function is intended to be called in ``Solver.set_objective``
+        This function is intended to be called in ``Solver.warm_up``
         method to avoid taking into account a solver's warmup costs.
 
         Parameters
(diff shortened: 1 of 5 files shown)


benchopt/base.py
```diff
@@ -230,6 +230,18 @@ def run_once(self, stop_val=1):
...
+        """User specified warm up step, called once before the runs.
+        The time it takes to run this function is not taken into account.
+        The function `Solver.run_once` can be used here for solvers that
+        require jit compilation."""
+        ...
+
+    def _warm_up(self):
+        if not getattr(self, 'warmup_done', True):
+            self.warm_up()
+            self.warmup_done = True
```
mathurinm, line 243: should we rename this to `_warmup_done` ?

tomMoral (approved):
Some nitpicks and docstring improvements but otherwith look good to me.

Issue #600, opened 2023-07-05
Enable to change the axis labels on the plots


apmellot:
file:///home/amellot/Pictures/Screenshot%20from%202023-07-05%2013-54-27.png

In this example the x-axis does not correspond to the number of iterations but to the train set size.  Could it be possible to change the associated label ?

PR #603, opened 2023-07-05
PLT - Allow editing of title and axis label

Badr-MOUFAD:
Just scrutinized the [Plotly JS documentation.](https://plotly.com/javascript/configuration-options/)
I find out that it is possible to enable editing of title and axis labels.

https://github.com/benchopt/benchopt/assets/65614794/7d0db536-b6b7-401d-9324-c52158eb0ddf

Here is a [live preview](https://badr-moufad.github.io/share-benchmarks/editable_plots/benchmark_lasso_benchopt_run_2023-07-05_15h56m19.html) of the feature.

@apmellot your feedback is valuable!
Also related to #600.

### DIFF
--- a/benchopt/plotting/html/static/result.js
+++ b/benchopt/plotting/html/static/result.js
@@ -67,7 +67,7 @@ const makePlot = () => {
   const data = isBarChart() ? getBarData() : getScatterCurves();
   const layout = isBarChart() ? getBarChartLayout() : getScatterChartLayout();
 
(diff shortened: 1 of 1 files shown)


Badr-MOUFAD:
That being said, I'm not 100% convinced with the features
- It is not super intuitive
- The editing is lost when viewing the plot with other params

@tomMoral WDYT?

mathurinm:
shouldn't this rather be done through a config file?

PR #607, opened 2023-07-06
Grid solver

TheoGuyard:
This PR aims to add functionalities helping to set custom x-axis values in `benchopt` plots, that is other than `accuracy` and `iteration`.

### Status - WIP

- [x] `StoppingCriterion` class
- [ ] Tests
- [ ] X-axis name customization

### Current idea

Define a new `GridStoppingCriterion` derived from `StoppingCriterion` and combine it with the iteration-based stopping strategy: each iteration corresponds to one element of the x-axis grid values. The `GridStoppingCriterion` stops the solver once the method has been run on all elements of the grid. **If we think of another way to do this, feel free to share it.** In addition, a functionality to customise the x-axis grid name would be useful.

### A toy illustration 

We want to compare algorithms fitting a Lasso regularization path, *i.e.*, plot a graph with the regularization weight `lmbd` on the x-axis and some performance statistics on the y-axis. With `GridRunCriterion`, this can be done as follows.

```python
# File solver.py
class Solver(BaseSolver):

  name = "lasso_path_solver"
  lmbd_grid = np.logspace(0, -2, 20)
  stopping_strategy = GridRunCriterion(grid=lmbd_grid)

  def set_objective(self, X, y): ...
  def get_result(self): ...
  def run(self, lmbd):
    """Run the solver on a given value of `lmbd` in `lmbd_grid`."""
    self.result = solve_lasso(self.X, self.y, lmbd)
```

In the above example, the `stopping_strategy` is hardcoded in the class. Nevertheless, it could be interesting to set  `lmbd_grid=np.logspace(lmbd_max, lmbd_min, lmbd_num)`, where the arguments of the logarithmic grid are specified by the dataset or the objective of thebenchmark. A potential solution is to pass these arguments in the
`set_objective()` function and to set the `stopping_strategy` here.


paquiteau:
For what is worth, getting this kind information is already possible from benchopt (I think), If the lambda weights are parameters of the solver, then the result file contains lines for every configured solver. It is then a matter of parsing this file, and creating a nice figure from it.  This is in fact close to a discussion I had with @mathurinm, on how to easily parse ans filter the results data... Some kind of ORM filtering (à-la  Django or SQLalchemy) would be nice to have, at least in the API, if not also exposed to the local web frontend. 

Another note: In the case of regularization parameter, the loss is modified, so comparing on this metric won't make sense, as it is solving a different problem.

mathurinm:
I guess what @TheoGuyard has in mind is the case where the loss is the distance to true parameters (for sparse recovery) so the Objective would not depend on lambda.

If one solver depends on lambda and one solver depends on iterations (eg OMP), how can we compare them on a single figure ?

TheoGuyard:
In fact, the feature would be more generic than just fitting regularization paths. Imagine you want to compare the solve time of different methods to optimize some parametric cost function `f(w,theta)` over `w` while varying `theta` such as
 
 - Decision tree models where `theta` is the maximum depth of the tree
 - SVC with polymonial kernel where `theta` is the polynomial degree 
 - Lasso problem where `theta` is the regularization weight
 - ...

Users would probably want to plot the solve time in function of `theta` for all the different methods. My idea is to define some grid of `theta` values and implement a new stopping criterion that allows to call `run` over all the grid values and stops when the end of the grid is reached. Currently, we can tweak the `SufficientProgressCriterion` by setting a `patience=2` and the `get_next()` function of the solver to do this task but I think the workflow could be improved.

tomMoral:
Closing in favor of #671

Issue #597, opened 2023-07-05
BUG safe_import_context incompatible with typing in CI


francois-rozet:
Hello, the CI tests cannot read the requirements of classes (Objective, Solver, Dataset, ...) when their methods are typed.

```python
with safe_import_context() as import_ctx:
    from torch import Tensor

class Objective(BaseObjective):
    name = "fake"
    parameters = {}
    min_benchopt_version = "1.3"

    requirements = [
        "pip:torch",
    ]

    def set_data(self, data: Tensor):
        pass
```

Badr-MOUFAD:
Thanks for flagging this!

I doubt that typed parameters would impair detecting/installing requirements.
Could you pls provide a link to the CI error? Also, a link to the repository would be great!

francois-rozet:
Hello, https://github.com/JuliaLinhart/benchmark_sbi/actions/runs/5461731758/jobs/9940110771.

`Tensor` is imported in `safe_import_context` and used as typing in `set_data`.

agramfort:
it's not a typing issue I think but more that pytorch is not properly installed on the SBI benchmark.

francois-rozet:
It is a typing issue. Switching to the following (outside of the `sage_import_context`) fixes the CI issue (but it is ugly).

```python
try:
    from torch import Tensor
    from torch.distributions import Distribution
except:
    Tensor = TypeVar("Tensor")
    Distribution = TypeVar("Distribution")
```

agramfort:
great !

francois-rozet:
Some features of `importlib` (like `MetaPathFinder`) can be used to modify the behavior of `import`, for instance to create "fake" modules when they are not available, which would solve the current issue.

See
* https://python.plainenglish.io/metapathfinders-or-how-to-change-python-import-behavior-a1cf3b5a13ec
* https://docs.python.org/3/library/importlib.html#importlib.abc.MetaPathFinder.

I think this could be used within `safe_import_context`. What do you think @tomMoral, @mathurinm, @Badr-MOUFAD ? If you think this is worth it, I can try to write a minimal working example.

tomMoral:
This looks indeed promising, creating a fake object automatically when the module is not found would solve many issues.

If you can craft a minimal working example, that would be super helpful

francois-rozet:
[...]

PR #610, opened 2023-07-07
Replace skip import by mock import

francois-rozet:
This is a first draft to fix `safe_import_context` (see #597)


tomMoral:
It would be nice to add some tests and check if this feature can be added during the sprint next week, as this would simplify a lot the issues with missing components.

francois-rozet:
What kind of tests are you thinking of?

francois-rozet:
@tomMoral Your merge reverted part of this PR and broke it (see [diff](https://github.com/benchopt/benchopt/pull/610/files/57b48025df8085825860479c0373f2d86ae7a5cd..edeb6fd35698720fa90b642deee4979f6264ec55)).

tomMoral:
I don't understand the diff you sent. When I look at the change file for the PR, it seems that the merge is okay no?
https://github.com/benchopt/benchopt/pull/610/files

francois-rozet:
The PR introduced a `MOCK_IMPORT` constant and a `mock_import` function in `benchopt/utils/safe_import.py` (see 57b48025df8085825860479c0373f2d86ae7a5cd) which were deleted by the merge.

tomMoral:
Indeed, sorry for the mess, I force-pushed a new merge commit reverting the bad conflict resolution.

For this PR, I would like to add test checking that it works as expected:
- The solver can always be imported, even if the lib does not exist and we access random attributes.
- The fact that the lib is not available is still detectable.

I think that the current implementation did not change anything on the import error for `benchopt install`, and I think a similar strategy could be used. I think @Melvin-klein wanted to take a look at this if you don't have the time @francois-rozet .

francois-rozet:
Ok thx! Yeah, I don't have a lot of time this week/month, so I am glad if @Melvin-klein can take over. I will try to review though!

Melvin-klein:
Hello, I'm working on this PR this week !

benchopt/utils/safe_import.py
```diff
@@ -69,43 +132,29 @@ def __init__(self):
...
-                RuntimeError, "Should not check install with skip import", None
-            )
-            return True
-
-        silence_error = False
+            self.import_error = exc_type, exc_value, tb
 
-        # prevent import error from propagating and tag
+        # Prevent import error from propagating and tag
         if exc_type is not None and issubclass(exc_type, ImportError):
```
tomMoral, line 152: Should never have import error anymore as the import errors are mocked.

benchopt/utils/safe_import.py
```diff
@@ -69,43 +132,29 @@ def __init__(self):
...
-        return self
 
-    def trace(self, frame, event, arg):
-        raise SkipWithBlock()
+        return self
 
     def __exit__(self, exc_type, exc_value, tb):
-
-        if SKIP_IMPORT:
+        if MOCK_ALL_IMPORT or MOCK_FAILED_IMPORT:
```
tomMoral, line 146: This should only be if `MOCK_ALL_IMPORT` or if at least one import has been mocked.

tomMoral (commented):
some comments but overall it looks quite good.

benchopt/utils/safe_import.py
```diff
@@ -1,31 +1,87 @@
...
+            return None
+
+        try:
+            if MOCK_ALL_IMPORT:
+                raise ImportError
+
+            self.specs.append(fullname)
+            # Check if we can import the module {fullname}
+            importlib.import_module(fullname)
+            # If the module can be imported, we let other finders to import it,
```
francois-rozet, line 67: I am not sure mocking only missing modules should be the goal of `mock_import`. `skip_import` skipped every import such that it was fast for benchopt to read the dependency list. Some dependencies are very large and slow to import (for example `torch` takes 1-2s to import).

benchopt/utils/safe_import.py
```diff
@@ -1,31 +1,87 @@
...
 
 
-class SkipWithBlock(Exception):
-    pass
+class MockLoader(Loader):
+    def __init__(self, name):
+        self.name = name
 
+    def create_module(self, spec):
+        return Mock(name=self.name)
```
francois-rozet, line 25: I don't know if it would be worth it, but it is possible to configure `Mock` sub-classes to throw errors when some event happens. For example, when an attribute is called (`mocked_module.attribute()` raises an exception). With the base `Mock` object, calls return another `Mock` object.

For more versatility, we could also use [`MagicMock`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock) which implements `__getitem__` and `__setitem__`. This would enable things like `mocked_module.CONFIGS["key"]`.

tomMoral:
Thanks a lot @francois-rozet for the PR, it was a super interesting discussion/idea.
We finally went with a different direction to parse static arguments directly with `ast`.
It is less complicated and lead to less corner cases.
Closing this one as it significantly diverged from the `main` branch.

Issue #606, opened 2023-07-06
Matlab Support ?


paquiteau:
Hello there, thanks again for the sprint ! I might be opening pandora's box: Should Benchopt provide the possibility to run Benchmark of MATLAB codes ?  There exists some algorithms I would like to compare in the literature, and to be fair I would like to use the author's original implementation. 

Running MATLAB from python is fairly easy (using the `matlabengine` package,  you can even gives numpy array as input). I guess this won't be very different from the Julia or R solver already available.

agramfort:
not objection on my end but it would need to see how to test this on CIs

paquiteau:
The CI/testing is also one of my main concern, Turns out that there is https://github.com/matlab-actions Which does most of the heavy lifting, I will propose a draft PR soon to see if and how we can get this done. Help will be welcome :wink:

tomMoral:
No objection on my side either of we can provide a `MatlabSolver` that we are able to  test on the CI.

PR #612, opened 2023-07-10
Add support for MATLAB solver

paquiteau:
This add support for MATLAB-based solver, and is based on the [matlab-engine](https://www.mathworks.com/help/matlab/matlab_external/install-the-matlab-engine-for-python.html) package.

I also provide a dummy example of PGD solver using MATLAB. 

As mentioned in #606 , the CI is the main issue here. Running matlab from the engine in CI is not supported by Mathworks 
[^1] and would required to have a self hosted runner with a MATLAB license.

[^1]: https://github.com/matlab-actions/setup-matlab/issues/22

This make the CI fail for now.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,10 +52,6 @@ jobs:
           # Use miniforge to only get conda-forge as default channel.
           miniforge-version: latest
 
-      - name: Set up MATLAB
-        uses: matlab-actions/setup-matlab@v1
-      - run: conda info
-
       - name: Install benchopt and its dependencies
         run: |
           conda info
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,6 +52,8 @@ jobs:
           # Use miniforge to only get conda-forge as default channel.
           miniforge-version: latest
 
+      - name: Set up MATLAB
+        uses: matlab-actions/setup-matlab@v1
       - run: conda info
 
       - name: Install benchopt and its dependencies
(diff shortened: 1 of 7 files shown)


paquiteau:
Creating a self hosting runner is pretty easy [^1], choosing who is going to pay for it and the associated MATLAB license is harder (short answer: not me) . Then its a matter of switching the CI to the self hosted runner in the repo setting.

[^1]: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners

tomMoral:
It seems to me that `github-action` allows to use matlab for testing purposes without license on public repo no?
https://github.com/matlab-actions#setup-matlab

paquiteau:
Let me try, but the issue I mentioned in the opening post does not really give hope 😕

tomMoral:
Ah ouais ça a l'air de crash venère... ^^

paquiteau:
So the CI should now pass, as the MATLAB solver is not run if we can't find a matlab installation. 
If a dedicated runner can be set up that would be ideal but that's the best I can do for now.

paquiteau:
Hello there *swiftly hiding the shovel to dig this PR out* 
There is now  MATLAB github actions that could do the trick for us:

https://github.com/matlab-actions/setup-matlab/
https://github.com/matlab-actions/run-command

They don't require licence key  for our use case (public, and no use of MATLAB compilers).

I will see in the coming days if I can propose a correct CI setup for this, but any help is welcome

PR #623, opened 2023-07-17
ENH implement cv

chris-mrn:


benchopt/base.py
```diff
@@ -517,3 +518,45 @@ def __reduce__(self):
...
+                    "named split in your objective class. "
+                    "the type of the function should be "
+                    "(self: Self@Objective, index_train "
+                    "index_test) -> splitted_data"
+                )
+            else:
+                return self.custom_split(*arrays)
+
+        if not hasattr(self, "_cv"):
+            self._cv = itertools.cycle(self.cv.split(*arrays))
```
tomMoral, line 561: This should go above l.547, so you only do it once in this function, and don't need to do it in `custom_split`.

benchopt/base.py
```diff
@@ -517,3 +518,45 @@ def __reduce__(self):
...
+
+        if not hasattr(self, "cv"):
+            raise ValueError(
+                "To use get_split, you need to define "
+                "a cv attribute in your objective class "
+                "It should be a sklearn.model_selection.BaseCrossValidator "
+                "object."
+            )
+
+        if not str(type(arrays[0])) == "<class 'numpy.ndarray'>":
```
tomMoral, line 547: Even if the array is a `ndarray`, you might want a custom split, so I would not have this test.

Also, to properly test if you have an `ndarray`, you can use `isinstance(arrays[0], np.ndarray)`

tomMoral: Also, typically, such split works with `pandas.DataFrame`  `torch.Tensor`, `sccipy.sparse.csc_matrix`, so it will be very hard to check if this is gonna work or not. I would just say, if no `split` is defined, just call the `default_split`.
If it does not work, raise an error saying that this is not compatible.

benchopt/base.py
```diff
@@ -517,3 +518,45 @@ def __reduce__(self):
...
+        i_train, i_test = next(self._cv)
+        res = ()
+        for x in arrays:
+            res = (*res, x[i_train], x[i_test])
+        return res
+
+    def custom_split(self, *arrays):
+        if not hasattr(self, "_cv"):
+            # this define _cv only once
+            self._cv = itertools.cycle(self.cv.split(*arrays))
```
tomMoral, line 532:
```suggestion
```

benchopt/base.py
```diff
@@ -517,3 +518,45 @@ def __reduce__(self):
...
+        i_train, i_test = next(self._cv)
+        return self.split(i_train, i_test)
```
tomMoral, line 534: It may be that `cv` returns something more complex, so I would be more generic:
```suggestion
        split_data = next(self._cv)
        return self.split(*split_data)
```

benchopt/base.py
chris-mrn: Okay 👍  I look at this test also but I thought that it was less convenient because we need to import numpy

chris-mrn: Okay 👍 I looked at this test also but I thought that it was less convenient because we need to import numpy

benchopt/base.py
chris-mrn: maybe you forgot to change something here? I don't see the difference

benchopt/base.py
chris-mrn: But if we say to the user that he needs to define a cv object from sklearn, next(cv) will always return train_index, test_index no?

benchopt/benchmark.py
chris-mrn, line 1: maybe you forgot here buffer_iterator ?

benchopt/benchmark.py
```diff
@@ -486,6 +483,10 @@ def get_all_runs(self, solver_names=None, forced_solvers=None,
                 output.show_status('not installed', dataset=True)
                 continue
             output.display_dataset()
+            all_objectives, objective_buffer = _filter_classes(
```
tomMoral, line 486:
```suggestion
            all_objectives = _filter_classes(
```

benchopt/benchmark.py
tomMoral: No I simply forgot to remove the `objective_buffer` output variable.

doc/user_guide/ml_benchmark.rst
```diff
@@ -0,0 +1,74 @@
...
+        ...
+        def set_data(self, X, y):
+            self.X, self.y = X, y
+            # Specify a cross-validation splitter as the ``cv`` attribute.
+            # This will be automatically used in ``self.get_split`` to split
+            # the arrays provided.
+            self.cv = GroupKFold(n_splits=5, random_state=self.seed)
+
+            # If the cross-validation requires some metadata, it can be
+            # provided in the ``cv_metadata`` attribute. This will be passed
```
mathurinm, line 37: To be explicit this will be passed to self.cv.split and self.cv.get_n_splits ?

tomMoral: yes!

mathurinm (approved):

PR #625, opened 2023-07-18
FIX ignore hidden files when loading datasets, solvers, etc

antoinecollas:
When working on remote machines (e.g. with sshfs), macos automatically creates hidden files.
For example, if there is a file `datasets/mnist.py`, then it automatically creates a metadata file `datasets/._mnist.py`.
Thus, benchopt loads `datasets/mnist.py` but also `datasets/._mnist.py` which creates errors since  `datasets/._mnist.py` only contains metadata. To prevent this, this PR changes the regex that loads submodules to ignore hidden files.

### DIFF
--- a/benchopt/tests/test_benchmark_features.py
+++ b/benchopt/tests/test_benchmark_features.py
@@ -103,22 +103,29 @@ def test_error_reporting(error, raise_install_error):
 
 
 def test_ignore_hidden_files():
+    # Non-regression test to make sure hidden files in datasets and solvers
+    # are ignored. If this is not the case, the call to run will fail if it
+    # is not ignored as there is no Dataset/Solver defined in the file.
     with tempfile.NamedTemporaryFile(
         dir=str(DUMMY_BENCHMARK_PATH / 'datasets'),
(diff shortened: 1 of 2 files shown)


agramfort (approved):

mathurinm:
Should we add a UT ?

tomMoral:
Yes, can you add a test? You can use something similar to what is done here: https://github.com/benchopt/benchopt/blob/79e4608b6a26a02d832668bc7d18b1960b311b69/benchopt/tests/test_benchmark_features.py#L54C1-L54C1

benchopt/tests/test_benchmark_features.py
```diff
@@ -99,3 +100,32 @@ def raise_error():
             )
     finally:
         os.environ['BENCHOPT_RAISE_INSTALL_ERROR'] = prev_value
+
+
+def test_ignore_hidden_files():
```
tomMoral, line 105: I just simplified a bit this test but it looks super good! thanks @antoinecollas

tomMoral (approved):
LGTM!

PR #630, opened 2023-07-21
BUG - ``evaluate_result`` breaks down when it accepts a single dict argument

Badr-MOUFAD:
After #576,

It happens that the depreciation cycle doesn't work for benchmarks that anticipated passing in ``solver_result`` as a dictionary.
Indeed, this check destruct the input dict

https://github.com/benchopt/benchopt/blob/c689d242ac7910d9f1b4dbf9b0dff915925559e2/benchopt/base.py#L439-L446 

resulting in

```bash
TypeError: compute() got an unexpected keyword argument 'x'
```

as illustrated in the PR change.

------

Pinning @agonon for flagging that

### DIFF
--- a/benchopt/tests/test_benchmarks/dummy_benchmark/objective.py
+++ b/benchopt/tests/test_benchmarks/dummy_benchmark/objective.py
@@ -38,7 +38,9 @@ class Objective(BaseObjective):
     def get_one_solution(self):
         return np.zeros(self.X.shape[1])
 
-    def evaluate_result(self, beta):
+    def evaluate_result(self, solver_result):
+        beta = solver_result["beta"]
+
         diff = self.y - self.X.dot(beta)
         objective_value = .5 * diff.dot(diff) + self.lmbd * abs(beta).sum()
 
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -432,12 +432,17 @@ class BaseObjective(ParametrizedNameMixin, DependenciesMixin):
             #   "Rename objective.compute to objective.evaluate_result")
             warnings.warn(
                 "objective.compute was renamed `objective.evaluate_result` in "
-                "v 1.5", FutureWarning,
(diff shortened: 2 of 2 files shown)


benchopt/tests/test_benchmarks/dummy_benchmark/objective.py
```diff
@@ -38,7 +38,9 @@ def skip(self, X, y):
     def get_one_solution(self):
         return np.zeros(self.X.shape[1])
 
-    def evaluate_result(self, beta):
+    def evaluate_result(self, solver_result):
+        beta = solver_result["beta"]
```
mathurinm, line 42: I don't see how this fixes the issue you mention, are all files committed?

tomMoral:
This issue should be fixed by #631

benchopt/tests/test_benchmarks/dummy_benchmark/objective.py
mathurinm: I'm not sure this setting should happen: the objective no longer has a compute method, so we expect the benchmark has been updated, and evaluate_result is correctly implemented (takes an unpacked dict as input). 

Do you confirm @Badr-MOUFAD ?

tomMoral:
This should have been resolved in #631 . Feel free to reopen if this is not the case.

benchopt/tests/test_benchmarks/dummy_benchmark/objective.py
Badr-MOUFAD: > I don't see how this fixes the issue you mention, are all files committed?

It wasn't actually a fix, it was to point out where it would break

> I'm not sure this setting should happen: the objective no longer has a compute method, so we expect the benchmark has been updated, and evaluate_result is correctly implemented (takes an unpacked dict as input).

Aren't we going through a deprecation cycle? In fact, #576 breaks the benchmark passing in dict to compute.

PR #631, opened 2023-07-21
FIX abstract classes in BaseObjective

tomMoral:
This fixes the fact that `BaseObjective` was not raising error if an abstractmethod is not implemented.
It also makes the backward compatibility between `compute/evaluate_result` work with results that are dictionaries.

Finally, it renames `Objective.get_one_solution` into `Objective.get_one_result`.

### DIFF
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -422,8 +422,9 @@ class BaseObjective(ParametrizedNameMixin, DependenciesMixin, ABC):
         # XXX remove in version 1.5 and make this method abstract
         if hasattr(self, "compute"):
             warnings.warn(
-                "`Objective.compute` was renamed `Objective.evaluate_result`. "
-                "This will raise an error starting v1.5", FutureWarning,
+                "`Objective.compute` was renamed `Objective.evaluate_result` " 
+                "in v 1.5, and now takes as input the unpacked dict returned "
+                "by `Solver.get_result`", FutureWarning,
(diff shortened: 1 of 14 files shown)


doc/whats_new.rst
```diff
@@ -13,6 +13,10 @@ Version 1.5 (dev)
 API
 ~~~
 
+- Deprecate ``Objective.get_one_solution`` in favor of ``Obejctive.get_one_result``
```
mathurinm, line 16:
```suggestion
- Deprecate ``Objective.get_one_solution`` in favor of ``Objective.get_one_result``
```

benchopt/base.py
```diff
@@ -418,32 +419,34 @@ def evaluate_result(self, **solver_result):
             dictionary, multiple metric values can be stored at once instead
             of running each separately.
         """
-        ...
+        # XXX remove in version 1.5 and make this method abstract
+        if hasattr(self, "compute"):
+            warnings.warn(
+                "`Objective.compute` was renamed `Objective.evaluate_result`. "
+                "This will raise an error starting v1.5", FutureWarning,
```
mathurinm, line 426:
```suggestion
                "`Objective.compute` was renamed `Objective.evaluate_result` " 
                "in v 1.5, and now takes as input the unpacked dict returned "
                "by Solver.get_result", FutureWarning,
```

PR #629, opened 2023-07-21
DOC - Enhance installation, running and visualizing benchmarks

Badr-MOUFAD:
Followup of #619. It brings the following changes

- [x] reformulate get started page
- [x] add run benchmark page
- [x] add visualize benchmark page

### Checks before merging PR
- [ ] ~added documentation for any new feature~
- [ ] ~added unit test~
- [x] edited the [what's new](../../whatsnew.rst)

PR #619, opened 2023-07-17
DOC - rearrange ``benchopt`` documentation

Badr-MOUFAD:
This enhances the accessibility of the documentation by

- [x] changing the theme of documentation
- [x] reorganizing its content

[Link to preview](https://output.circle-artifacts.com/output/job/581c8468-aaa2-488c-8e95-41bd69f372b3/artifacts/0/dev/index.html).
 
**PS.** This PR is a plain reuse of the old documentation content. Hence there might be typos and/or missing sections ...

PR #635, opened 2023-07-31
DOC - Tutorial to add a new solver

Badr-MOUFAD:
Building on #619 and #629,
This adds a use case for adding a new solver to a benchmark.

### Checks before merging PR
- [ ] ~added documentation for any new feature~
- [ ] ~added unit test~
- [ ] edited the [what's new](../../whatsnew.rst)

### DIFF
--- a/doc/tutorials/add_solver.rst
+++ b/doc/tutorials/add_solver.rst
@@ -17,7 +17,7 @@ Before the implementation
 -------------------------
 
 A solver is a Python class that lives in a standalone Python file.
-It has a unique name that distinguish it from other solvers in the benchmark.
+It has a unique name that distinguishes it from other solvers in the benchmark.
 
 Start by adding a new file ``skglm.py`` to the ``solvers/`` directory.
 Notice that we named the python file ``skglm`` as well as the solver,
@@ -43,13 +43,14 @@ imports in the context manager ``safe_import_context``, except benchopt imports.
     skip uninstalled solvers, ... For more details refer to
     :class:`~benchopt.safe_import_context` documentation.
 
+
 Implementation
 --------------
 
 - **Specifying the solver parameters**
 
 Define a class attribute ``parameters`` to add parameters describing the internal
-functioning of the solver. This attribute is dictionary whose keys are the parameters
+functioning of the solver. This attribute is a dictionary whose keys are the parameters
 of the solver.
 
 For ``skglm``, the Lasso estimator has arguments to specify
@@ -74,7 +75,7 @@ the working set strategy and its initial size.
 - **Initializing the setup**
 
 Use the method ``set_objective`` to pass in the dataset and the objective parameters to
-the solver. Besides, it the ideal method to define unchanging object across the solver run.
+the solver. Besides, it is the ideal method to define unchanging objects across the solver run.
 
 Here we use it to store references to the dataset ``X, y``.
 Also, we store the Lasso estimator as it will not change during the runs. 
@@ -95,11 +96,11 @@ Also, we store the Lasso estimator as it will not change during the runs.
 
 - **Describing the run procedure**
 
-The ``run`` method combined with ``sampling_strategy`` describe how the
+The ``run`` method combined with ``sampling_strategy`` describes how the
 performance curves are constructed. In particular, the ``sampling_strategy`` dictates
 the input signature of ``run`` and how it will be called by benchopt.
 
-Here we use *iteration* as sampling strategy. Following this choice, the ``run``
+Here we use *iteration* as a sampling strategy. Following this choice, the ``run``
 will be called repetitively with an increasing number of iterations.
 
 .. code-block:: python
@@ -131,8 +132,8 @@ will be called repetitively with an increasing number of iterations.
 We define a ``get_result`` method to pass the result of the ``run`` back
 to the objective.
 
-Here we simply define a method that returns the solution as dictionary since
-we are not post processing on the solution.
+Here we simply define a method that returns the solution as a dictionary since
+we are not post-processing on the solution.
 
 .. code-block:: python
 
@@ -188,13 +189,13 @@ Refinement
 ----------
 
 ``skglm`` relies on Numba JIT-compilation for fast numerical computation
-which comes at the expense of a initial overhead in the first run.
+which comes at the expense of an initial overhead in the first run.
 Ideally, we would like to disregard that in the benchmark results.
 
 To address this need, benchopt features a ``warm_up`` hook that is called
 once before the actual solver run.
 
-in our case, we define it as follows
+In our case, we define it as follows
 
 .. code-block:: python
 
@@ -206,4 +207,5 @@ in our case, we define it as follows
 
 .. hint::
 
-    Learn about the other hooks of benchopt in the :ref:`API references <benchopt_hooks>` page.
+    Learn about the other hooks of benchopt in the
+    :ref:`API references <benchopt_hooks>` page.
--- a/doc/tutorials/add_solver.rst
+++ b/doc/tutorials/add_solver.rst
@@ -38,9 +38,26 @@ import in the context manager ``safe_import_context``.
 
 Implementation
 --------------
-- constructor for getting parameters
-- set_objective specify the setup (combination of dataset and objective parameters)
-- get_results called by objective to evaluate metrics
+
+Specifying the solver parameters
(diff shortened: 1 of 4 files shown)


doc/tutorials/add_solver.rst
```diff
@@ -0,0 +1,331 @@
...
+        ...
+        def get_result(self):
+            return {'beta': self.beta}
+        ...
+
+
+With these methods being implemented, your solver is now ready to be run!
+
+
+Additional features
```
tomMoral, line 216: This feels like too much information for a tuto, and more like a user guide of the solver features.
I would move this to another page on `The Solver object` maybe?
To me, a tuto is more step-by-step and the written file should work at the end (a bit like what jeremy did, with a screenshot of the resulting change).

tomMoral (changes requested):
I have a few comments on the tuto:

- I feel like the lasso benchmark is too big to be a good tutorial benchmark.
I would rather have a `tuto_benchmark` in `benchopt`, with a simple task like ridge regression and clone this one.
- To me, a tutorial is a step by step construction that allow to run something at the end of the tuto. right now, this is not possible as neither of the 3 implementations works. I would focus on a single implementation that works, maybe using a `scikit-learn` solver if it is easier?
- The additional features should be part of the user_guide more than the tuto.

mathurinm:
> I have a few comments on the tuto:
> 
>     * I feel like the lasso benchmark is too big to be a good tutorial benchmark.
>       I would rather have a `tuto_benchmark` in `benchopt`, with a simple task 
>       like ridge regression and clone this one.
> 
>     * To me, a tutorial is a step by step construction that allow to run something at the end of the tuto. 
>      right now, this is not possible as neither of the 3 implementations works. I would focus  
>      on a single implementation that works, maybe using a `scikit-learn` solver if it is easier?
> 
>     * The additional features should be part of the user_guide more than the tuto.

Did a pass:
- replaced Lasso by Ridge
- explicit implementations of set_objective and run in the case of an sklearn
- moved "parameters" after implementation of the 3 methods, can be reverted. I liked having 2 parameters and apart from "solver", sklearn does not really have a useful one ; + the parameter name is `solver` which is unfortunate.
 
I still believe that people won't follow it exactly, but read it and adapt on the fly to their benchmark ; so I'm not convinced they will run the bench in the end and that we need to have a `benchmark_tuto`.
In fact I included the relevant parts of `objective.py`, so the tutorial does not require reading content from `benchmark_ridge`.

doc/tutorials/add_solver.rst
```diff
@@ -0,0 +1,331 @@
...
+Next, we move to the implementation of the three key methods a solver must define.
+As a reminder, the workflow of benchopt is depicted in the figure below.
+We'll implement the methods in the order in which they get called when running the benchmark: firstly ``set_objective``, then ``run`` and finally ``get_result``.
+
+.. figure:: ../_static/benchopt_schema_dependency.svg
+   :align: center
+   :width: 90 %
+
+
+Initializing the setup
```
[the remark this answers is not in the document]
mathurinm: done!

doc/tutorials/add_solver.rst
```diff
@@ -0,0 +1,331 @@
...
+The ``run`` method combined with ``sampling_strategy`` describes how the performance curves of the solver will be constructed.
+
+.. hint::
+
+    The :ref:`Performance curves page <performance_curves>` provides a complete guide on the way benchopt constructs performance curves, and on the different sampling strategies.
+
+There are three possible choices for ``sampling_strategy``: **iteration**, **tolerance**, and **callback**.
+We show how to implement the ``run`` method for each one of them.
+
+- **iteration**
```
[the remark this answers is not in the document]
mathurinm: slightly disagree, I don't think it's too verbose/complicated to have the 3 strategies here and it has the benefit of  making the page self contained

doc/tutorials/add_solver.rst
mathurinm: makes sense!

doc/tutorials/add_solver.rst
```diff
@@ -234,24 +234,21 @@ Specifying the solver parameters
 If your solver has hyperparameters, you can specify them by adding an attribute ``parameters``.
 This attribute is a dictionary whose keys are the solver's hyperparameters.
 
-For example, if our solver has two hyperparameters, ``stepsize`` and ``momentum``, we implement them as follows:
+For our ridge estimator, we can set the two hyperparameters ``positive`` and ``solver`` as follows
 
 .. code-block:: python
-    :caption: benchmark_ridge/solvers/mysolver.py
+    :caption: benchmark_ridge/solvers/sklearn.py
```
mathurinm, line 239: I'm not convinced: the sklearn file already exists in the benchmark. I'd keep mysolver.

doc/tutorials/add_solver.rst
```diff
@@ -234,24 +234,21 @@ Specifying the solver parameters
 If your solver has hyperparameters, you can specify them by adding an attribute ``parameters``.
 This attribute is a dictionary whose keys are the solver's hyperparameters.
 
-For example, if our solver has two hyperparameters, ``stepsize`` and ``momentum``, we implement them as follows:
```
mathurinm: this is just an illustration, doesn't have to match the solver we implemented. It comes at the end of the tuto so people should be more familiar with the principle.

The proposed change is not great either because `positive` changes the optimization problem/

doc/tutorials/add_solver.rst
```diff
@@ -0,0 +1,252 @@
...
+.. code-block:: bash
+
+    benchmark_ridge/
+    ├── objective.py     # existing implementation of the Objective
+    ├── datasets/
+    │   ├── ...          # existing datasets
+    └── solvers/
+        ├── mysolver.py  # our newly added solver
+        ├── ...          # other solvers
+
```
mathurinm, line 40: last comment : I'm not sure we need this anymore, since all snippets have the full path of the file as title

mathurinm (approved):

sofianetanji:
small typos:
- In "Describing the solver run procedure", the code snippet for the tolerance bullet point has "def run(self, n_iter):" instead of "def run(self, tol):"
- Callback bullet point: "exposes the intermediate values **of** the iterates." and "Here is a ~~as~~ snippet that illustrate how it could be implemented.

Also, "Objective.evaluate_result" appears many times in the documentation and is also present in the "template_benchmark". Is it the old name of what is now "Objective.compute" ?

mathurinm:
Thanks for the feedback @sofianetanji !
Are you sure you're looking at the latest doc build for this PR? I can't see the 2 points you mentioned in it.
https://output.circle-artifacts.com/output/job/c9eec291-5753-430b-b226-efe54035a8f3/artifacts/0/dev/tutorials/add_solver.html  

To see the latest build on the PR, click the green arrow next to the latest commit, then click on the action `ci/circleci: build_docs artifact`

For compute/evaluate_result, it's the opposite: evaluate_result is the new name (see https://github.com/benchopt/benchopt/pull/576 and the whatsnew)

sofianetanji:
> Thanks for the feedback @sofianetanji ! Are you sure you're looking at the latest doc build for this PR? I can't see the 2 points you mentioned in it. https://output.circle-artifacts.com/output/job/c9eec291-5753-430b-b226-efe54035a8f3/artifacts/0/dev/tutorials/add_solver.html
> 
> To see the latest build on the PR, click the green arrow next to the latest commit, then click on the action `ci/circleci: build_docs artifact`
> 
> For compute/evaluate_result, it's the opposite: evaluate_result is the new name (see #576 and the whatsnew)

My bad ! Just checked the last commit, it's all good !

mathurinm (approved):

doc/tutorials/add_solver.rst
```diff
@@ -0,0 +1,252 @@
...
+        ...
+        sampling_strategy = "callback"
+        ...
+
+        def run(self, callback):
+            X, y = self.X, self.y
+            n_features = self.X.shape[1]
+
+            # init vars
+            beta = np.zeros(n_features)
```
tomMoral, line 191: Need to define `self.beta` for the `callback` to work with `get_result`.

```suggestion
            self.beta = np.zeros(n_features)
```

tomMoral (approved):
LGTM, very nice work to both of you!

Just a small comment on the callback that seems to be not working (missing reg to `self.beta` in the cb.

PR #644, opened 2023-08-21
FIX call `warm_up` when the attribute `_warmup_done` don't exist

### LINKED ISSUE #642: Compilation time in the `warm_up` method
After adding the `warm_up` method from #602 in a solver, it seems that the compilation time is accounted in the results. 
image

MatDag:
This fixes #642.
The `warm_up` method was not called when the attribute `_warmup_done` did not exist.

### DIFF
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -259,7 +259,7 @@ class BaseSolver(ParametrizedNameMixin, DependenciesMixin, ABC):
         ...
 
     def _warm_up(self):
-        if not getattr(self, '_warmup_done', True):
+        if not getattr(self, '_warmup_done', False):
             self.warm_up()
             self._warmup_done = True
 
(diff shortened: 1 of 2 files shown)


benchopt/base.py
```diff
@@ -259,7 +259,7 @@ def warm_up(self):
         ...
 
     def _warm_up(self):
-        if not getattr(self, '_warmup_done', True):
+        if not getattr(self, '_warmup_done', False):
```
Badr-MOUFAD, line 262: It seems to me that this implementation is a bit unclear

WDYT about this suggestion @MatDag?

```suggestion
        
        if getattr(self, '_warmup_done', None):
            # already warmed up
            return
        
        self.warm_up()
        self._warmup_done = True
```

MatDag: Seems clearer indeed

tomMoral (approved):
LGTM

PR #646, opened 2023-08-27
FIX call to warmup and add tests

### LINKED ISSUE #642: Compilation time in the `warm_up` method

tomMoral:
Fix the call to warmup, which was always skipped.
Add non-regression tests.

### Checks before merging PR
- [x] ~added documentation for any new feature~
- [x] added unit test
- [x] ~edited the [what's new](../../whatsnew.rst) (if applicable)~

Fixes #642

### DIFF
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -259,7 +259,7 @@ class BaseSolver(ParametrizedNameMixin, DependenciesMixin, ABC):
         ...
 
     def _warm_up(self):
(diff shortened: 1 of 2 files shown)


tomMoral:
Oh, I missed the fact that #644 was actually fixing the same, I will add the tests on #644, closing this one.

PR #647, opened 2023-09-04
Fix HTML legend that was not displayed on some cases.

Melvin-klein:
Fix legend that was not displayed on same cases. The problem was due to the fact that the legend colours were not reset when the `generate-results` command was run.

It works but I have to check if some code is now outdated.

@tomMoral


tomMoral (approved):
LGTM, let's regenerate the results website with this and see if this fixes everything.


@Melvin-klein could you also look at how we could add a non-regression test for this? (on later PR)

PR #663, opened 2023-09-21
#175 TST add a test that a solver can run with n_iter=0 

### LINKED ISSUE #175: TST add a test that a solver can run with n_iter=0
Some solvers do not support `n_iter=0` or `INFINITY` for solver based on tolerance.
It would be nice to add a test in `test_benchmark` that check this.

JOELPOG:
closes #175  TST add a test that a solver can run with n_iter=0 #175 

added test to `test_benchamarks.py` for some solvers which do not support `n_iter=0` or `INFINITY` for solver based on tolerance.

### Checks before merging PR
- [x] added unit test

### DIFF
--- a/benchopt/tests/test_benchmarks.py
+++ b/benchopt/tests/test_benchmarks.py
@@ -125,19 +125,12 @@ def test_solver_install(test_env_name, benchmark, solver_class, check_test):
     )
 
 
-def test_solver(benchmark, solver_class, check_test):
-    # Check that a solver run with at least one configuration of a simulated
-    # dataset.
-
-    if check_test is not None:
-        check_test(solver_class)
-
+def _test_solver_with_parameters(benchmark, solver_class, objective_config, dataset_config):
     if not solver_class.is_installed():
         pytest.skip("Solver is not installed")
 
-    test_config = getattr(solver_class, "test_config", {})
-    objective_config = test_config.get('objective', {})
-    dataset_config = test_config.get('dataset', {})
+    if solver_class._solver_strategy not in ['tolerance', 'auto']:
+        pytest.skip("Solver is not based on tolerance")
 
     objective_class = benchmark.get_benchmark_objective()
     objective = objective_class.get_instance(**objective_config)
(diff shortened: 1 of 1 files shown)


benchopt/tests/test_benchmarks.py
```diff
@@ -206,3 +206,54 @@ def _test_solver_one_objective(solver, objective):
...
+
+        # Test for n_iter=0
+        with pytest.raises(ValueError) as excinfo:
+            solver.run_once(0)
+        assert "n_iter must be a positive integer" in str(excinfo.value)
+
+        # Test for n_iter=np.inf
+        with pytest.raises(ValueError) as excinfo:
+            solver.run_once(np.inf)
+        assert "n_iter must be a positive integer" in str(excinfo.value)
```
tomMoral, line 259: This is testing the reverse of the expected behavior: the goal of the test would be to check that each solver can run with the initial `stop_val`.
So this would mean that you need to call `solver.run_once(stop_val)` with the right `stop_val` depending on the strategy.

benchopt/tests/test_benchmarks.py
```diff
@@ -206,3 +206,54 @@ def _test_solver_one_objective(solver, objective):
...
+        dataset_test_parameters = [{}]
+    for params in dataset_test_parameters:
+        params.update(dataset_config)
+        dataset = dataset_class.get_instance(**params)
+
+        objective.set_dataset(dataset)
+        solver = solver_class.get_instance()
+        skip = solver._set_objective(objective)
+        if skip:
+            continue
```
tomMoral, line 249: This part of the code is duplicated with the code from `test_solver`, this might be good to either test this in `_test_solver_one_objective` or to make an helper for this pattern of looping on the datasets to find one compatible.

tomMoral (changes requested):
Thanks for the PR, a few comments on the code.
I think this PR does not test the correct behavior so this should be changed.

benchopt/tests/test_benchmarks.py
JOELPOG: i think i misunderstood, it seems like i was checking the solvers do not support `n_iter=0` or  `n_iter=np.inf ` for solvers based on tolerance, but the right way of doing it should be to ensure that each solver can run with an appropriate `n_iter value` depending on the strategy.

```python
# Depending on the solver strategy, choose an appropriate n_iter value.
if solver._solver_strategy in ['iteration', 'callback']:
    stop_val = 5000 if getattr(objective, "is_convex", False) else 2
else:
    stop_val = 1e-10 if getattr(objective, "is_convex", False) else 1e-2

# Test that the solver can run with the chosen n_iter value.
solver.run_once(stop_val)

PR #675, opened 2023-11-10
Boxplots

Melvin-klein:
This PR adds boxplots in HTML and no HTML results. Users can show boxplots with 'iterations' or ('solvers') on the x axis and 'objective column' ('final objective column') or 'time' ('final_time') on the y axis.

I have also refactored a few small parts of the JS and added some behavior (show, hide) selectors when the right parameters are selected.


tomMoral:
Hello! Looks quite nice.
When playing with the interface, here are a few thoughts:

- could we display the mean or median here, and not all quantiles:
  image
- I am not sure I understand what it means when I set `X = iteration`.
- why are the markers black now? Also, the lines do not appear in the legend anymore.
  image
- I don't understand why changing from linear to log changes the plot here:
*log*
image
*linear*
image

benchopt/plotting/html/templates/result.mako.html
```diff
@@ -91,30 +91,44 @@ <h1 class="ml11">
...
-                                   <!-- options will be added dynamically -->
+                                    <option value="Time" class="hidden">Time</option>
+                                    <option value="Iteration" class="hidden">Iteration</option>
+                                    <option value="Solver" class="hidden">Solver</option>
                                 </select>
                             </div>
                         </div>
-                        <div class="flex items-center justify-between mt-8 px-8">
+                        <div id="yaxis-type-form-group" class="mt-8 px-8">
+                            <label for="change_yaxis_type" class="block text-sm font-medium text-white">Y-axis</label>
```
tomMoral, line 115: We could put `runtime` as a value the same form as the `objective` column, that would be clearer I think?

We would put this option only for the box plot?

benchopt/plotting/html/static/result.js
```diff
@@ -543,6 +753,11 @@ const _getScale = (scale) => {
         xaxis: 'log',
         yaxis: 'log',
       };
+    case 'log': // used for boxplot
+      return {
+        xaxis: 'log',
+        yaxis: 'log',
```
tomMoral, line 759: why not keep log-log?

tomMoral:
I will merge this PR as is, we will improve it as we go in follow-up PRs.

PR #676, opened 2023-12-07
ENH add support for fish shell

jolars:
Running shell commands through benchopt doesn't work with the fish shell at the moment.
This commit checks if fish is in use and modifies the call appropriately.

There's also a small change for the input to `subprocess.getstatusoutput`, which was a list but should be a string (according to the docs), but it's working anyway so I guess it's a cosmetic change.

### DIFF
--- a/benchopt/utils/shell_cmd.py
+++ b/benchopt/utils/shell_cmd.py
@@ -42,8 +42,13 @@ def _run_shell(script, raise_on_error=None, capture_stdout=True,
             'return_output=True can only be used with capture_stdout=True'
         )
 
+    is_fish = 'fish' in f"{SHELL}"
+
     # Make sure the script fail at first failure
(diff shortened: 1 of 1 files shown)


tomMoral (approved):
LGTM!

For the `getstatusoutput`, this seems to actually be legacy code now, we should have a follow up PR to use [`subprocess.run`](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module).

tomMoral:
Do you think it would be possible to test different shells in the CI?

jolars:
Hm, I'm not sure. Github actions only support a limited numer of shells through the `shell` parameter: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

Maybe just running `fish <script>` would work, I'm not sure.

PR #682, opened 2024-02-23
FIX pin julia version to <=1.10.0

mathurinm:
Temporary fix as 1.10.1 seems to have broken some downloads functionalities : https://github.com/JuliaLang/julia/issues/53339, causing the tests to fail on `main`: https://github.com/benchopt/benchopt/actions/runs/8018415344/job/21904251404#step:6:1302

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,7 +58,7 @@ jobs:
           command: |
             # remove the cache index
             mamba clean -i -y
-            mamba install python=3.10 julia<=1.10.0  r-base rpy2 \
+            mamba install python=3.10 julia r-base rpy2 \
                           numpy cython -yq
             pip install --upgrade "setuptools<58.5"
             pip install --upgrade --progress-bar off git+https://github.com/scikit-learn-contrib/lightning.git
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,10 +58,9 @@ jobs:
           command: |
             # remove the cache index
             mamba clean -i -y
-            mamba install python=3.10 "julia<=1.10.0" r-base rpy2 \
+            mamba install python=3.10 r-base rpy2 \
                           numpy cython -yq
             pip install --upgrade "setuptools<58.5"
-            pip install --upgrade --progress-bar off julia
             pip install --upgrade --progress-bar off git+https://github.com/scikit-learn-contrib/lightning.git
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -61,7 +61,6 @@ jobs:
             mamba install python=3.10 julia r-base rpy2 \
```
tomMoral, line 61: why remove Julia instead of pinning the version here?

tomMoral: it uses Julia for one of the example

mathurinm: Julia is installed here and 2 lines below too, so I removed the second instance

tomMoral: The second one is for `pyjulia`, which is necessary to get the curve with julia in the example with R and Julia.
(the trick is: `conda install julia` installs the julia interpreter while `pip install julia` install the interface to call julia from python, both are needed).

.circleci/config.yml
```diff
@@ -61,7 +61,6 @@ jobs:
...
                           numpy cython -yq
             pip install --upgrade "setuptools<58.5"
```
tomMoral, line 63:
```suggestion
            pip install --upgrade "setuptools<58.5"
            pip install --upgrade --progress-bar off julia
```

.circleci/config.yml
tomMoral, line 61:
```suggestion
            mamba install python=3.10 "julia<=1.10.0" r-base rpy2 \
```

.circleci/config.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           command: |
             # remove the cache index
             mamba clean -i -y
-            mamba install python=3.10 julia r-base rpy2 \
+            mamba install python=3.10 "julia<=1.10.0" r-base rpy2 \
                           numpy cython -yq
             pip install --upgrade "setuptools<58.5"
             pip install --upgrade --progress-bar off julia
```
tomMoral, line 64: Sorry, my bad, actually the example only compares R and python as the name suggests it...
https://github.com/benchopt/benchopt/blob/main/examples/plot_run_benchmark_python_R.py

So we can remove julia from CircleCI altogether.

```suggestion
            mamba install python=3.10 r-base rpy2 \
                          numpy cython -yq
            pip install --upgrade "setuptools<58.5"
```

PR #685, opened 2024-03-07
[WIP] Make `benchopt.cli`an executable module

rflamary:
This PR aim at alowing calling benchopt from any python interpreter with something like 

```
python -m benchopt.cli
```
Basically we just need to add a file in the cli module

I find this interesting because it allow to run benchopt from different environments without the need to activate them. Similar feature are available for pytest and pip

### Checks before merging PR
- [ ] added unit test

### DIFF
--- a/benchopt/cli/__main__.py
+++ b/benchopt/cli/__main__.py
@@ -1,0 +1,4 @@
+from benchopt.cli import benchopt
+
+if __name__ == '__main__':
+    benchopt()


benchopt/cli/__main__.py
tomMoral, line 1: I would move this file to `benchopt`, so we can call `python -m benchopt` instead, as the `benchopt.cli` is probably not intuitive for users.

tomMoral (changes requested):
Hello, thanks for the PR!

Issue #627, opened 2023-07-18
ENH how to specify path for data dir


tomMoral:
For some `Dataset`, it is not possible to easily download the data from the internet, even though they are open.
This is typically the case for large neuroimaging datasets such as CAMCAN or HCP, which are tens of Tb and are usually downloaded directly on specific system.

One idea would be to be able to set the path of the dataset in the benchmark config file, that would be retrievable though `benchopt.get_setting`?

tomMoral:
After discussion with @Melvin-klein , we converged to the following API:

- Add a `data_home` entry in the config file, default to `benchmark_dir/data`.
- Allow config file to have `data_paths` entry, with a dictionary `key -> path` to set specific path per key.
- Add a helper `get_data_path(key=None)`, that gives the path associated to `key` if it exists in the config file, other wise `$data_home/key`. If key is `None`, this will return `$data_home`.

PR #701  is implementing this behavior.

tomMoral:
This has been implemented in #701

PR #701, opened 2024-04-17
Add "paths" key in available config keys

Melvin-klein:
Allow the key `paths` in benchmark config.


tomMoral (commented):
overall, this looks good but we should add documentation and test for it.

tomMoral:
This would close #627

benchopt/tests/test_benchmark_features.py
```diff
@@ -311,3 +312,45 @@ def get_result(self, **data):
...
         out.check_output(rf"RUNONCE\({n_iter}\)", repetition=1)
+
+
+def test_paths_config_key():
+    with temp_config_file() as config_file:
+        with open(config_file, 'r+') as f:
+            f.write("""
+        paths:
+            data: /path/to/data
+        """)
```
tomMoral, line 323:
```suggestion
                paths:
                    data: /path/to/data
            """)
```

benchopt/tests/test_benchmark_features.py
```diff
@@ -311,3 +312,45 @@ def get_result(self, **data):
...
+            class Dataset(BaseDataset):
+                name = "Simulated"
+                requirements = []
+
+                def get_data(self):
+                    paths = config.get_setting(
+                        "paths",
+                        Path("{config_file}"),
+                        "temp_benchmark"
+                    )
```
tomMoral, line 342: here, we should use `get_data_path`

benchopt/config.py
```diff
@@ -261,6 +293,29 @@ def get_setting(name, config_file=None, benchmark_name=None,
     return value
 
 
+def get_data_path(key: str = None, config_file=None, benchmark_name=None,
+                  default_config=None):
```
tomMoral, line 297: We should only have parameter `key`, the other should be inferred from the benchmark.
To get the current benchmark, you can use the novel `get_running_benchmark` function.

tomMoral (approved):
LGTM!

PR #717, opened 2024-05-23
added Windows compatibility in shell manipulation

wassimmazouz:
- [x] added condition on `sys.platform` to differentiate `cmd` and other shells
- [x] added entry in CI matrix for Windows testing with corresponding `.bat` test script.
- [x] added new `OsSpecificNamedTemporaryFile()` function which returns `NamedTemporaryFile()` in case of Unix usage, otherwise a modified version with additional parameter `delete_on_close=False` to force compatibility with how Windows treats temporary files. (cf. this [documentation](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) for more info). However, this compatibility on Windows requires Python 3.12 or higher to work.
- [x] `julia[version">=1.10.0"]` is unavailable on Windows. The latest `julia` is installed. `version=0.6.2` in this case.
- [x] `^` is added before expanding wildcards in cmd to treat the `*` as a normal character.
- [x] `re.escape()` is used to escape some characters that could be misinterpreted (typically the `\` used in Windows path)
- [x] Normalizing some paths to be cross-platform using `os.path.join()` or `pathlib.Path()`.
- [x] In `stopping_criterion.py`,  `>` changed to `>=` as the `time` module treats time calculations differently on Windows than on Unix.   
- [x] All tests working in CI


benchopt/utils/shell_cmd.py
```diff
@@ -3,14 +3,19 @@
...
+    settings = {
+        'shell': 'bash' if os.name != 'nt' else 'cmd'  # Default to cmd on Windows
+    }
+    return settings.get(setting_name, 'bash')
 
 
 SHELL = get_setting('shell')
 
 
-def _run_shell(script, raise_on_error=None, capture_stdout=True,
```
mathurinm, line 12: @wassimmazouz avoid cosmetic changes like this, they are not related to the changes you introduce but make the diff longer hence harder to review. Your autoformatter is probably the culprit here

benchopt/utils/shell_cmd.py
```diff
@@ -3,7 +3,13 @@
 import subprocess
 
 from ..config import DEBUG
-from ..config import get_setting
+
+
+def get_setting(setting_name):
```
mathurinm, line 8: @wassimmazouz it seems risky to have two functions with the same name; I don't see why we do it this way ?

tomMoral (changes requested):
Hello!

Thanks a lot for looking into this, this looks super good!
To make this process easier, it would be nice to add an entry in the CI matrix to this file to have a test on windows to see how far we are from having something that works on windows.
https://github.com/benchopt/benchopt/blob/main/.github/workflows/test.yml#L30

It is a nice step toward making `benchopt` work on windows but I don't think this will solve all the issues. In particular, an annoying issue is that the python library `conda` does not detect the current conda env, (this does not use any shell call to check this). But this can probably be tackled in a later PR.

tomMoral:
There are some issues with undefined variables on linux in the tests.

wassimmazouz:
@mathurinm Thank you for the insight on the problem. `tempfile` [documentation](https://docs.python.org/3/library/tempfile.html) shows a new parameter `delete_on_close` added in version `3.12`. Using the combination of `delete=True` and `delete_on_close=False` could solve this problem. However, some opens which do not have full access will have to be treated differently as to not raise a `PermissionError`. Also, CI tests will have to be run on `windows-latest, 3.12`. I will look into that promptly.

wassimmazouz:
@mathurinm CI is 🟢. However, 4 tests remain failing for the time being:
- `test_solver_install` : does not work for the `R-PGD` solver as it needs R to be added manually to `Path` (cf. this [link](https://stackoverflow.com/questions/63863449/oserror-cannot-load-library-c-program-files-r-r-4-0-2-bin-x64-r-dll-error-0))
- `test_config_file_permission_no_warning` : displays a warning on the permission of the temporary file. I believe permission modification is more restricted in Windows (cf. this [link](https://stackoverflow.com/questions/72944021/pythons-os-chmod-behaves-differently-on-linux-vs-windows))
- `test_benchopt_install_in_env_with_requirements` : No clear idea yet why this fails. (Here's the [error message](https://github.com/benchopt/benchopt/actions/runs/9679242536/job/26704802386?pr=717#step:10:1201))
- `test_no_error_minimal_requirements` : This only fails when the `--skip-install` parameter is added onto `pytest`. Here also, the reason for which it fails is still unclear. (here's the [error message](https://github.com/benchopt/benchopt/actions/runs/9679505377/job/26705655670?pr=717#step:10:2179))

I believe the first 2 tests should be skipped, as making them work may need some more in-depth modifications of the code and they don't seem vital for the proper functioning of Benchopt. However, I am unsure about the remaining two. I've skipped all four tests for the time being, awaiting your feedback.

benchopt/cli/tests/test_cli.py
```diff
@@ -388,7 +395,7 @@ class Solver(BaseSolver):
                 name = "test_import_ctx"
 
             """)
-        with tempfile.NamedTemporaryFile(
+        with OSSpecificNamedTempFile(
```
mathurinm, line 398: should we call this `OSSpecificNamedTemporaryFile` for consistency?

benchopt/stopping_criterion.py
```diff
@@ -153,7 +154,10 @@ def get_runner_instance(self, max_runs=1, timeout=None, output=None,
 
         # Store running arguments
         if timeout is not None:
-            stopping_criterion._deadline = time.time() + timeout
+            if sys.platform == 'win32':
+                stopping_criterion._deadline = time.monotonic() + timeout
```
mathurinm, line 158: can you add a comment as to why this is needed here?

benchopt/stopping_criterion.py
```diff
@@ -222,10 +226,16 @@ def should_stop(self, stop_val, objective_list):
 
         # check the different conditions:
         #     diverging / timeout / max_runs / stopping_criterion
+        if sys.platform == 'win32':
```
mathurinm, line 229: if it fits on one line: `current_time = a if condition else b`

benchopt/tests/__init__.py
```diff
@@ -11,7 +11,7 @@
 
 
 # Default benchmark
-DUMMY_BENCHMARK_PATH = Path(__file__).parent / 'dummy_benchmark'
+DUMMY_BENCHMARK_PATH = Path(Path(__file__).parent / 'dummy_benchmark')
```
mathurinm, line 14: this did not work on windows?

benchopt/tests/test_benchmark_features.py
```diff
@@ -389,11 +392,10 @@ def get_data(self):
             ], standalone_mode=False)
 
         if test_case == "without_data_home":
-            out.check_output(r"path/to/data", repetition=1)
+            expected_path = re.escape(data_path)
+            out.check_output(expected_path, repetition=1)
```
mathurinm, line 396: cosmetic: you can avoid introducing a new variable and just add the re.escape in checkout output; it will make the diff 1 line  shorter

benchopt/tests/test_benchmarks.py
```diff
@@ -112,6 +113,8 @@ def test_solver_install_api(benchmark, solver_class):
         assert hasattr(solver_class, 'install_script')
 
 
+@pytest.mark.skipif(sys.platform == 'win32',
+                    reason="Skipping test on Windows because of R-PGD")
```
mathurinm, line 117: can you skip only the R-PGD solver instead of skipping the whole test ?

benchopt/tests/test_config.py
```diff
@@ -13,8 +14,15 @@
...
 def temp_config_file(permission='600'):
     permission = int(f"100{permission}", base=8)
-    config_file = Path() / 'test_config_file.yml'
-    config_file.touch(mode=permission, exist_ok=False)
+    config_file = Path(Path() / 'test_config_file.yml')
+    if sys.platform != 'win32':
+        config_file.touch(mode=permission, exist_ok=False)
+    else:
+        config_file.touch(exist_ok=False)
+        if permission == 0o600:
```
mathurinm, line 22: if permission is something else, we put 666 by default, is it what we want?

benchopt/tests/test_stopping_criterion.py
```diff
@@ -95,7 +95,8 @@ def test_diverged(criterion_class, strategy):
 def test_key_to_monitor(criterion_class, strategy):
     "Check that the criterion tracks the right objective key."
     key = 'test'
-    criterion = criterion_class(strategy=strategy, key_to_monitor=key)
+    criterion = criterion_class(strategy=strategy,
```
mathurinm, line 98: revert

benchopt/utils/misc.py
```diff
@@ -84,3 +85,23 @@ def get_benchopt_requirement(pytest=False):
             )
 
     return req, False
+
+
+def OSSpecificNamedTempFile(dir=None, mode='w+b', prefix=None, suffix=None):
```
mathurinm, line 90: can you add the explanation you mad ein the first PR message here, so that future users understand why this is needed?

benchopt/utils/misc.py
```diff
@@ -84,3 +85,23 @@ def get_benchopt_requirement(pytest=False):
...
+    if sys.platform != 'win32':
+        return tempfile.NamedTemporaryFile(dir=dir, mode=mode, prefix=prefix,
+                                           suffix=suffix)
+
+    else:
+        required_version = (3, 12)
+        current_version = sys.version_info
+
+        if current_version < required_version:
+            raise EnvironmentError("Your current Python version is "
```
mathurinm, line 100: you can indent the code by breaking line after the opening parenthesis to have longer lines and thus fewer lines

benchopt/tests/test_config.py
wassimmazouz: Yes I believe changing the `temp_file_config()` function to default on `permission=666` would make the test pass, but I don't know if it's fine to create config files that are writable by others.

The error message for this test is the following : 

```
UserWarning: BenchOpt config file test_config_file.yml is with mode 666.
As it stores sensitive information such as the github token,
it is advised to use mode 600 (user rw only).
```

So I believe `permission=600` is what's recommended.

The test `test_config_file_permission_no_warning` calls `temp_config_file()` with no parameters, so it should pass the default `permission=600`. Yet, it doesn't. One explanation in the forum I linked is that "Windows only support `stat.S_IWRITE` and `stat.S_IREAD`. The other bits are ignored.". So on Windows, even if we set the file to permission 600, it will still set itself to 666 internally.

Either this test should be skipped, or its code should be changed so that the default permission is 666.

wassimmazouz:
@tomMoral Could you take a look at this.
If there's anything to improve, please let me know. But I believe it's good for now.

benchopt/cli/tests/test_cli.py
```diff
@@ -625,13 +635,16 @@ def get_result(self): pass
...
+            with pytest.raises(AttributeError, match=re.escape(match)):
                 with CaptureRunOutput():
                     install([
                         *f'{benchmark.benchmark_dir} -s buggy-class -y '
                         f'--env-name {test_env_name}'.split()
                     ], 'benchopt', standalone_mode=False)
 
+    @pytest.mark.skipif(sys.platform == 'win32',
+                        reason="Skipping test on Windows because "
+                        "of --skip-install")
```
tomMoral, line 647: Not sure I understand why this is needed?

tomMoral (commented):
Some nitpicks changes and some questions on half the PR, I did not have time to finish the other half but I will take a look next week.

benchopt/utils/terminal_output.py
```diff
@@ -9,8 +9,8 @@
 MIN_LINE_LENGTH = 20
 BLACK, RED, GREEN, YELLOW, BLUE, MAGENTA, CYAN, WHITE = range(30, 38)
 
-CROSS = u'\u2717'
-TICK = u'\u2713'
+CROSS = '\u2717'
+TICK = '\u2713'
```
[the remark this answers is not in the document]
wassimmazouz: Tried it [here](https://github.com/benchopt/benchopt/pull/717/commits/dc08a8c32284105a055b3579dc95774f7a891be3). Still shows errors. It seems `'X'` and `'V'` are the only solutions.

benchopt/cli/tests/test_cli.py
wassimmazouz: I skipped this test because I couldn't find any solution to the errors it displayed.

`test_no_error_minimal_requirements`  only fails when the `--skip-install` parameter is added onto `pytest`. 
 (here's the [error message](https://github.com/benchopt/benchopt/actions/runs/9679505377/job/26705655670?pr=717#step:10:2179))

I also couldn't solve the errors of 3 other tests which you'll find I skipped : 

- `test_solver_install` : does not work for the `R-PGD` solver as it needs R to be added manually to `Path` (cf. this [link](https://stackoverflow.com/questions/63863449/oserror-cannot-load-library-c-program-files-r-r-4-0-2-bin-x64-r-dll-error-0))
- `test_config_file_permission_no_warning` : displays a warning on the permission of the temporary file. I believe permission modification is more restricted in Windows (cf. this [link](https://stackoverflow.com/questions/72944021/pythons-os-chmod-behaves-differently-on-linux-vs-windows))
- `test_benchopt_install_in_env_with_requirements` : No clear idea yet why this fails. (Here's the [error message](https://github.com/benchopt/benchopt/actions/runs/9679242536/job/26704802386?pr=717#step:10:1201))

I believe the first 2 tests should be skipped, as making them work may need some more in-depth modifications of the code and they don't seem vital for the proper functioning of Benchopt. However, I am unsure about the remaining two. I've skipped all four tests for the time being, awaiting your feedback.

benchopt/utils/terminal_output.py
```diff
@@ -2,15 +2,16 @@
...
 from ..config import DEBUG
 
 
 MIN_LINE_LENGTH = 20
 BLACK, RED, GREEN, YELLOW, BLUE, MAGENTA, CYAN, WHITE = range(30, 38)
 
-CROSS = u'\u2717'
-TICK = u'\u2713'
+CROSS = '\u2717' if sys.platform != 'win32' else 'X'
+TICK = '\u2713' if sys.platform != 'win32' else 'V'
```
tomMoral, line 14: The support of unicode in the windows shell is not great, and I understand the need to have different cases here.

Just looking at some encoding solutions, maybe a good solution would be to set this:
```python
import sys
sys.stdout.reconfigure(encoding='utf-8')
```

Could you check it?

benchopt/tests/test_config.py
tomMoral: In the config file, we are stocking the GitHub token of the user, which should not be readable by other users.
This is why we are using a file that is only accessible to the user with permission `600` on linux.
I would say it is okay to skip the tests on windows if there is no way to make the file only readable by the user.

But we should put a warning when reading the config file if the `github_token` is set on windows as this is not safe.

benchopt/tests/test_benchmarks.py
```diff
@@ -114,6 +115,19 @@ def test_solver_install_api(benchmark, solver_class):
...
+            benchmark_name = benchmark.name
+        else:
+            benchmark_name = str(benchmark)
+        solver_class_name = solver_class.__module__ + "."
+        solver_class_name += solver_class.__name__
+        if (
+            benchmark_name == "dummy_benchmark" and solver_class_name ==
+            "benchopt_benchmarks.dummy_benchmark.solvers.r_pgd.Solver"
+        ):
+            pytest.skip("Skipping the specific R-PGD solver for Windows")
```
tomMoral, line 130: It would be better if we can set this skip directly in the benchmark: https://github.com/benchopt/benchopt/blob/main/benchopt/tests/dummy_benchmark/test_config.py

benchopt/tests/test_benchmark_features.py
```diff
@@ -351,16 +350,20 @@ def get_result(self, **data):
 
 @pytest.mark.parametrize("test_case", ["without_data_home", "with_data_home"])
 def test_paths_config_key(test_case):
+    data_path = os.path.normpath("/path/to/data")
+    home_data_path = os.path.normpath("/path/to/home_data")
+    relative_data_path = os.path.normpath("path/to/data")
```
tomMoral, line 355: This is not the right test case as the idea with `relative_data_path` is that it should not be absolute (start with `/`).
Putting `normpath` will make it an absolute path.

Maybe keep the original string but compare them with conversion to `Path` objects.

continuous_integration/test_script_windows.bat
```diff
@@ -0,0 +1,39 @@
...
+call conda info
+if %errorlevel% neq 0 exit /b %errorlevel%
+
+call pip freeze
+if %errorlevel% neq 0 exit /b %errorlevel%
+
+call conda activate "%CONDA_ENV%"
+if %errorlevel% neq 0 exit /b %errorlevel%
+
+set TEST_CMD=python -m pytest -vs --durations=20 --junitxml="%JUNIT_XML%"
```
tomMoral, line 14: This line is not useful.
```suggestion
```

continuous_integration/test_script_windows.bat
```diff
@@ -0,0 +1,39 @@
+@echo off
+
+setlocal enabledelayedexpansion
+
+call conda info
+if %errorlevel% neq 0 exit /b %errorlevel%
```
tomMoral, line 6: is this necessary to have this at each line with `@echo off`?

benchopt/cli/tests/test_cli.py
tomMoral: I agree for the first two tests, I tried to push a fix for the second ones. Let's see the debug log :)

tomMoral (commented):
Some more comments but we are close to the end!

continuous_integration/test_script_windows.bat
wassimmazouz: Apparently, it is mandatory to add it at each line to achieve the same result as `set -e` does in `.sh` scripts. [source](https://stackoverflow.com/questions/734598/how-do-i-make-a-batch-file-terminate-upon-encountering-an-error)

benchopt/utils/terminal_output.py
wassimmazouz: It worked ! Thanks @tomMoral

.github/workflows/test.yml
```diff
@@ -60,17 +62,41 @@ jobs:
...
 
+      - name: Install Julia on Windows
+        if: matrix.os == 'windows-latest'
+        run:
+          pip install julia
+
+      - name: Install Julia on Ubuntu/OSX
+        if: matrix.os != 'windows-latest'
+        run:
+          conda install -yq julia[version">=1.10.0"]
```
tomMoral, line 73: These two lines are doing very different things:
1. `conda install julia` install the julia language.
2. `pip install julia` install the pyjulia python package.

It seems weird that this is necessary to specify this to make the test pass?

.github/workflows/test.yml
```diff
@@ -60,17 +62,41 @@ jobs:
...
+
+      - name: Install benchopt and its dependencies Windows
+        if: matrix.os == 'windows-latest'
+        run: |
+          call conda info
+          call conda install -yq pip
+          call pip install -e .[test]
+          call echo using conda
```
tomMoral, line 81: I would not duplicate this step but create a variable `CONDA_CMD` in the `yaml` env section that is `conda` on linux and `call conda` on windows.
You can probably reuse `conda_cmd` from the matrix.

benchopt/config.py
```diff
@@ -14,13 +15,18 @@
...
+    DEFAULT_SHELL = 'bash'
+else:
+    DEFAULT_SHELL = 'cmd.exe'
+
 DEFAULT_GLOBAL_CONFIG = {
     'debug': False,
     'raise_install_error': False,
     'github_token': None,
     'data_dir': './data/',
     'conda_cmd': 'conda',
```
tomMoral, line 28:
```suggestion
    'conda_cmd': 'conda' if sys.platform != 'win32' else 'call conda',
```

benchopt/utils/conda_env_cmd.py
```diff
@@ -1,19 +1,23 @@
...
 
 from ..config import DEBUG
 from ..config import get_setting
 
 SHELL = get_setting('shell')
 CONDA_CMD = get_setting('conda_cmd')
+PIP_CMD = f"{sys.executable} -m pip"
+
+if sys.platform == 'win32':
+    CONDA_CMD = f"CALL {CONDA_CMD}"
```
tomMoral, line 20:
```suggestion
# On windows, calling conda without call exit the cmd script:
# https://github.com/conda/conda/issues/12418
if sys.platform == 'win32' and not CONDA_CMD.lower().startswith('call'):
    CONDA_CMD = f"CALL {CONDA_CMD}"
```

benchopt/utils/conda_env_cmd.py
```diff
@@ -102,9 +106,8 @@ def create_conda_env(
     print(f"Creating conda env '{env_name}':... ", end='', flush=True)
     if DEBUG:
         print(f"\nconda env config:\n{'-' * 40}{benchopt_env}{'-' * 40}")
-    env_yaml = tempfile.NamedTemporaryFile(
-        mode="w+", prefix='conda_env_', suffix='.yml'
-    )
+    env_yaml = NamedTemporaryFile(
+        mode="w+", prefix='conda_env_', suffix='.yml')
```
tomMoral, line 110:
```suggestion
        mode="w+", prefix='conda_env_', suffix='.yml'
    )
```

benchopt/cli/tests/test_cli.py
```diff
@@ -579,22 +581,22 @@ def get_objective(self): pass
         """
 
         # Some solvers are not installable, only keep a simple one.
-        solver = (DUMMY_BENCHMARK_PATH / "solvers" / "python_pgd.py")
+        solver = Path(DUMMY_BENCHMARK_PATH / "solvers" / "python_pgd.py")
```
tomMoral, line 584: Not clear why `Path` is needed?

tomMoral (approved):
Some more comments but we are close to the end!!!

benchopt/utils/terminal_output.py
tomMoral: :tada: !!!

.github/workflows/test.yml
wassimmazouz: `conda install julia` doesn't work on [Windows](https://discourse.julialang.org/t/how-to-request-for-conda-to-support-julia-install-for-windows-os/54092). I tried to find a channel that works but no luck.
I wasn't aware of the differences between the two installs. Do you think there's a workaround, or should we skip all julia-based tests altogether?

.github/workflows/test.yml
wassimmazouz: @tomMoral the last command being written in bash script, I've tried to [condition](https://github.com/benchopt/benchopt/pull/717/commits/cf212a9300cf04c7f76aa509994b2eae617d6837) on the running OS before executing it, but no luck yet. Do you have any solution to that?

Also as you'll probably notice, I've moved the comment outside the run block because `cmd` seems to dislike `#`s.

.github/workflows/test.yml
tomMoral: There is a workaround :) (conda-forge/julia-feedstock#242)
We could try to add `-c -c https://repo.prefix.dev/julia-forge`. I already modified it in the julia helper.

.github/workflows/test.yml
```diff
@@ -60,17 +62,36 @@ jobs:
...
 
+      - name: Install Julia on Windows
+        if: matrix.os == 'windows-latest'
+        run:
+          pip install julia
+
+      - name: Install Julia on Ubuntu/OSX
+        if: matrix.os != 'windows-latest'
+        run:
+          conda install -yq julia[version">=1.10.0"]
```
tomMoral, line 73:
```suggestion

      - name: Install Julia
        run:
          conda install -yq -c https://repo.prefix.dev/julia-forge julia[version">=1.10.0"]
```

benchopt/cli/tests/test_cli.py
```diff
@@ -248,10 +248,11 @@ def test_profiling(self, test_env_name):
                      '-n', '1', '-r', '1', '-o', SELECT_ONE_OBJECTIVE,
                      '--profile', '--no-plot'],
                     'benchopt', standalone_mode=False)
-
+        python_pgd_path = os.path.join(DUMMY_BENCHMARK_PATH, 'solvers',
+                                       'python_pgd.py')
```
tomMoral, line 252: If possible, avoid using `os.path` and use `pathlib` everywhere.
```suggestion
        python_pgd_path = (
            DUMMY_BENCHMARK_PATH / 'solvers' / 'python_pgd.py'
        )
```

benchopt/cli/tests/test_cli.py
```diff
@@ -248,10 +248,11 @@ def test_profiling(self, test_env_name):
...
         out.check_output('using profiling', repetition=1)
         out.check_output(
-            f"File: .*{DUMMY_BENCHMARK_PATH}/solvers/python_pgd.py",
+            f"File: .*{re.escape(python_pgd_path)}",
```
tomMoral, line 255:
```suggestion
            f"File: .*{re.escape(solver)}",
```

benchopt/utils/shell_cmd.py
```diff
@@ -62,10 +73,15 @@ def _run_shell(script, raise_on_error=None, capture_stdout=True,
     if raise_on_error is True:
         raise_on_error = "{output}"
 
+    if IS_CMD:
+        command = f'cmd /c "{tmp.name}"'
+    else:
+        command = f"{SHELL} {tmp.name}"
```
tomMoral, line 79: Can't we use this:
```suggestion
    command = f'{SHELL} "{tmp.name}"'
```

tomMoral (approved):
I would say last round of review, it looks very good :)
A few more comments but I think we are almost there, thanks a lot @wassimmazouz

tomMoral:
Tests are green event with julia!!! :tada:

benchopt/tests/test_runner.py
```diff
@@ -323,8 +324,12 @@ def get_data(self):
     out.check_output("ValueError: Failing solver.", repetition=2)
 
 
+# Under windows, the function needs to be pickleable
+# for parallel jobs to work with joblib
 @pytest.mark.parametrize('n_jobs', [1, 2])
 def test_benchopt_run_script(n_jobs, no_debug_log):
+    if sys.platform == 'win32' and n_jobs > 1:
+        pytest.skip("Parallel jobs are incompatible with Windows")
```
wassimmazouz, line 332: This test, added in the [following commit](https://github.com/benchopt/benchopt/commit/ace41416f1915ab7809c5b19c2f588e939bd6974#diff-cd2463708ecfd4cb8e59743e94eb96708f22b4662a4851c871635ee469223b4d), seems to fail on Windows because it needs [pickleable functions](https://github.com/joblib/joblib/issues/174#issuecomment-65410575). I've skipped it for now.

tomMoral (commented):
A few more comments, probably almost there.

Note that you need to be careful with force push, you erased part of the changes I did before ^^

tomMoral:
IT IS GREEN!!! :green_circle: :green_circle: :green_circle:

tomMoral:
I merged the PR, thanks you so much @wassimmazouz for the hard work!!!!
It is super cool!

PR #728, opened 2024-06-03
MNT Execute only the last push commit in CI

bruAristimunha:
In a PR, each push executes continuous integration tests; however, if you commit and push multiple times within a PR, you create a queue of jobs and testing.

If you have tests that take longer, this ends up generating a long queue to have one test done. This is meaningless, as we always want to test only the last commit.

In this PR, we ensure that CI stops the tests for the previous push and runs the test only for the last one.

### DIFF
--- a/.github/workflows/expose_link.yml
+++ b/.github/workflows/expose_link.yml
@@ -1,4 +1,6 @@
 name: Expose result artifact
+# Cancel in-progress workflows when pushing
+# a new commit on the same branch
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
   cancel-in-progress: true
--- a/.github/workflows/expose_link.yml
+++ b/.github/workflows/expose_link.yml
@@ -2,7 +2,7 @@ name: Expose result artifact
 # Cancel in-progress workflows when pushing
 # a new commit on the same branch
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 on: [status]
 jobs:
--- a/.github/workflows/expose_link.yml
+++ b/.github/workflows/expose_link.yml
@@ -1,10 +1,10 @@
 name: Expose result artifact
+on: [status]
 # Cancel in-progress workflows when pushing
 # a new commit on the same branch
 concurrency:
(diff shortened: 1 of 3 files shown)


.github/workflows/expose_link.yml
```diff
@@ -1,4 +1,7 @@
 name: Expose result artifact
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
```
tomMoral, line 3: Looking at the [doc] (https://docs.github.com/en/actions/using-jobs/using-concurrency), it seems that we can simply put:

```suggestion
  group: ${{ github.workflow }}-${{ github.ref }}
```

.github/workflows/expose_link.yml
```diff
@@ -1,4 +1,7 @@
 name: Expose result artifact
+concurrency:
```
tomMoral, line 2: This section should go bellow the `on` section.


```suggestion
# Cancel in-progress workflows when pushing
# a new commit on the same branch
concurrency:
```

.github/workflows/linting.yml
```diff
@@ -1,5 +1,7 @@
 name: linting
-
+concurrency:
```
tomMoral, line 2: ditto

.github/workflows/linting.yml
```diff
@@ -1,5 +1,7 @@
...
+  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
```
tomMoral, line 3:
```suggestion
# Cancel in-progress workflows when pushing 
# a new commit on the same branch
concurrency: 
  group: ${{ github.workflow }}-${{ github.ref }}
```

.github/workflows/test.yml
```diff
@@ -1,5 +1,7 @@
 name: test
-
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
```
tomMoral, line 3:
```suggestion
# Cancel in-progress workflows when pushing 
# a new commit on the same branch 
concurrency: 
  group: ${{ github.workflow }}-${{ github.ref }}
```

tomMoral (commented):
putting this everywhere would make sense, and moving the block bellow `on`

bruAristimunha:
Solved @tomMoral and @mathurinm

mathurinm (approved):

PR #757, opened 2024-07-17
FIX rename boxplot_chart->boxplot in result.js

tomMoral:
Follow up on #675 

cc @Melvin-klein


tomMoral:
It seems that I cannot see the legend?
image

Issue #747, opened 2024-06-14
Missing error message for negative `objective_value` in logarithmic scale


wassimmazouz:
If `objective_value` is negative, results HTML file should display an error message asking to change the scale option if the user chooses `semilog-y` or `loglog` options.
Right now, it only displays an empty graph:
![image](https://github.com/benchopt/benchopt/assets/165368548/8b3d9930-c314-4aae-b3bd-b3de8a38a020)

PR #765, opened 2024-08-16
Handle negative logarithmic scale

Melvin-klein:
When no solver has at least 3 positive values, we hide the loglog and semilog-y options

#747


benchopt/plotting/html/static/result.js
```diff
@@ -975,6 +979,18 @@ const show = (HTMLElements, style = 'initial') => {
   HTMLElements.forEach(h => h.style.display = style);
 };
 
+const scatterPlotHasMoreThan2PositiveValues = () => {
+  let res = true
+
+  getSolvers().forEach(solver => {
+    let isLogScaleAvailable = data(solver).scatter.is_log_scale_available
+
+    res = res && isLogScaleAvailable
```
tomMoral, line 988: Is it a `and` or a `or`? If you have one curve that is positive, we could want to have a logscale no?

tomMoral (approved):
LGTM, just check on the and/or?

benchopt/plotting/html/static/result.js
etyl: +1, it should be ```or``` and ```res = false```

etyl:
Now the log scale is hidden when there is a negative value (one negative x hides log-x, one negative y hides log-y, either hide loglog), works for all plot types and custom plots.

PR #772, opened 2024-09-11
FIX installing pip dependencies in conda environment

tomMoral:
The current use of `PIP_CMD` which use `sys.executable` does not point toward `python` in the right environment. As using only `pip` can cause issues on windows, I propose to create an environment file with both `conda` and `pip` deps and install everything in one go with [`conda env update`](https://docs.conda.io/projects/conda/en/latest/commands/env/update.html)


tomMoral:
The issue showed up while fixing tests in scikit-adaptation/skada-bench#161

PR #774, opened 2024-09-25
Add energy consumption as an X-axis on display

mickaelsv:
### Checks before merging PR
- [ ] added unit test

---

Added a preview of energy consumption of the job based on a TDP of 15W (and multiplying it by execution time)
It is not a measure but a preview of what could your program consumed during a run 
Also added a .md to explain how to add a measurement of the consumption by using benchopt on a cluster with slurm.

### DIFF
--- a/elec_consumption.md
+++ b/elec_consumption.md
@@ -1,0 +1,9 @@
+Quick explanations on how to add properly electric consumptions for Benchopt : 
+- Modify TDP value with 2 possibilities :  
+  - Ask the user which TDP he wants to use (probably annoying)
+  - Automatic detection of the hardware (CPU or GPU) and get the correct TDP for the calculation
+- In order to make it work with AIPowerMeter (Software made for a Cluster to collect data consumption of slurm jobs), we need to :
+  - Add a new stored data : In callback.py, we need to store the duration of a run AND the date. With this, with a post treatment first, we will be able to fit the consumption with AIPowerMeter measures and to get the correct consumption.
+  - The post treatment program will probably take as input the .parquet file to add/modify a column of data. With this we will be able to print it in benchopt. 
+
+It will be easier to do it with a post treatment since AIPowerMeter record in real time electric consumptions but the script that uses the data to process it run one time per day.


benchopt/plotting/html/static/result.js
```diff
@@ -96,7 +96,7 @@ const getChartData = () => {
     return getScatterCurves();
   } else if (isChart('bar_chart')) {
     return getBarData();
-  } else if (isChart('boxplot')) {
+  } else if (isChart('boxplot_chart')) {
```
tomMoral, line 99: Not sure this changes are intended?

benchopt/plotting/html/static/result.js
```diff
@@ -235,12 +235,16 @@ const getScatterCurves = () => {
     const solverSamplingStrategy = data(solver)['sampling_strategy'];
 
     // plot only solvers that were stopped using xaxis type
-    // plot all solver if xaxis type is `time`
```
tomMoral, line 238: Keep the comment but add "if xaxis type is `time` or `Energy`."

benchopt/plotting/html/static/result.js
```diff
@@ -235,12 +235,16 @@ const getScatterCurves = () => {
...
+    if(xaxisType !== "Time" && xaxisType !== "Energy consumption (Joules)" && solverSamplingStrategy !== xaxisType) {
       return
     }
 
-    const ScatterXaxisProperty = xaxisType === "Time" ? 'x' : 'stop_val';
+    let ScatterXaxisProperty = xaxisType === "Time" || xaxisType === "Energy consumption (Joules)" ? 'x' : 'stop_val';
+    let xValues = data(solver).scatter[ScatterXaxisProperty];
+
+    if (xaxisType === "Energy consumption (Joules)") {
+      xValues = xValues.map(value => value * 15); // 15 is the TDP
```
tomMoral, line 246: This value should be a global constant.

tomMoral (commented):
Thanks for taking a look into this important topic.

I am not in favor of doing this axis transform in the javascript. I would rather add an extra column in the data frame, that we can display in the javascript.
As for how the column is constructed, I am fine with having a GPU/no GPU TDP as a first approximation to scale the time. This should be done both in the `callback.py` and in the `runner.py`.

For more complex workflows based on `AIPowerMeter`, we can iterate in follow up PR.

For the current PR, there seem to be a lot of changes related to the `boxplot` functionality, was there  a bad merge commit?

PR #775, opened 2024-11-08
API update slurm executor to limit the number of parallel jobs.

Jad-yehya:
Using `slurm_array_configuration` in submitit to limit the number of parallel jobs on slurm.

### Checks before merging PR
- [ ] added unit test

### DIFF
--- a/benchopt/utils/slurm_executor.py
+++ b/benchopt/utils/slurm_executor.py
@@ -35,9 +35,16 @@ def get_slurm_executor(benchmark, slurm_config, timeout=100):
         # Timeout is in second in benchopt
         config['slurm_time'] = f"00:{int(1.5*timeout)}"
 
+    # Setting the array parallelism if specified in the config file
+    max_jobs = config.pop('max_jobs', None)
+
     slurm_folder = benchmark.get_slurm_folder()
     executor = submitit.AutoExecutor(slurm_folder)
     executor.update_parameters(**config)
+
+    # Updating the array parallelism parameter
+    if max_jobs is not None:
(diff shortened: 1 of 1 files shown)


tomMoral:
Is it necessary to put this argument? wouldn't putting `slurm_array_parallelism` directly in the config be sufficient?

Jad-yehya:
The code modification is completely unnecessary, because of `executor.update_parameters(**config)` the executor will receive every argument in the config. So one can just specify `slurm_array_parallelism` in the config file.

I will revert the code and change the doc.

doc/user_guide/advanced.rst
```diff
@@ -57,6 +57,11 @@ If ``slurm_time`` is not set in the config file, benchopt uses by default
 the value of ``--timeout`` multiplied by ``1.5`` for each job.
 Note that the logs of each benchmark run can be found in ``./benchopt_run/``.
 
+``slurm_array_parallelism`` can be used to limit the number of simultaneous
+jobs that are run. This is useful when the cluster has a limit on the number of
+simultaneous jobs that can be run. By default, this value is not set, which
+means that no limitation is used.
```
tomMoral, line 63:
```suggestion
``slurm_array_parallelism`` can be used to limit the number of simultaneous
jobs that are run. This is useful when the cluster has no limit on the number of
simultaneous jobs that can be run. By default, this value is not set, leaving it to
the cluster to adjust this.
```

tomMoral (approved):
Small rewording, otherwise LGTM.

tomMoral (approved):

PR #778, opened 2024-12-05
ENH allow objective to produce multiple entries at once

tomMoral:
In some cases, we want to be able to flatten the objective output and produce multiple values in the dataframe at once.
This PR allows `Objective.evaluate_result` to output a dict or a list of dict, that are extended with the meta data and appended in the result dataframe.

### Checks before merging PR
- [x] added unit test

### DIFF
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -348,12 +348,12 @@ class BaseObjective(ParametrizedNameMixin, DependenciesMixin, ABC):
 
     - `evaluate_result(**result)`: evaluate the metrics on the results of a
       solver. Its arguments should correspond to the key of the dictionary
-      returned by `Solver.get_result` and it can return a scalar value or
-      a dictionary.
+      returned by `Solver.get_result` and it can return a scalar value,
+      a dictionary or a list of dictionaries.
       If it returns a dictionary, it should at least contain a key
       `value` associated to a scalar value which will be used to
       detect convergence. With a dictionary, multiple metric values can be
-      stored at once instead of running each separately.
+      stored at once instead of running each separately. With a list of dictionaries, these metrics can be computed on different objects.
 
     - `get_one_result()`: return one result for which the objective can be
       evaluated. This should be a dictionary where the keys correspond to the
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -444,24 +444,31 @@ class BaseObjective(ParametrizedNameMixin, DependenciesMixin, ABC):
         pass
 
     def format_objective_dict(self, objective_dict):
-        """Format that the output of the objective.
+        """Format the output of the objective.
+        
+        This will prefix all keys in the dictionary with `objective_`
+        to make the objective part of the results clear.
 
         Parameters
         ----------
         objective_dict: dict
             The output of the objective function, which should be a dictionary
-            not containing the key 'name'. The keys of the dictionary should
-            will be prefixed with 'objective_'.
+            not containing the key 'name'.
 
         Returns
         -------
         objective_dict : dict
-            The objective formatted for the result.
+            The formatted objective to include in the DataFrame.
         """
 
-        if 'name' in objective_dict:
+        if not isinstance(objective_dict, dict):
             raise ValueError(
-                "objective output cannot be called 'name'."
+                "The output of the objective should be a list of dictionary "
+                "with no 'name' key in it"
+            )
+        elif 'name' in objective_dict:
+            raise ValueError(
+                "objective output cannot contain a key 'name'"
             )
         return {
             f'objective_{k}': v for k, v in objective_dict.items()
--- a/benchopt/base.py
+++ b/benchopt/base.py
@@ -464,9 +464,9 @@ class BaseObjective(ParametrizedNameMixin, DependenciesMixin, ABC):
 
         if not isinstance(objective_dict, dict):
             raise ValueError(
-                "The output of Objective.evaluate_result should be either a single "
-                "dictionary or a list of dictionaries. Note that these dictionaries "
-                "cannot contain a key 'name'"
+                "The output of Objective.evaluate_result should be either a "
+                "single dictionary or a list of dictionaries. Note that these "
+                "dictionaries cannot contain a key 'name'"
             )
         elif 'name' in objective_dict:
             raise ValueError(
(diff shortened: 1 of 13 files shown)


benchopt/base.py
```diff
@@ -456,25 +480,22 @@ def __call__(self, result):
...
+        else:
+            objective_list = objective_output
 
-        # To make the objective part clear in the results, we prefix all
-        # keys with `objective_`.
-        objective_dict = {
-            f'objective_{k}': v for k, v in objective_dict.items()
-        }
+        objective_list = [
+            self.format_objective_dict(obj) for obj in objective_list
```
tomMoral, line 493:
```suggestion
            self.format_objective_dict(d) for d in objective_list
```

benchopt/base.py
```diff
@@ -456,25 +480,22 @@ def __call__(self, result):
...
+        ]
 
-        return objective_dict
+        return objective_list
 
-    # Save the dataset object used to get the objective data so we can avoid
+    # Saved
     # hashing the data directly.
```
tomMoral, line 499:
```suggestion
    # Save the dataset object used to get the objective data so we can avoid
    # hashing the data directly.
```

benchopt/base.py
```diff
@@ -443,6 +443,30 @@ def save_final_results(self, **solver_result):
         """
         pass
 
+    def format_objective_dict(self, objective_dict):
+        """Format that the output of the objective.
```
tomMoral, line 447:
```suggestion
        """Format the output of the objective.
        
        This will prefix all keys in the dictionary with `objective_`
        to make the objective part of the results clear.
```

benchopt/base.py
```diff
@@ -443,6 +443,30 @@ def save_final_results(self, **solver_result):
...
+
+        Parameters
+        ----------
+        objective_dict: dict
+            The output of the objective function, which should be a dictionary
+            not containing the key 'name'. The keys of the dictionary should
+            will be prefixed with 'objective_'.
```
tomMoral, line 454:
```suggestion
            not containing the key 'name'.
```

benchopt/base.py
```diff
@@ -443,6 +443,30 @@ def save_final_results(self, **solver_result):
...
+
+        Returns
+        -------
+        objective_dict : dict
+            The objective formatted for the result.
```
tomMoral, line 459:
```suggestion
            The formatted objective to include in the DataFrame.
```

benchopt/base.py
```diff
@@ -443,6 +443,30 @@ def save_final_results(self, **solver_result):
...
+        """
+
+        if 'name' in objective_dict:
+            raise ValueError(
+                "objective output cannot be called 'name'."
+            )
```
tomMoral, line 465:
```suggestion
        if not isinstance(objective_dict, dict):
            raise ValueError(
                "The output of the objective should be a list of dictionary "
                "with no 'name' key in it"
            )
        elif 'name' in objective_dict:
            raise ValueError(
                "objective output cannot contain a key 'name'"
            )
```

benchopt/base.py
```diff
@@ -420,7 +420,7 @@ def evaluate_result(self, **solver_result):
 
         Returns
         -------
-        objective_value : float or dict {'name': float}
+        objective_value : float or dict {'name': float} or list of dict
```
mathurinm, line 423: btw this may be interpreted as: the dict must have a single key whose name is `'name'`

benchopt/base.py
```diff
@@ -443,6 +443,37 @@ def save_final_results(self, **solver_result):
         """
         pass
 
+    def format_objective_dict(self, objective_dict):
+        """Format the output of the objective.
```
mathurinm, line 447:
```suggestion
        """Format the output of Objective.evaluate_results.
```

benchopt/base.py
```diff
@@ -443,6 +443,37 @@ def save_final_results(self, **solver_result):
...
+
+        Returns
+        -------
+        objective_dict : dict
+            The formatted objective to include in the DataFrame.
+        """
+
+        if not isinstance(objective_dict, dict):
+            raise ValueError(
+                "The output of the objective should be a list of dictionary "
```
mathurinm, line 466: This is a bit misleading as we allow a single float or lists of dicts ; also I'd remove the 'name' key part of the msg as is not related to checking if it's a dict or not

benchopt/tests/test_objective_outputs.py
```diff
@@ -111,11 +146,43 @@ def get_objective(self):
...
         with CaptureRunOutput() as out:
             run([str(benchmark.benchmark_dir),
                  *'-s test-solver -d test-dataset -n 1 -r 1 --no-display'
                  .split()], standalone_mode=False)
 
         assert out.result_files[0].endswith('.csv')
+
+
+def test_objective_multiple_points(no_debug_log):
+    # Check that if Obejctive.evaluate_result returns a list, we get
```
mathurinm, line 160:
```suggestion
    # Check that if Objective.evaluate_result returns a list, we get
```

mathurinm (changes requested):
LGTM, just missing a whatsnew entry and some documentation. 
I feel like the info is somehow scattered between tutorials, template_benchmark and docstrings. 
Is this an advanced enough feature that it should go into https://benchopt.github.io/user_guide/advanced.html ?

benchopt/base.py
mathurinm: Anyways, because of the not isinstance(list, dict) test performed in call and what we do here https://github.com/benchopt/benchopt/pull/778/files#diff-16b4460abb25e65fd02f1ded976a13c1ab21cbf0b69e1eaa586ca38c7dee12daR493  I'm not sure there are many cases we end up with a non dict over there 

The only case I see would be when objective_output is a list of whatever.

benchopt/base.py
tomMoral:
```suggestion
        objective_value : float or dict {str: float} or list of dict
```

benchopt/base.py
tomMoral: Yes, we only end up here if we have an element of the list that is not a dict so we should point that we should either have a dict or a list of dict.

benchopt/base.py
```diff
@@ -443,6 +444,37 @@ def save_final_results(self, **solver_result):
...
+                "with no 'name' key in it"
```
tomMoral, line 468:
```suggestion
                "The output of Objective.evaluate_result should be either a single "
                "dictionary or a list of dictionaries. Note that these dictionaries "
                "cannot contain a key 'name'"
```

mathurinm (approved):

Issue #749, opened 2024-06-18
The --minimal option of the install command does not work


Melvin-klein:
In the Benchmark's function `install_all_requirements()`, we do not check the value of the parameter "minimal".

lionelkusch:
The option was removed in the PR #718 .

PR #786, opened 2025-03-24
FIX ignored --minimal parameter in benchopt install

lionelkusch:
### Checks before merging PR
- [x] added unit test

This is a fix of the issue #749


tomMoral:
Hello!

Thanks for looking into this and for the PR.
Instead of removing `minimal`, I would be more inclined to fix it by changing line 446 to skip the for loop on `solvers` and `datasets` if `minimal` is `True`

benchopt/benchmark.py
```diff
@@ -445,23 +445,24 @@ def install_all_requirements(self, include_solvers, include_datasets,
...
+                # comprehensible error.
+                if missing is not None:
+                    missings.append(missing)
+
+                conda_reqs += reqs
+                shell_install_scripts += scripts
+                post_install_hooks += hooks
+                if len(scripts) > 0 or len(reqs) > 0:
+                    check_installs += [klass]
+            print(colorify(' done', GREEN))
```
tomMoral, line 465:
```suggestion
        print(colorify(' done', GREEN))
```

tomMoral (approved):
LGTM! thanks a lot

 If you have a bit more time, it would be great to add a test, which would check with a temporary benchmark that this works (maybe adding a non-existent deps for the dataset/solver and checking that it runs fine)
But if you do not have time, we can merge as is and I will add a test later on.

benchopt/benchmark.py
```diff
@@ -449,22 +449,24 @@ def install_all_requirements(self, include_solvers, include_datasets,
...
+
+        if not minimal:
+            for klass in to_install:
+                reqs, scripts, hooks, missing = (
+                    klass.collect(env_name=env_name, force=force, gpu=gpu)
+                )
+                # If a class is not importable but has no requirements,
+                # it might be because the requirements are specified
+                # as global ones in the Objective. We keep track of them
+                # to check and raise a comprehensive error after the install
```
tomMoral, line 461:
```suggestion
                # to check and raise a comprehensive error after the install
                # if it is still not importable.
```

tomMoral (approved):
I fixed the test which was not doing the expected check: it was passing without `--minimal`.
A good practice is to check that the test fails if we have the wrong behavior.

I also did a few tweaks but overall, it LGTM.

PR #787, opened 2025-03-24
CI update the github workflow and add dependabot

lionelkusch:
This is an update of the github workflow action and the addition of the dependency bot.

For the Dependabot bot, there will be a need to activate it in the setting of the github (for more information, see this [documentation](https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide#enabling-dependabot-for-your-repository))


.github/workflows/test.yml
```diff
@@ -86,12 +86,6 @@ jobs:
         if: matrix.os == 'windows-latest'
         run: continuous_integration\test_script_windows.bat
 
-      - name: Publish Test Report
```
jolars, line 97: I'm not sure what it does, really, but why did you remove this action in this PR?

lionelkusch: I removed it because it's failing in some PR and nobody fixes it.
Moreover, the version of git action is quite old; version 5 should be used.

This is a reason for some warnings in the summary of the GitHub action:
image

tomMoral: Yes, this is not really used so I guess this would make sense to remove all junit stuff, in particular also update `test_script.sh` to remove their generation.

lionelkusch: If you want to continue to use the junit report, you can use the new feature of CodeCov:
https://docs.codecov.com/docs/test-analytics

.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+# Set update schedule for GitHub Actions
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"
```
tomMoral:
```suggestion
      interval: "monthly"
```

lionelkusch:
> Could you remove junit generation in the `test_script.sh`? (Or let me know if you want me to take over)

I removed it, I hope that I don't forget a place where it was used.

lionelkusch:
@tomMoral The error is independent of this PR. It's a problem with the cache, I created the PR #823 to solving it.

Can you rerun the CircleCI job after removing the cache?

tomMoral (approved):
LGTM! thanks @lionelkusch

PR #790, opened 2025-04-02
Fix result page css

Melvin-klein:
Fix the following css bug of the result web page as shown in the screen below.

Capture d’écran 2025-04-02 à 18 04 52

Removed an old image reached by its URL that is not available anymore and frozen the height of items with the same height as the old image.


tomMoral (approved):
LGTM! thanks @Melvin-klein

Issue #751, opened 2024-07-05
Legend colours are not displaying correctly


wassimmazouz:
The colours in the HTML file are all in black
![image](https://github.com/benchopt/benchopt/assets/165368548/c1c2175f-2cca-4aca-a1d7-4e6432d12daa)

Melvin-klein:
Hello Wassim, can you post the HTML that produces this output ? I can't reproduce it.

wassimmazouz:
Thanks @Melvin-klein for looking into this. I unfortunately don't have the HTML file anymore. Closing the issue as this might've been only a problem from my computer that doesn't occur anymore

wassimmazouz:
Thanks @Melvin-klein  for looking into this. I unfortunately don't have the HTML file anymore. Closing the issue as this might've been only a problem from my computer that doesn't occur anymore

Melvin-klein:
I reopen it because you're not the only to have this issue.

PR #791, opened 2025-04-03
Fix missing colors on benchmark plots

### LINKED ISSUE #785: BUG VIZ marker colors not showing
In recent runs, the markers all appear black:

Image

@Melvin-klein did you also observe this ?

Melvin-klein:
This PR fix #785 and #751.

Capture d’écran 2025-04-03 à 16 18 22

### DIFF
--- a/benchopt/plotting/plot_objective_curve.py
+++ b/benchopt/plotting/plot_objective_curve.py
@@ -164,7 +164,7 @@ def get_solver_style(solver, plotly=True):
     marker = MARKERS[idx % len(MARKERS)]
 
     if plotly:
-        color = tuple(255*x if i != 3 else x for i, x in enumerate(color))
+        color = tuple(255*float(x) if i != 3 else float(x) for i, x in enumerate(color))
         color = f'rgba{color}'
         marker = idx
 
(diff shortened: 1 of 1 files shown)


mathurinm (approved):
Worked like a charm in my previsouly problematic environment

benchopt/plotting/plot_objective_curve.py
```diff
@@ -164,7 +164,7 @@ def get_solver_style(solver, plotly=True):
     marker = MARKERS[idx % len(MARKERS)]
 
     if plotly:
-        color = tuple(255*x if i != 3 else x for i, x in enumerate(color))
+        color = tuple(255*float(x) if i != 3 else float(x) for i, x in enumerate(color))
```
tomMoral, line 167: Wouldn't it make more sense?


```suggestion
        color = tuple(int(255*x) if i != 3 else float(x) for i, x in enumerate(color))
```

tomMoral (commented):
LGTM

Just a question on why would we convert to float then multiply by 255?
Usually the format is int8 no?

tomMoral:
Looking at the plotly doc, it seems to be 3 int + float so I committed the change.
Will merge, thanks @Melvin-klein !

PR #794, opened 2025-04-25
HTML results table

Melvin-klein:
### Checks before merging PR
- [ ] added unit test


tomMoral (approved):
This looks super nice!
Maybe an extra feature that would be cool:
instead of setting how to select the value globally, could we do this per metric?
For instance with a top right button `...`, which open an overlay where for each column, we can select if it is visible, change the precision, and select how to select the value with `min,max,last`?

PR #799, opened 2025-05-06
FIX slow parallel run when large parameter space

tomMoral:
When the parameter space is too large but each run is very fast, we can have a bottleneck in generating the list of parameters.
This PR avoids this by not creating the list but consuming the parameters as they come.


bruAristimunha:
thaaank you @tomMoral 🙏🏽 this was affecting me

PR #806, opened 2025-06-13
FIX cache miss when changing the order of the solvers

### LINKED ISSUE #807: BUG Cache miss when changing the order of the solvers
On latest dev version (1.6.1.dev26), changing the order of a solver in the cli command prevents the use of the cached version. 
This bug wasn't present with the latest release of benchopt on pypi (1.6.0).

MWE: 
on [logreg_l2 benchmark ](https://github.com/benchopt/benchmark_logreg_l2) with a clean cache, run first : 
`benchopt run . -o "L2 Logistic Regression[lmbd=0.1]" -d simulated[n_features=500,n_samples=200] -s sklearn[solver=newton-cg] -s python-gd`
and then 
`benchopt run . -o "L2 Logistic Regression[lmbd=0.1]" -d simulated[n_features=500,n_samples=200] -s python-gd -s sklearn[solver=newton-cg]`

Theoretically, second run should use the cache, but instead it reruns them and the corresponding runs are hashed differently when we check the cache. 

associated test : [https://github.com/benchopt/benchopt/pull/806](https://github.com/benchopt/benchopt/pull/806)

tomMoral:
### Checks before merging PR
- [x] added unit test

Fixes #807

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -64,6 +64,9 @@ FIX
 - Fix ignored ``--minimal`` option in ``benchopt install``.
   By `Lionel Kusch`_ (:gh:`786`)
 
+- Fix cache miss when order of the solver changes.
+  By `Thomas Moreau`_ (:gh:`806`)
+
 .. _changes_1_6:
 
 Version 1.6 - 15/07/2024


benchopt/base.py
```diff
@@ -206,9 +202,6 @@ def run_once(self, stop_val=1):
             the solver on an easy to solve problem.
         """
 
-        if hasattr(self, '_output') and self._output is not None:
-            self._output.progress('caching warmup times.')
```
gilinca, line 210: Best option I see to keep this warmup message : 
put TerminalOutput.progress logic into a function separated from TerminalOutput class, call it here, and keep a TerminalOutput.progress method wrapping the progress function to not break other calls
 
Other considered options : 
just use print_normalize -> code duplication
give output as an attribute of warm_up -> not intuitive
set the output attribute of solver once in run_one_to_cvg -> weird 
separate i_solver from output and ignore it in cache mechanisms -> weird and heavy

benchopt/base.py
```diff
@@ -65,7 +65,7 @@ def _solver_strategy(self):
             or 'iteration'
         )
 
-    def _set_objective(self, objective, output=None):
+    def _set_objective(self, objective):
```
gilinca, line 68: More intuitive, it makes more sense for none of the dataset, the objective or the solver to have the output as attribute as it is already given to run_one_to_cvg

benchopt/tests/test_runner.py
```diff
@@ -480,3 +480,26 @@ def get_result(self): return dict(beta=1)
...
         # error message should be displayed twice
         out.check_output("ValueError: Failing solver.", repetition=3)
+
+    @pytest.mark.parametrize('n_reps', [1, 4])
+    def test_cache_order(self, no_debug_log, n_reps):
+        with temp_benchmark(
+                objective=self.objective, datasets=self.dataset,
+                solvers=[
+                    self.solver,
+                    self.solver.replace("test-solver", "test-solver2")
```
gilinca, line 490: also tested separately, FIX works

benchopt/base.py
tomMoral: I think we can just ignore this for now, it is not so used.

PR #795, opened 2025-04-28
FIX update julia install command in CI

mathurinm:
This should fix failures in other PRs e.g. https://github.com/benchopt/benchopt/actions/runs/14702343568/job/41254194085?pr=778#step:5:1

PR #833, opened 2025-10-01
FIX parse single line in config

Etyl:
Added a check for single lines for certain values when parsing the config file.
See #795

### DIFF
--- a/benchopt/cli/main.py
+++ b/benchopt/cli/main.py
@@ -37,6 +37,10 @@ def _get_run_args(cli_kwargs, config_file_kwargs):
             raise ValueError(
                 f"Invalid config file option {k}. "
                 "See list of valid options with `benchopt run -h`.")
+            
+        # parse if value is on a single line
+        if not isinstance(v, list) and var_name in ["objective", "dataset", "solver"]:
+            v = [v]
 
         # only override CLI variables if they have their default value
         if (ctx.get_parameter_source(var_name) is not None and


benchopt/cli/tests/test_cli.py
```diff
@@ -241,6 +242,21 @@ def test_config_file(self, no_debug_log):
...
+        n_reps = 2
+        
+        config = f"""
+        objective: test-objective
+        dataset: test-dataset
+        solver: test-solver[param1=42]
+        n-repetitions: {n_reps}
+        max-runs: 0
+        """ 
+        self.run_config_file(config, n_reps, no_debug_log)
```
tomMoral: The `run_config` function is making a lot of tests at once.

As this test is supposed to be a non-regression test for the single line config, I would keep the two test separated, with the `test_config_file_single_line` just checking that it runs (so the first part of the test).

From a code-comprehension point of view: `run_config_file` should just run some code, not perform checks. if it performs check, I would name the method `check_run_config` to make it more explicit.

benchopt/cli/tests/test_cli.py
```diff
@@ -229,6 +229,34 @@ def get_result(self): return dict(beta=None)
...
+            def run(self, _): print(f"Solver#RUN#{self.param1}")
+            def get_result(self): return dict(beta=1)
+        """
+
+        with temp_benchmark(config=config, solvers=solver) as bench:
+            run(
+                f"{bench.benchmark_dir} --no-plot --config "
+                f"{bench.benchmark_dir / 'config.yml'}".split(),
+                'benchopt', standalone_mode=False
+            )
```
tomMoral: Maybe still check that the output contains `test-solver[param1=42]`?

tomMoral (approved):
LGTM, thx @Etyl !

PR #840, opened 2025-10-08
FIX silent param override

tomMoral:
When a base class has a `__init__` method which sets some parameters that are also present in `parameters`, the value set in `__init__` is silently overwritten by the parameter value.
This PR raises an exception in this situation.

### Checks before merging PR
- [ ] added unit test


benchopt/utils/parametrized_name_mixin.py
```diff
@@ -19,6 +19,12 @@ def save_parameters(self, **parameters):
         for k, v in _parameters.items():
             if not hasattr(self, k):
                 setattr(self, k, v)
+            elif getattr(self, k) != v:
+                v_set = getattr(self, k)
+                raise ValueError(
+                    f"Parameter {k} set in {self} `__init__` to {v_set}, "
+                    f"while the parameter value is {v}."
```
felixdivo, line 26:
```suggestion
                    f"Parameter {k} set in {self} `__init__` to {repr(v_set)}, "
                    f"while the parameter value is {repr(v)}."
```

PR #842, opened 2025-10-09
ENH custom plots

etyl:
This PR allows the possibility to create custom plots via the BasePlot class, for example:

```python
class Plot(BasePlot):
    name = "Custom plot 1"
    type = "scatter"
    dropdown = {
        "dataset": ...,  # Will fetch the dataset names from the df
        "color": ["blue", "red"]
    }

    def plot(self, df, dataset, color):
        df = df[(df['data_name'] == dataset)]
        return [
            {
                "x": df["time"].values.tolist(),
                "y": (
                    df[(df['solver_name'] == solver)]
                    ["objective_value"].values.tolist()),
                "color": color,
                "marker": "x",
                "label": solver,
            }
            for solver in df['solver_name'].unique()
        ]

    def get_metadata(self, df, dataset, color):
        df = df[(df['data_name'] == dataset)]
        title = f"Custom Plot 1\nData: {dataset}\nColor: {color}"
        return {
            "title": title,
            "xlabel": "Custom X-axis",
            "ylabel": "Custom Y-axis",
        }

```
This will add new selectors in the html render, which will be shown only when the custom plot kind is selected. Is is possible to add as many parameters as needed, all the plots will be precomputed for all the combinations of the parameters.

### Checks before merging PR
- [x] added unit test

### DIFF
--- a/benchopt/plotting/generate_html.py
+++ b/benchopt/plotting/generate_html.py
@@ -122,7 +122,7 @@ def get_results(fnames, html_root, benchmark, config=None, copy=False):
         # JSON
         # TODO remove json in the future when all plots use custom_data
         result['json'] = json.dumps(shape_datasets_for_html(df))
-        result['json_custom_plots'] = custom_data
+        result['json_custom_plots'] = json.dumps(custom_data)
 
         results.append(result)
 


benchopt/base.py
```diff
@@ -592,3 +594,96 @@ def repeat():
         cv_fold = next(self._cv)
         split_ = getattr(self, "split", self._default_split)
         return split_(cv_fold, *arrays)
+
+
+class BasePlot(ParametrizedNameMixin, DependenciesMixin, ABC):
```
tomMoral: Maybe move to `plotting` as this is not directly the core of `benchopt`?

examples/minimal_benchmark/plots/custom_plot.py
```diff
@@ -0,0 +1,31 @@
+from benchopt import BasePlot
+
+
+class Plot(BasePlot):
+    name = "Custom plot 1"
+    type = "scatter"
+    params = {
```
tomMoral: Call this `dropdown`?

benchopt/base.py
```diff
@@ -592,3 +594,96 @@ def repeat():
...
+        if not hasattr(self, 'type'):
+            raise ValueError("Plot should have a `type` attribute.")
+        supported_types = ['scatter']
+        if self.type not in supported_types:
+            raise ValueError(
+                f"Plot type should be one of {' '.join(supported_types)}. "
+                f"Got {self.type}."
+            )
+
+    def check_params(self):
```
tomMoral: I think using `dropdown` would be more explicit here, as this is what is gonna be used to create dropdown?
```suggestion
    def check_dropdown(self):
```

benchopt/benchmark.py
```diff
@@ -119,6 +119,9 @@ def __init__(
         # replace dots to avoid issues with `with_suffix``
         self.name = self.name.replace('.', '-')
 
+        # Check format of plots
+        self.check_custom_plots()
```
tomMoral: I would not raise an error at `__init__` if a custom plot is broken, only check it when it is used.

benchopt/base.py
tomMoral: Add a `get_style(self, label)` which proposes a dictionary with marker, color, that is unique for this label.
It if is called multiple times for different custom plot, reuse the same style for a given label.

benchopt/base.py
```diff
@@ -592,3 +594,96 @@ def repeat():
...
+    _base_class_name = 'Plot'
+
+    @abstractmethod
+    def plot(self, df, **kwargs):
+        ...
+
+    def get_name(self):
+        return self.name.replace(" ", "_")
+
+    def check(self):
```
tomMoral: Make private:

```suggestion
    def _check(self):
```

benchopt/base.py
```diff
@@ -592,3 +594,96 @@ def repeat():
...
+                if key == "dataset":
+                    values.append(df['data_name'].unique().tolist())
+                elif key == "solver":
+                    values.append(df['solver_name'].unique().tolist())
+                else:
+                    values.append(df[key].unique().tolist())
+            else:
+                values.append(self.params[key])
+
+        combinations = [
```
tomMoral: potentially use `product_params` from `benchopt.utils.parametrized ...`

benchopt/base.py
```diff
@@ -592,3 +594,96 @@ def repeat():
...
+        for key in keys:
+            if self.params[key] is Ellipsis:
+                if key == "dataset":
+                    values.append(df['data_name'].unique().tolist())
+                elif key == "solver":
+                    values.append(df['solver_name'].unique().tolist())
+                else:
+                    values.append(df[key].unique().tolist())
+            else:
+                values.append(self.params[key])
```
tomMoral:
```suggestion
        params = {**self.params}
        for k, v in params.items():
            if v is Ellipsis:
               params[k] = ...
```

benchopt/base.py
```diff
@@ -592,3 +594,96 @@ def repeat():
...
+
+        combinations = [
+            dict(zip(keys, v))
+            for v in itertools.product(*values)
+        ]
+
+        plots = []
+        for kwargs in combinations:
+            data = {}
```
tomMoral:
```suggestion
            data = self.export_meta_data()
```

benchopt/base.py
```diff
@@ -592,3 +594,96 @@ def repeat():
...
+            data["data"] = self.plot(df, **kwargs)
+            data["type"] = self.type
+            data["title"] = self.title
+            data["xlabel"] = self.xlabel
+            data["ylabel"] = self.ylabel
+            for key in kwargs:
+                data[self.get_name() + '_' + key] = kwargs[key]
```
tomMoral: Is this needed?

tomMoral: Maybe use a hashing of the kwargs to make it easier to find which data to use.

benchopt/base.py
```diff
@@ -592,3 +594,96 @@ def repeat():
...
+            plots.append(data)
+
+        return plots
```
tomMoral: you could return dropdown here.

benchopt/__init__.py
```diff
@@ -4,6 +4,7 @@
 from .base import BaseSolver
 from .base import BaseDataset
 from .base import BaseObjective
+from .plotting.base_plots import BasePlot
```
tomMoral:
```suggestion
from .plotting.base import BasePlot
```

benchopt/benchmark.py
```diff
@@ -197,6 +197,32 @@ def check_dataset_patterns(self, dataset_patterns, class_only=False):
             class_only=class_only
         )
 
+    def get_custom_plots(self):
```
tomMoral, line 200: Missing doc strings.

benchopt/benchmark.py
```diff
@@ -197,6 +197,32 @@ def check_dataset_patterns(self, dataset_patterns, class_only=False):
...
+    def get_custom_plot_names(self):
+        return [
+            plot._get_name() for plot in self.get_custom_plots()
+        ]
+
+    def check_custom_plots(self):
+        for plot in self.get_custom_plots():
+            plot._check()
+
+    def get_plt_plots(self, df, output_dir, kind=None):
```
tomMoral: `plt` is kind of ambiguous. Maybe use `mpl` or even better `matplotlib`?
```suggestion
    def get_matpotlib_plots(self, df, output_dir, kind=None):
```

also, not sure why you need this here?

benchopt/cli/tests/test_cli.py
```diff
@@ -482,11 +482,45 @@ def test_complete_datasets(self, dataset_completion_cases):  # noqa: F811
 
 
 class TestPlotCmd:
+    custom_plot = """from benchopt import BasePlot
+
+class Plot(BasePlot):
```
tomMoral: You can indent here and use `inspect.cleandoc` to get proper file.

benchopt/cli/tests/test_cli.py
```diff
@@ -482,11 +482,45 @@ def test_complete_datasets(self, dataset_completion_cases):  # noqa: F811
...
+            "ylabel": "Custom Y-axis",
+        }
+
+    """
 
     @classmethod
     def setup_class(cls):
         "Make sure at least one result file is available"
-        cls.ctx = temp_benchmark()
+        cls.ctx = temp_benchmark(plot=cls.custom_plot)
```
tomMoral:
```suggestion
        cls.ctx = temp_benchmark(plots=cls.custom_plot)
```

benchopt/plotting/__init__.py
```diff
@@ -56,6 +53,12 @@ def plot_benchmark(fname, benchmark, kinds=None, display=True, plotly=False,
     if kinds is not None and len(kinds) > 0:
         config["plots"] = kinds
```
tomMoral, line 55: Don't you need to handle the case where `kinds` is `None`, as defautl config is also `None`?

benchopt/plotting/__init__.py
```diff
@@ -72,7 +75,21 @@ def plot_benchmark(fname, benchmark, kinds=None, display=True, plotly=False,
...
+            if kind in benchmark.get_custom_plot_names():
+                kind_figs = benchmark.get_plt_plots(df, output_dir, kind=kind)
+                figs.extend(kind_figs)
+            elif kind not in PLOT_KINDS:
+                valid_kinds = (
+                    list(PLOT_KINDS.keys()) +
+                    benchmark.get_custom_plot_names()
+                )
+                raise ValueError(f"Unknown plot kind: {kind}. Should be "
+                                 f"one of {valid_kinds}")
```
tomMoral: Why no do `benchmark.get_plt_plot(df, output_dir, kind=config['plots'])`?
Here, you loop multiple times on the same list of names.

benchopt/plotting/__init__.py
```diff
@@ -85,10 +102,7 @@ def plot_benchmark(fname, benchmark, kinds=None, display=True, plotly=False,
                         config["plots"], obj_cols
                 ):
                     if kind not in PLOT_KINDS:
-                        raise ValueError(
-                            f"Requesting invalid plot '{kind}'."
-                            f"Should be in:\n{PLOT_KINDS}"
-                        )
+                        continue
```
tomMoral, line 110: This case has already been handled (raised an error) on l.90 so not sure it is still necessary?

benchopt/plotting/base_plots.py
```diff
@@ -0,0 +1,226 @@
...
+                    f"The values of dropdown should be non empty. "
+                    f"Got {values} for key {key}."
+                )
+
+        keys = set(self.dropdown.keys())
+        plot_kwargs = set([
+            name for name, _ in
+            inspect.signature(self.plot).parameters.items()
+        ])
+        plot_kwargs.remove('df')
```
tomMoral:
```suggestion
        plot_kwargs = set([
            name for name in inspect.signature(self.plot).parameters
            if name != 'df'
        ])
```

benchopt/plotting/base_plots.py
```diff
@@ -0,0 +1,226 @@
...
+
+        return plots, dropdown
+
+    def _get_plt_plot(self, df, output_dir):
+        if self.type == "scatter":
+            return self._get_plt_plot_scatter(df, output_dir)
+        else:
+            raise NotImplementedError(
+                f"Plot type {self.type} not implemented for matplotlib."
+            )
```
tomMoral: I would not put this in this base class.
I think we should separate plot data generation (given in this class), from rendering  (matplotlib or HTML).
So this would go in  a separate `generate_matplotlib.py` file?

benchopt/plotting/base_plots.py
```diff
@@ -0,0 +1,226 @@
...
+            else:
+                save_name = save_name.with_suffix('.pdf')
+                plt.savefig(save_name)
+            print(f'Save {key} as: {save_name}')
+            figs.append(fig)
+
+        return figs
+
+
+class ObjectiveCurvePlot(BasePlot):
```
tomMoral: Would move to a separate file too, maybe `basic_plots`?

benchopt/plotting/base_plots.py
```diff
@@ -0,0 +1,226 @@
...
+
+        plots = []
+        for solver in df['solver_name'].unique():
+            y = (
+                df[(df['solver_name'] == solver)][objective_column]
+                .values.tolist()
+            )
+            x = df[(df['solver_name'] == solver)]["time"].values.tolist()
+            if X_axis == "Iteration":
+                x = list(range(len(x)))
```
tomMoral: Is it what is done now? I think usually, iteration are based on `stop_val` in the dataframe?

benchopt/plotting/base_plots.py
```diff
@@ -0,0 +1,226 @@
...
+        df = df[df['data_name'] == dataset]
+        df = df[df['objective_name'] == objective]
+
+        plots = []
+        for solver in df['solver_name'].unique():
+            y = (
+                df[(df['solver_name'] == solver)][objective_column]
+                .values.tolist()
+            )
+            x = df[(df['solver_name'] == solver)]["time"].values.tolist()
```
tomMoral: `Median/q1/q9` are missing?

benchopt/plotting/generate_html.py
```diff
@@ -108,12 +108,16 @@ def get_results(fnames, html_root, benchmark, config=None, copy=False):
             objective_names=df['objective_name'].unique(),
             obj_cols=[k for k in df.columns if k.startswith('objective_')
                       and k != 'objective_name'],
-            kinds=config_.get('plots', list(PLOT_KINDS)),
+            kinds=config_.get(
+                'plots',
+                list(PLOT_KINDS) + benchmark.get_custom_plot_names()),
             metadata=get_metadata(df, config_.get('plot_configs', {})),
```
tomMoral: nitpick
```suggestion
                list(PLOT_KINDS) + benchmark.get_custom_plot_names()
            ), metadata=get_metadata(df, config_.get('plot_configs', {})),
```

benchopt/plotting/generate_html.py
```diff
@@ -163,6 +167,18 @@ def get_metadata(df, plot_configs):
...
+def get_custom_plots_for_html(df, benchmark):
+    benchmark.check_custom_plots()
+    custom_data = {}
+    custom_dropdown = {}
+    for plot in benchmark.get_custom_plots():
+        data, dropdown = plot._get_all_plots(df)
+        custom_data[plot._get_name()] = data
+        custom_dropdown[plot._get_name()] = dropdown
+
+    return custom_data, custom_dropdown
```
tomMoral: Would move this to `Benchmark.get_all_data_plots`, so here we only handle rendering the data?

benchopt/plotting/base.py
```diff
@@ -0,0 +1,226 @@
...
+            color = tuple(
+                int(255*x) if i != 3 else float(x)
+                for i, x in enumerate(color)
+            )
+            color = f'rgba{color}'
+            marker = idx
+        else:
+            marker = MARKERS[idx % len(MARKERS)]
+
+        return color, marker
```
tomMoral: duplicated from `helpers`?
Also, how to handle plotly vs matplotlib if we separate the render from data production?
I think the l.33-41 need to be handled in each render backend.

benchopt/plotting/html/templates/result.mako.html
```diff
@@ -34,8 +34,18 @@ <h1 class="ml11">
                             </span>
                         </h1>
                     </div>
+                    <div class="mt-8 px-8">
+                        <label for="plot_kind" class="block text-sm font-medium text-white">Chart type</label>
+                        <div class="mt-1 rounded-md shadow-sm">
+                            <select id="plot_kind" class="focus:ring-blue-500 focus:border-blue-500 block w-full h-8 px-4 sm:text-sm border-gray-300 rounded-md" onchange="setState({plot_kind: this.value, scale: (this.value === 'boxplot' || state().plot_kind === 'boxplot' ? 'linear' : state().scale)})">
```
tomMoral: not sure why you need both `this.value and state().plot_kind` for `linear`.
Wouldn't `this.value` be sufficient as this is the new value you will get?

benchopt/plotting/html/templates/result.mako.html
```diff
@@ -127,6 +127,36 @@ <h1 class="ml11">
...
+                                        <select id="change_${custom_kind}_${custom_param}" class="focus:ring-blue-500 focus:border-blue-500 block w-full h-8 px-4 sm:text-sm border-gray-300 rounded-md" onchange="setState({${custom_kind}_${custom_param}: this.value})">
+                                            % for option in result["custom_plot_params"][custom_kind][custom_param]:
+                                                <option value="${option}">${option}</option>
+                                            % endfor
+                                        </select>
+                                    </div>
+                                </div>
+                            % endfor
+                            </div>
+                        % endfor
```
tomMoral, line 159: Not a big fan of generating all possible divs before hand but I guess this makes the `js` much simpler?

tomMoral: My main issue with this is that if 2 plot kinds share the same params, changing plot kind will reset selection?

benchopt/tests/fixtures.py
```diff
@@ -43,6 +43,9 @@ def pytest_configure(config):
     """Setup pytest for benchopt testing"""
     global _TEST_BENCHMARK
 
+    import matplotlib
+    matplotlib.use("Agg")
```
tomMoral, line 47: Great idea! :)

benchopt/tests/test_benchmark_features.py
```diff
@@ -559,3 +559,112 @@ def run(self, n_iter):
 
         # Make sure warmup is called exactly once
         out.check_output("3 passed, 1 skipped, 5 deselected", repetition=1)
+
+
+def test_custom_plot(no_debug_log):
```
tomMoral: Would move in `benchopt/plotting/tests/test_custom_plot`

benchopt/utils/temp_benchmark.py
```diff
@@ -49,8 +49,8 @@ def get_result(self): return dict(beta=None)
 
 @contextlib.contextmanager
 def temp_benchmark(
-        objective=None, datasets=None, solvers=None, config=None,
-        benchmark_utils=None, extra_files=None
+        objective=None, datasets=None, solvers=None, plot=None,
```
tomMoral:
```suggestion
        objective=None, datasets=None, solvers=None, plots=None,
```

benchopt/utils/temp_benchmark.py
```diff
@@ -65,6 +65,8 @@ def temp_benchmark(
     solvers: str | list of str | dict of str | None (default=None)
         Content of the solver.py file(s). If None, defaults to
         ``DEFAULT_SOLVERS``.
+    plot: str | list of str | dict of str | None (default=None)
```
tomMoral:
```suggestion
    plots: str | list of str | dict of str | None (default=None)
```

benchopt/utils/temp_benchmark.py
```diff
@@ -87,6 +89,8 @@ def temp_benchmark(
         datasets = [datasets]
     if isinstance(solvers, str):
         solvers = [solvers]
+    if isinstance(plot, str):
+        plot = [plot]
```
tomMoral:
```suggestion
    if isinstance(plots, str):
        plots = [plots]
```

benchopt/utils/temp_benchmark.py
```diff
@@ -98,6 +102,9 @@ def temp_benchmark(
     else:
         datasets = {**DEFAULT_DATASETS, **datasets}
 
+    if isinstance(plot, list):
+        plot = {f"plot_{idx}.py": p for idx, p in enumerate(plot)}
```
tomMoral:
```suggestion
        plots = {f"plot_{idx}.py": p for idx, p in enumerate(plots)}
```

benchopt/utils/temp_benchmark.py
```diff
@@ -98,6 +102,9 @@ def temp_benchmark(
     else:
         datasets = {**DEFAULT_DATASETS, **datasets}
 
+    if isinstance(plot, list):
```
tomMoral:
```suggestion
    if isinstance(plots, list):
```

benchopt/utils/temp_benchmark.py
```diff
@@ -119,6 +127,11 @@ def temp_benchmark(
             fname = temp_path / "datasets" / fname
             fname.write_text(inspect.cleandoc(content), encoding='utf-8')
 
+        if plot is not None:
+            for fname, content in plot.items():
+                fname = temp_path / "plots" / fname
+                fname.write_text(inspect.cleandoc(content), encoding='utf-8')
```
tomMoral:
```suggestion
        if plots is not None:
            for fname, content in plots.items():
                fname = temp_path / "plots" / fname
                fname.write_text(inspect.cleandoc(content), encoding='utf-8')
```

benchopt/utils/tests/test_parquet.py
```diff
@@ -85,11 +82,7 @@ def test_metadata_saving():
             run(run_cmd, 'benchopt', standalone_mode=False)
 
         config = get_metadata(Path(out.result_files[0]))
-        expected_config = {
-            k: v for k, v in DEFAULT_BENCHMARK_CONFIG.items()
-            if k in ['plots', 'plot_configs']
-        }
-        assert config == expected_config
+        assert config == {'plot_configs': {}}
```
tomMoral, line 85: Might need to check it matches with custom plots?

tomMoral: Oh, this will actually make our life harder for generating the results website (https://github.com/benchopt/results) as we will need to keep the custom plot on the metadata for the benchmark...

Need to think about this for later.

doc/benchmark_workflow/write_benchmark.rst
```diff
@@ -148,3 +150,34 @@ Example
   once during the benchmark.
 
 
+4. Plots
+--------
+
+Warning: this is still an experimental feature and the API may change in future releases (and may also contain bugs).
```
tomMoral:
```suggestion
.. warning::

   This feature is experimental and the API may change in future releases without further notice.
```

doc/benchmark_workflow/write_benchmark.rst
```diff
@@ -148,3 +150,34 @@ Example
   once during the benchmark.
 
 
+4. Plots
```
tomMoral, line 155:
```suggestion
.. _custom_plots:

4. Plots
```

doc/whats_new.rst
```diff
@@ -31,6 +31,10 @@ CLI
 API
 ---
 
+- Add ``BasePlot`` class to allow custom plotting functions
+  in benchmarks.
+  By `Hippolyte Verninas`_ (:gh:`842`)
+
```
```suggestion
- Add the possibility of creating custom plots for each benchmark.
  See `custom_plots`_ for the documentation. 
  By `Hippolyte Verninas`_ (:gh:`842`)

```

benchopt/plotting/html/static/result.js
```diff
@@ -219,83 +217,89 @@ const getIterationBoxplotData = () => {
...
-
-    // plot only solvers that were stopped using xaxis type
-    // plot all solver if xaxis type is `time`
-    if(xaxisType !== "Time" && solverSamplingStrategy !== xaxisType) {
-      return
+  let min_y = Infinity;
+  getCustomPlotData().data.forEach(curveData => {
+    min_y = Math.min(min_y, ...curveData.y);
+  });
+  getCustomPlotData().data.forEach(curveData => {
```
tomMoral, line 243: Why do you need two loops?
```suggestion
  getCustomPlotData().data.forEach(curveData => {
    min_y = Math.min(min_y, ...curveData.y);
```

tomMoral (commented):
A first batch of comments.

Overall, this looks super nice, thanks @Etyl  :)
Many comments are about the form of the code, and to try to simplify it as much as possible.

benchopt/benchmark.py
etyl: I thought it was better to put the plotting function here since it just aggregates information from the custom plots, but it could be somewhere else

benchopt/plotting/html/static/result.js
etyl: To get the global minimum over all curves

benchopt/plotting/html/templates/result.mako.html
etyl: yes that's right, the dropdowns are saved for each plot type. We could maybe change this in the future to generate the dropdowns on the from the js. It would probably be easier once all the plots have been converted to custom plots

benchopt/plotting/__init__.py
etyl: Actually, if config["plots"] is None, it's not registered in the config when initializing the dict, so this case is handled by l.56

etyl: I will change this to make it more clear

benchopt/plotting/__init__.py
etyl: because some of the default plots (bar chart + box plot) are not custom plots, so we need to make the distinction. This will not be an issue when all plots will be converted to custom plots

benchopt/plotting/__init__.py
etyl: This means that this plot not a default plot, so it must not be handled by this function. This loop will be removed in the future

benchopt/plotting/base.py
etyl: The one from helpers is for the boxplot and barchart (will be removed in the future)
Yeah I agree for handling the colors and markers, I'm going to change this

benchopt/utils/tests/test_parquet.py
etyl: The issue here is that the default config has plots:None. It would be best to save the plots in the metadata indeed

benchopt/plotting/html/static/result.js
tomMoral: Indeed, maybe add a comment

benchopt/plotting/base.py
tomMoral: Oh ok!

benchopt/plotting/html/templates/result.mako.html
etyl: This was the intended behavior, if the scale is log and the user switches to boxplot then the scale should become linear (since the user cannot modify the scale for the box plot)

tomMoral: Yes I get this but I don't understand why it is not `this.value === 'boxplot' ? 'linear' : state().scale)`

etyl: ah yes, it's a typo, should be boxplot and barchart I guess

benchopt/benchmark.py
```diff
@@ -5,6 +5,7 @@
 import importlib
 import itertools
 from pathlib import Path
+import matplotlib.pyplot as plt
```
tomMoral: If possible, let's avoid this import.
Importing matplotlib is slow (almost 200ms), and this slows down completion. (I know it is a bit extreme but for completion, it is always chasing after any unecessary imports).

Better to import it only in the function when it is needed.

benchopt/benchmark.py
```diff
@@ -197,6 +200,54 @@ def check_dataset_patterns(self, dataset_patterns, class_only=False):
...
+                            for i, x in enumerate(color)
+                        )
+                        data[key]["data"][idx]["color"] = f"rgba{color}"
+                    else:
+                        marker = data[key]["data"][idx]["marker"]
+                        marker = MARKERS[marker % len(MARKERS)]
+                        data[key]["data"][idx]["marker"] = marker
+
+            custom_data[plot._get_name()] = data
+            custom_dropdown[plot._get_name()] = dropdown
```
tomMoral:
```suggestion
            plot_name = plt._get_name()
            data, dropdown = plot._get_all_plots(df)
            for key in data:
                for idx in range(len(data[key]["data"])):
                    if plotly_style:
                        color = data[key]["data"][idx]["color"]
                        color = tuple(
                            int(255*x) if i != 3 else float(x)
                            for i, x in enumerate(color)
                        )
                        data[key]["data"][idx]["color"] = f"rgba{color}"
                    else:
                        marker = data[key]["data"][idx]["marker"]
                        marker = MARKERS[marker % len(MARKERS)]
                        data[key]["data"][idx]["marker"] = marker

            custom_data[plot_name] = data
            custom_dropdown[plot_name] = dropdown
```

benchopt/cli/tests/test_cli.py
```diff
@@ -482,11 +482,48 @@ def test_complete_datasets(self, dataset_completion_cases):  # noqa: F811
...
+                        "x": (
+                            df[(df['solver_name'] == solver)]
+                            ["time"].values.tolist()
+                        ),
+                        "y": (
+                            df[(df['solver_name'] == solver)]
+                            ["objective_value"].values.tolist()
+                        ),
+                        "color": [0,0,0,1],
+                        "marker": 0,
```
tomMoral, line 508: I would like it to work with this kind of setting if possible
```suggestion
                        "color": "blue",
                        "marker": "o",
```

benchopt/cli/tests/test_cli.py
```diff
@@ -482,11 +482,48 @@ def test_complete_datasets(self, dataset_completion_cases):  # noqa: F811
...
+                        "label": solver,
+                    }
+                    for solver in df['solver_name'].unique()
+                ]
+
+            def get_metadata(self, df, dataset):
+                df = df[(df['data_name'] == dataset)]
```
tomMoral: It is unused?
```suggestion
```

benchopt/cli/tests/test_cli.py
```diff
@@ -482,11 +482,48 @@ def test_complete_datasets(self, dataset_completion_cases):  # noqa: F811
...
+                title = "Custom Plot 1"
```
tomMoral:
```suggestion
                title = f"Custom Plot - {dataset}"
```

benchopt/plotting/base.py
```diff
@@ -0,0 +1,116 @@
...
+        # Get all combinations
+        dropdown = {**self.dropdown}
+        for k, v in dropdown.items():
+            if v is Ellipsis:
+                if k == "dataset":
+                    dropdown[k] = df['data_name'].unique().tolist()
+                elif k == "solver":
+                    dropdown[k] = df['solver_name'].unique().tolist()
+                elif k == "objective":
+                    dropdown[k] = df['objective_name'].unique().tolist()
```
tomMoral: We should really rename `data_name` with `dataset_name`, to get something more coherent.
Also, we could have:
```suggestion
                if k in ["dataset", "solver", "objective"]:
                    dropdown[k] = df[f"{k}_name"].unique().tolist()
```

benchopt/plotting/default_plots.py
```diff
@@ -0,0 +1,64 @@
...
+            if X_axis == "Iteration":
+                x = df_filtered_median["time"].index.tolist()
+
+            curve_data = {
+                "x": x,
+                "y": y,
+                "color": self.get_style(solver)[0],
+                "marker": self.get_style(solver)[1],
+                "label": solver,
+            }
```
tomMoral, line 41: Update `get_style` to return a dictionary?
```suggestion

            curve_data = {
                "x": x,
                "y": y,
                "label": solver,
                **self.get_style(solver)
            }
```

benchopt/plotting/default_plots.py
```diff
@@ -0,0 +1,64 @@
+from .base import BasePlot
+
+
+class ObjectiveCurvePlot(BasePlot):
```
tomMoral, line 4: Very clean, I like it :)

benchopt/plotting/generate_html.py
```diff
@@ -98,6 +96,10 @@ def get_results(fnames, html_root, benchmark, config=None, copy=False):
             fname = fname_in_output
         fname = fname.absolute().relative_to(html_root.absolute())
 
+        custom_data, custom_dropdown = benchmark.get_plot_data(
+            df, plotly_style=True
+        )
+
```
tomMoral: I would rather have an extra conversion step around here instead of having the in the plot generation.
```suggestion
        custom_data, custom_dropdown = benchmark.get_plot_data(df)
        custom_data = convert_plotly_style(custom_data)

```

benchopt/plotting/generate_matplotlib.py
```diff
@@ -0,0 +1,61 @@
...
+    plt.title(plot_data["title"], fontsize=14)
+    plt.tight_layout()
+
+    save_name = output_dir / f"{key}"
+    if hasattr(fig, 'write_html'):
+        save_name = save_name.with_suffix('.html')
+        fig.write_html(str(save_name), include_mathjax='cdn')
+    else:
+        save_name = save_name.with_suffix('.pdf')
+        plt.savefig(save_name)
```
tomMoral: Is it possible to have `write_html` here? Might be leftover code no?
```suggestion
    save_name = save_name.with_suffix('.pdf')
    plt.savefig(save_name)
```

benchopt/plotting/generate_matplotlib.py
```diff
@@ -0,0 +1,61 @@
...
+            return get_plot_scatter(key, plot_data, df, output_dir)
+        else:
+            raise NotImplementedError(
+                f"Plot type {plot_data['type']} "
+                f"not implemented for matplotlib."
+            )
+
+
+def get_plot_scatter(key, plot_data, df, output_dir):
+    df = df.copy()
```
tomMoral: Not sure it is needed?
```suggestion
def get_plot_scatter(key, plot_data, output_dir):
```

benchopt/plotting/generate_matplotlib.py
```diff
@@ -0,0 +1,61 @@
...
+        if plot_name not in kinds:
+            continue
+        figs.append(get_plot_figure(plot_data[plot_name], df, output_dir))
+    return figs
+
+
+def get_plot_figure(plot_datas, df, output_dir):
+    for key, plot_data in plot_datas.items():
+        if plot_data["type"] == "scatter":
+            return get_plot_scatter(key, plot_data, df, output_dir)
```
tomMoral:
```suggestion
def get_plot_figure(plot_datas, output_dir):
    for key, plot_data in plot_datas.items():
        if plot_data["type"] == "scatter":
            return get_plot_scatter(key, plot_data, output_dir)
```

benchopt/plotting/generate_matplotlib.py
```diff
@@ -0,0 +1,61 @@
...
+def get_plot_scatter(key, plot_data, df, output_dir):
+    df = df.copy()
+
+    fig = plt.figure()
+    for curve_data in plot_data["data"]:
+        plt.loglog(
+            curve_data["x"], curve_data["y"], color=curve_data["color"],
+            marker=curve_data["marker"], label=curve_data["label"],
+            linewidth=3
+        )
```
tomMoral, line 36: probably need to have in metadata if `scale/yscale` are configurable or not?
`loglog` by default is not always a good idea I think.

benchopt/plotting/base.py
```diff
@@ -0,0 +1,116 @@
...
+    def get_metadata(self, df, **kwargs):
+        ...
+
+    def get_style(self, label):
+        idx = self.label_dict.get(label, len(self.label_dict))
+        self.label_dict[label] = idx
+
+        color = COLORS[idx % len(COLORS)]
+
+        return color, idx
```
tomMoral:
```suggestion
        return {'color': color, 'marker': idx}
```

benchopt/plotting/generate_matplotlib.py
etyl: Yeah, I added an issue #849

benchopt/plotting/tests/test_custom_plot.py
```diff
@@ -0,0 +1,115 @@
...
+                "ylabel": "custom objective value",
+            }
+    """
+
+    error_plot = plot.replace("type = 'scatter'", "")
+    with temp_benchmark(plots=error_plot) as bench:
+        with pytest.raises(SystemExit):
+            run([str(bench.benchmark_dir),
+                 *'-n 1 -r 1 --no-display'
+                 .split()], standalone_mode=False)
```
tomMoral: These tests should check that a sensible error is raised, saying that it is missing a `type`.
smthg like:
```suggestion
    with temp_benchmark(plots=error_plot) as bench:
        with CaptureCmdOutput(exit=1) as out:
            run(f"{bench.benchmark_dir} -n 1 -r 1 --no-display'
                 .split(), standalone_mode=False)
        out.check_output("type is missing")
```

benchopt/plotting/tests/test_custom_plot.py
```diff
@@ -0,0 +1,115 @@
...
+
+    error_plot = plot.replace("type = 'scatter'", "type = 'curving'")
+    with temp_benchmark(plots=error_plot) as bench:
+        with pytest.raises(SystemExit):
+            run([str(bench.benchmark_dir),
+                 *'-n 1 -r 1 --no-display'
+                 .split()], standalone_mode=False)
```
tomMoral: All these tests are similar, to avoid redundant test, please use parametrized test with:

```
@pytest.mark.parametrized("pat, replace, msg", [
    ("type = 'scatter'", "", "no type"),
    ("type = 'scatter'", "type = 'fake'", "does not exist"),
    ...
]
def test_custom_plot(pat, replace, msg):
        error_plot = plot.replace(pat, replace)
        with temp_benchmark(plots=error_plot) as bench:
            with CaptureCmdOutput(exit=1) as out:
                run(f"{bench.benchmark_dir} -n 1 -r 1 --no-display'
                 .split(), standalone_mode=False)
       out.check_output(msg)

File: benchopt/plotting/base.py [hunk 58]
```diff
@@ -0,0 +1,112 @@
...
+
+    @abstractmethod
+    def plot(self, df, **kwargs):
+        ...
+
+    @abstractmethod
+    def get_metadata(self, df, **kwargs):
+        ...
+
+    def get_style(self, label):
```
tomMoral, line 58: Missing docstring here.

benchopt/plotting/default_plots.py
```diff
@@ -0,0 +1,63 @@
...
+            df_filtered_median = df_filtered.median()
+            if objective_column not in df_filtered_median:
+                continue
+            y = (
+                df_filtered_median[objective_column]
+                .values.tolist()
+            )
+            x = df_filtered_median["time"].values.tolist()
+            if X_axis == "Iteration":
+                x = df_filtered_median["time"].index.tolist()
```
tomMoral: I don't think this is the right value?
```suggestion
                x = df_filtered_median["stop_val"].index.tolist()
```

benchopt/plotting/helpers.py
```diff
@@ -2,6 +2,36 @@
...
+solvers_idx = {}
+
+CMAP = plt.get_cmap('tab20')
+COLORS = [CMAP(i) for i in range(CMAP.N)]
+COLORS = COLORS[::2] + COLORS[1::2]
+MARKERS = {i: v for i, v in enumerate(plt.Line2D.markers)}
+MARKERS_STR = {v: i for i, v in MARKERS.items()}
+
+
+def get_solver_style(solver, plotly=True):
```
tomMoral, line 17: This is for backward compat with previous plots? if so, add a comment.

benchopt/benchmark.py
```diff
@@ -197,6 +197,39 @@ def check_dataset_patterns(self, dataset_patterns, class_only=False):
...
+
+    def get_plot_data(self, df):
+        "Get the data to plot for the benchmark."
+        self.check_custom_plots()
+        custom_data = {}
+        custom_dropdown = {}
+        for plot in self.get_custom_plots():
+            data, dropdown = plot._get_all_plots(df)
+            custom_data[plot._get_name()] = data
+            custom_dropdown[plot._get_name()] = dropdown
```
tomMoral:
```suggestion
        custom_dropdown = {}
        for plot in self.get_custom_plots():
            plot_name = plot._get_name()
            data, dropdown = plot._get_all_plots(df)
            custom_data[plot_name] = data
            custom_dropdown[plot_name] = dropdown
```

benchopt/plotting/generate_html.py
```diff
@@ -98,22 +96,32 @@ def get_results(fnames, html_root, benchmark, config=None, copy=False):
...
+                'plots',
+                list(PLOT_KINDS) + benchmark.get_custom_plot_names()
+            ),
             metadata=get_metadata(df, config_.get('plot_configs', {})),
+            custom_plot_params=custom_dropdown,
         )
 
         # JSON
         result['json'] = json.dumps(shape_datasets_for_html(df))
+        result['json_custom_plots'] = custom_data
```
tomMoral: Unclear why it is not processed with `json`?

etyl: I intend to remove result["json"] in the future, since it's only for the old plots I believe

tomMoral: yes but I don't understand how you serialize the `custom_data` object to make sure this is `json`.
I would expect:
```suggestion
        result['json_custom_plots'] = json.dumps(custom_data)
```

benchopt/plotting/default_plots.py
```diff
@@ -31,7 +31,7 @@ def plot(self, df, dataset, objective, objective_column, X_axis):
             )
             x = df_filtered_median["time"].values.tolist()
             if X_axis == "Iteration":
-                x = df_filtered_median["time"].index.tolist()
+                x = df_filtered_median.index.tolist()
```
tomMoral, line 34: Ah, did not see the index ^^

benchopt/plotting/tests/test_custom_plot.py
tomMoral, line 1: Nice!

benchopt/plotting/default_plots.py
etyl: it's more clear now :)

PR #860, opened 2025-11-17
ENH parallelize reps

etyl:
### Checks before merging PR
- [x] added unit test


benchopt/tests/test_benchmark_features.py
```diff
@@ -486,12 +486,12 @@ def get_result(self): return dict(beta=1)
...
             expected_home.format(bench_dir=bench.benchmark_dir.as_posix())
         ).resolve()
-        out.check_output(re.escape(f"HOME:{expected_home}"), repetition=1)
+        out.check_output(re.escape(f"HOME:{expected_home}"), repetition=n_jobs)
 
         expected_path = Path(
             expected_path.format(bench_dir=bench.benchmark_dir.as_posix())
         ).resolve()
-        out.check_output(re.escape(f"PATH:{expected_path}"), repetition=1)
+        out.check_output(re.escape(f"PATH:{expected_path}"), repetition=n_jobs)
```
etyl, line 524: not sure if this is intended

benchopt/parallel_backends/slurm_executor.py
```diff
@@ -99,25 +98,23 @@ def run_on_slurm(
...
         if exc is not None:
-            for tt in tasks:
-                tt.cancel()
+            # Cancel remaining tasks and raise error
+            for t in tasks:
+                t.cancel()
             raise exc
 
-    return [t.results()[0] for t in tasks]
+        yield completed_future.result()
```
tomMoral: You should keep `results()[0]`, otherwise it won't work with distributed runs on large scale benchmark.

benchopt/parallel_backends/slurm_executor.py
```diff
@@ -99,25 +98,23 @@ def run_on_slurm(
...
-        exc = t.exception()
+    # Yield results as jobs finish (unordered)
+    for completed_future in as_completed(tasks):
+        exc = completed_future.exception()
         if exc is not None:
-            for tt in tasks:
-                tt.cancel()
+            # Cancel remaining tasks and raise error
+            for t in tasks:
+                t.cancel()
```
tomMoral: Why do you change the name from `tt` to `t`?
If there is no good reason, avoid extra changes that are unrelated to the PR.

benchopt/parallel_backends/slurm_executor.py
```diff
@@ -3,7 +3,6 @@
 try:
     import submitit
     from submitit.helpers import as_completed
-    from rich import progress
```
tomMoral, line 6: If you don't use `rich` anymore for the backend, remove it from the deps in `pyproject.toml`

benchopt/parallel_backends/tests/test_slurm_executor.py
```diff
@@ -92,15 +92,10 @@ def exception(self): return None
...
+        def result(self):
             func, args, kwargs = self.task
             res = func(*args, **kwargs)
-            res = [
-                {**r, **{f"s_{k}": v for k, v in self.config.items()}}
-                for r in res
-            ]
-
-            return [res]
+            return res
```
tomMoral: Unclear why you change this?

benchopt/base.py
```diff
@@ -235,12 +238,16 @@ def _warm_up(self):
...
     def _get_state(self):
         """Return the state of the objective for pickling."""
-        return dict(objective=getattr(self, '_objective', None))
+        return dict(
+            objective=getattr(self, '_objective', None),
+            objective_dict=getattr(self, '_objective_dict', None)
+        )
 
     def __setstate__(self, state):
+        objective_dict = state['objective_dict']
```
tomMoral: why do you need this extra thing?

benchopt/runner.py
```diff
@@ -96,8 +95,6 @@ def run_one_to_cvg(benchmark, objective, solver, meta, stopping_criterion,
     force : bool
         If force is set to True, ignore the cache and run the computations
         for the solver anyway. Else, use the cache if available.
-    terminal : TerminalOutput or None
-        Object to format string to display the progress of the solver.
```
tomMoral, line 112: This should stay but remove `stopping_criterion` and add `timeout, max_runs`

benchopt/runner.py
```diff
@@ -168,12 +172,19 @@ def run_one_to_cvg(benchmark, objective, solver, meta, stopping_criterion,
...
     if ctx.status in FAILURE_STATUS:
         raise FailedRun(ctx.status)
-    return curve, ctx.status
+
+    key = (
+        meta['dataset_name'],
+        meta['objective_name'],
+        meta['solver_name']
+    )
+    return curve, (key, ctx.status)
```
tomMoral: why not stay flat with `curve, key, status`?

benchopt/runner.py
```diff
@@ -396,14 +358,64 @@ def _run_benchmark(benchmark, solvers=None, forced_solvers=None,
...
+        try:
+            results = run_one_to_cvg_cached(**kwargs)
+        except FailedRun as e:
+            key = (
+                kwargs['meta']['dataset_name'],
+                kwargs['meta']['objective_name'],
+                kwargs['meta']['solver_name']
+            )
+            results = ([], (key, e.status))
+        return results
```
tomMoral: It would probably be easier to change the logic in `run_one_to_cvg`, to always return `[], (key, e.status)` instead of raising `FailedRun`. I think we were using the error logic to stop the run as fast as possible without continuing the repetition, but now we could directly return the failed status?

benchopt/runner.py
```diff
@@ -168,12 +172,19 @@ def run_one_to_cvg(benchmark, objective, solver, meta, stopping_criterion,
...
+        meta['objective_name'],
+        meta['solver_name']
+    )
+    return curve, (key, ctx.status)
 
 
-def run_one_solver(benchmark, dataset, objective, solver, n_repetitions,
-                   max_runs, timeout=None, force=False, collect=False,
-                   terminal=None, pdb=False):
+def get_run_args(
```
tomMoral: Can't this be integrated in the `_get_all_runs` method?

benchopt/runner.py
```diff
@@ -396,14 +358,64 @@ def _run_benchmark(benchmark, solvers=None, forced_solvers=None,
...
+            )
+            results = ([], (key, e.status))
+        return results
+
+    total_cvg_kwargs = []
+    for run_kwargs in all_runs:
+        cvg_kwargs = get_run_args(
+            **common_kwargs, **run_kwargs
+        )
+        total_cvg_kwargs.extend(cvg_kwargs)
```
tomMoral: This is usually bad practice to put all the runs in a list as it might be super large.
Usually, we use an iterator to avoid wasting memory by creating all the objects in memory before we really need them

benchopt/runner.py
```diff
@@ -396,14 +358,64 @@ def _run_benchmark(benchmark, solvers=None, forced_solvers=None,
...
+        config=parallel_config, collect=collect
+    )
+
+    status_dict = {}
+
+    for results in results_generator:
+        result = results[0]
+        run_statistics.extend(result)
+
+        key, status = results[1]
```
```suggestion
    for result, (key, status) in results_generator:
        run_statistics.extend(result)

```

benchopt/runner.py
```diff
@@ -396,14 +358,64 @@ def _run_benchmark(benchmark, solvers=None, forced_solvers=None,
...
+        if key not in status_dict:
+            status_dict[key] = status
+        elif status_dict[key] == 'done':
+            status_dict[key] = status
+
+    for key, status in status_dict.items():
+        terminal.set(dataset=key[0], objective=key[1], solver=key[2])
+        terminal.show_status(status)
```
tomMoral: The issue here is that the status should be updated as soon as we have all the info from one solver run.

tomMoral (commented):
Nice, this is much easier to review than the other one.
A first pass of comments. There are stuff to improve but overall, it is going in the right direction.

benchopt/base.py
etyl: otherwise the cv is recomputed in each job (and will have the wrong split) so the data needs to be pickled, maybe there is another solution

benchopt/runner.py
etyl: this is to avoid caching failed runs, it must throw an error to not cache the function

benchopt/runner.py
etyl: Maybe we can leave as is for this PR since this should be fixed with the rich logging ?

benchopt/runner.py
tomMoral: Could you add a comment ?

bruAristimunha:
This feature would be super nice! @tomMoral and @etyl

benchopt/cli/main.py
```diff
@@ -586,7 +586,7 @@ def test(benchmark, env_name, recreate, pytest_args):
     _bench_test_file = _bench_test_module / "test_benchmarks.py"
 
     cmd = (
-        f'pytest {pytest_args} {_bench_test_file} '
+        f'python -m pytest {pytest_args} {_bench_test_file} '
```
tomMoral, line 589: Why do you need this change?

benchopt/helpers/tests/test_language_helpers.py
```diff
@@ -148,7 +148,7 @@ def get_result(self): return {'beta': self.beta.ravel()}
         with CaptureCmdOutput() as out:
             run([
                 str(bench.benchmark_dir), '-s', 'julia_solver', '-n', '1',
-                '-r', 1, '-d', 'simulated', '--no-plot',
+                '-r', 1, '-d', 'simulated', '--no-plot', '--timeout', '500',
```
tomMoral, line 151: why do you need this change?

benchopt/tests/test_runner.py
```diff
@@ -399,7 +399,7 @@ class TestCache:
     class Solver(TempSolver):
         name = "test-solver"
         sampling_strategy = 'run_once'
-        def run(self, _): print("#RUN_SOLVER")
+        def run(self, _): print("#RUN_SOLVER", flush=True)
```
tomMoral, line 402: I feel that the flush should be done in the solver, to make sure we have reliable output, instead of putting flush everywhere in the tests.
Users would be surprised otherwise by the need to use `flush`.

benchopt/tests/test_benchmark_features.py
```diff
@@ -212,7 +212,8 @@ class Solver(TempSolver):
         name = "test-solver"
         sampling_strategy = 'run_once'
         def set_objective(self, X_train, y_train): pass
-        def run(self, n_iter): print("OK")
+        def run(self, n_iter): print("OK", flush=True)
+        def get_result(self): return dict(beta=1)
```
tomMoral, line 216: This should already be in `TempSolver` no?
```suggestion
```

benchopt/utils/terminal_output.py
```diff
@@ -73,24 +74,17 @@ def __init__(self, n_repetitions=None, show_progress=None):
             kernel32 = ctypes.windll.kernel32
             kernel32.SetConsoleMode(kernel32.GetStdHandle(-11), 7)
 
-        self.n_repetitions = n_repetitions
+        self.n_repetitions = n_repetitions if n_repetitions is not None else 1
```
tomMoral, line 77: Is it still needed, I think I fixed this in `runner.py` previously.

benchopt/runner.py
```diff
@@ -121,11 +124,33 @@ def run_one_to_cvg(benchmark, objective, solver, meta, stopping_criterion,
...
+    stopping_criterion = solver._stopping_criterion.get_runner_instance(
+        solver=solver,
+        max_runs=max_runs,
+        timeout=timeout,
+        terminal=terminal,
+    )
+    stopping_criterion.dataset = meta['dataset_name']
+    stopping_criterion.objective = meta['objective_name']
+    stopping_criterion.solver = meta['solver_name']
+    stopping_criterion.rep = meta['idx_rep']
```
tomMoral, line 143: This feels like the `stopping_criterion` only need the key here no? so the `key` could be computed at the beginning of `run_one_to_cvg` and pass to the stopping_criterion in `get_runner_instance`, stored as `run_key`?

benchopt/utils/terminal_output.py
```diff
@@ -137,29 +132,57 @@ def display_dataset(self):
...
         """Display progress in the CLI interface."""
         if self.show_progress:
             if isinstance(progress, float):
                 progress = f'{progress:6.1%}'
+            solver_tag = colorify(f"    |--{key[2]}:")
             print_normalize(
-                f"{self.solver_tag} {progress} "
-                f"({self.rep + 1} / {self.n_repetitions} reps)",
+                f"{solver_tag} {progress} "
+                f"({rep + 1} / {self.n_repetitions} reps)",
```
tomMoral, line 146: shouldn't rep be computed based on `self.rep`?

benchopt/base.py
```diff
@@ -299,12 +300,10 @@ def _warm_up(self):
 
     def _get_state(self):
         """Return the state of the objective for pickling."""
-        return dict(objective=getattr(self, '_objective', None))
+        return {}
```
tomMoral, line 303: why do you remove the pickling of the `objective` attribute here?
If it is really intended, then you can remove `__setstate__/_get_state` completely.

benchopt/benchmark.py
```diff
@@ -658,7 +658,7 @@ def install_all_requirements(self, include_solvers, include_datasets,
     def download_all_data(self, datasets, env_name, quiet):
         if len(datasets) == 0:
             return 0
-        cmd = f"benchopt check-data {self.benchmark_dir} -d "
+        cmd = f"python -m benchopt check-data {self.benchmark_dir} -d "
```
tomMoral, line 661: Why this change?

benchopt/benchmark.py
```diff
@@ -760,7 +760,7 @@ def _get_all_runs(self, solvers=None, forced_solvers=None,
                     )
                     yield dict(
                         dataset=dataset, objective=objective, solver=solver,
-                        force=force, terminal=terminal.clone()
+                        force=force, terminal=terminal
```
tomMoral, line 763: Why do you still need the terminal in the job? all print should be handled  in the main process no?
Or is it just to show progress?
```suggestion
                        force=force,
```

benchopt/runner.py
```diff
@@ -441,14 +422,54 @@ def _run_benchmark(benchmark, solvers=None, forced_solvers=None,
...
+
+    results_generator = parallel_run(
+        benchmark, run_one_to_cvg_final, total_cvg_kwargs_generator,
+        config=parallel_config, collect=collect
+    )
+
+    for result, key, status, reason in results_generator:
+        run_statistics.extend(result)
+        terminal.set(dataset=key[0], objective=key[1], solver=key[2])
+        terminal.show_status(status=status, reason=reason)
```
tomMoral, line 472: If `set` is not use elsewhere, why not simplify and pass `key` to show_status?
also, instead of show status, I would call this `update_status`.

benchopt/runner.py
tomMoral, line 277: This should be done in `run_one_to_cvg` no?

benchopt/runner.py
```diff
@@ -272,15 +290,13 @@ def run_one_to_cvg_cached(**kwargs):
...
-        if skip:
-            terminal.skip(reason)
-            return []
+    timeout = timeout / n_repetitions if timeout is not None else None
 
-        terminal.set(rep=rep)
+    for rep in range(n_repetitions):
+        objective_rep = copy.copy(objective)
+        objective_rep.repetition = rep
+        solver._objective = objective_rep
```
tomMoral, line 299: Why not move this in the `get_all_runs` function?

benchopt/runner.py
```diff
@@ -184,12 +209,19 @@ def run_one_to_cvg(benchmark, objective, solver, meta, stopping_criterion,
...
+    return curve, key, ctx.status, ""
 
 
-def run_one_solver(benchmark, dataset, objective, solver, n_repetitions,
-                   max_runs, timeout=None, force=False, collect=False,
-                   terminal=None, pdb=False):
+def get_solver_kwargs(
+    benchmark, dataset, objective, solver, n_repetitions, max_runs,
+    timeout=None, force=False, collect=False, terminal=None, pdb=False
+):
```
tomMoral, line 224: Overall, I think the generation of all configs is scattered across too many places. I would try to keep it as monolytique as possible, to allow making `groupby` by later on.

tomMoral (commented):
A first pass. happy to discuss it soon

benchopt/cli/main.py
etyl: I need this to run tests locally, in general this would be helpful in case the default pytest is not using the local environment.

benchopt/tests/test_runner.py
etyl: what do you mean by flushing from the solver?

benchopt/benchmark.py
etyl: same as for pytest

benchopt/runner.py
etyl: set is used at other places, sometimes we set only the dataset and objective (if skipped).

benchopt/runner.py
etyl: I'm not sure, the function might become too big, doing like this splits it in half.

benchopt/runner.py
etyl: In the end we get the config generator for all configs, we can then apply the groupby on it if needed.

benchopt/runner.py
etyl: It needs access to the seed in order to do `set_dataset`

benchopt/benchmark.py
etyl: yes it's just to show the progress (same behaviour as before)

benchopt/base.py
etyl: there are some caching issues if we keep it in the state

benchopt/tests/test_runner.py
tomMoral: Check that the `print(end='', flush=True)` that was at the end of the `run_one_solver` is still present at the end of a parallel job.

benchopt/runner.py
tomMoral: Might be better to move both functions in a new file called `_generate_runs.py` to make the logic more tied together, in particular to prepare with the `groupby` later on.

benchopt/helpers/tests/test_language_helpers.py
tomMoral: Unclear why?

PR #912, opened 2026-04-13
ENH add benchopt prepare to run dataset preparation operations

tomMoral:
### Checks before merging PR
- [x] added unit test


bruAristimunha:
Amazing PR @tomMoral!

benchopt/cli/main.py
```diff
@@ -465,7 +541,7 @@ def install(
         if not solver_names:
             solver_names = config.get("solver", tuple())
             forced_solvers = config.get("force-solver", tuple())
-            solver_names = list(set(solver_names).union(set(forced_solvers)))
+            solver_names = list(solver_names) + list(forced_solvers)
```
etyl, line 544: why are you keeping duplicates now?

benchopt/utils/parametrized_name_mixin.py
```diff
@@ -131,7 +135,53 @@ def expand(keys, values):
...
+    for update in product_param(params, ignore=ignore):
+        for default in product_param(klass.parameters, ignore=ignore):
+            default = default.copy()  # avoid modifying the original
+
+            # check that all parameters are defined in klass.parameters
+            for key in update:
+                if key not in default:
+                    raise ValueError(
+                        f"Unknown parameter '{key}', parameter must be in "
+                        f"{list(default.keys())}")
```
etyl:
```suggestion
            keys_unknown = set(update) - set(default)
            if keys_unknown:
                raise ValueError(
                    f"Unknown parameters {list(keys_unknown)}, parameters must be in "
                    f"{list(default.keys())}"
                ) 
```

benchopt/cli/main.py
tomMoral: Because actually, you can have `solver_names` as a dictionary of configs when it is parsed from a config file.

I think we should further improve this API in a follow up PR on but this is already better than the previous case.
But good catch I will add an issue.

benchopt/utils/parametrized_name_mixin.py
tomMoral: Much better indeed! thanks!

benchopt/cli/main.py
tomMoral: I opened #918 to track this one.

etyl: ah okay, yes this seems a bit confusing

benchopt/utils/tests/test_parametrized_name_mixin.py
```diff
@@ -0,0 +1,350 @@
...
+        # 2 x 3 = 6 combinations
+        assert len(result) == 6
+        for params in result:
+            assert set(params.keys()) == {'n', 'seed'}
+
+    def test_invalid(self):
+        class _Cls:
+            parameters = {'n': [1000, 10000], 'seed': [0, 1, 2]}
+
+        with pytest.raises(ValueError, match="Unknown parameter 'invalid'"):
```
tomMoral:
```suggestion
        with pytest.raises(ValueError, match="Unknown parameterd ['invalid']"):
```

PR #913, opened 2026-04-13
ENH improve mpl display for boxplots

etyl:
### Checks before merging PR
- [ ] added unit test

### DIFF
--- a/benchopt/plotting/generate_matplotlib.py
+++ b/benchopt/plotting/generate_matplotlib.py
@@ -125,6 +125,7 @@ def get_plot_barchart(plot_data):
 
 def get_plot_boxplot(plot_data):
     fig = plt.figure()
+    ax = fig.gca()
 
     # collect the union of all labels (x tick names)
     all_labels = []
@@ -140,7 +141,7 @@ def get_plot_boxplot(plot_data):
         boxplot = plt.boxplot(
             data["y"],
             positions=positions,
-            widths=0.6,           # you can keep this fixed
+            label=data["label"],
             patch_artist=True,
         )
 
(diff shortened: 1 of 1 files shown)


benchopt/plotting/generate_matplotlib.py
```diff
@@ -140,7 +141,7 @@ def get_plot_boxplot(plot_data):
         boxplot = plt.boxplot(
             data["y"],
             positions=positions,
-            widths=0.6,           # you can keep this fixed
```
tomMoral, line 143: Why did you remove this?

benchopt/plotting/generate_matplotlib.py
```diff
@@ -152,11 +153,15 @@ def get_plot_boxplot(plot_data):
...
             flier.set(color=color)
 
-    plt.xticks(range(len(all_labels)), all_labels, rotation=45)
-    plt.title(plot_data["title"])
-    plt.ylabel(plot_data["ylabel"])
+    ax.set_xticks(range(len(all_labels)), all_labels, rotation=45)
+    ax.set_title(plot_data["title"])
+    ax.set_ylabel(plot_data["ylabel"])
+    ax.legend()
+    fig.tight_layout()
```
tomMoral, line 173: I usually dislike using `tight_layout` which tends to change the font size compared to the configuration.

tomMoral (commented):
Can you share a before/after?

benchopt/plotting/generate_matplotlib.py
etyl: I'm not sure have a fixed size width is a good idea, if the x-axis of the box plots becomes tighter (for example plotting something at some intervals in time)  then the render will not be good anymore

benchopt/plotting/generate_matplotlib.py
etyl: It's what is currently being done for the other plots, so either we remove them all, or use it everywhere

tomMoral (approved):
LGTM!

PR #917, opened 2026-04-16
No-op if Benchmark initialized with Benchmark

### LINKED ISSUE #915: API function plot_benchmark takes a Benchmark object as argument
The run_benchmark function in the API takes as input a `Path` object to reference the benchmark, whereas plot_benchmark takes a `Benchmark` object. This forces the user to write something like

```python
from benchopt import run_benchmark
from benchopt.benchmark import Benchmark

benchmark_path = '.'
benchmark = Benchmark(benchmark_path,no_cache=False)

result_file = run_benchmark(benchmark_path=benchmark_path, options...)

plot_benchmark(
    fname=result_file,
    benchmark=benchmark,
)
```
I would be cleaner to have the same argument type for the two functions. I would advocate for using `Path` objects.

DamienLesens:
### Checks before merging PR
- [x] added unit test
- [x] ~~edited the [what's new](../../whatsnew.rst) (if applicable)~~

Fixes #915

### DIFF
--- a/benchopt/benchmark.py
+++ b/benchopt/benchmark.py
@@ -82,17 +82,21 @@ class Benchmark:
         Caching mechanism for the benchmark.
     """
 
+    def __new__(cls, benchmark, *args, **kwargs):
+        # if already a Benchmark, act as a no-op
+        if isinstance(benchmark, cls):
+            return benchmark
+        return super().__new__(cls)
+
     def __init__(
             self, benchmark_dir,
             seed=None,
             no_cache=False,
             allow_meta_from_json=False,
     ):
-        # if already a Benchmark, act as a no-op
         if isinstance(benchmark_dir, self.__class__):
-            self.__dict__ = benchmark_dir.__dict__
             return
-        
+
         self.benchmark_dir = Path(benchmark_dir)
         self.no_cache = no_cache
 
--- a/benchopt/benchmark.py
+++ b/benchopt/benchmark.py
@@ -61,10 +61,11 @@ class Benchmark:
 
     Parameters
     ----------
-    benchmark_dir : str or Path-like
+    benchmark_dir : str or Path-like or Benchmark object
         Folder containing the benchmark. The folder should at least
         contain an `objective.py` file defining the `Objective`
         function for the benchmark.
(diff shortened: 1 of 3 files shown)


benchopt/benchmark.py
```diff
@@ -87,6 +88,11 @@ def __init__(
             no_cache=False,
             allow_meta_from_json=False,
     ):
+        # if already a Benchmark, act as a no-op
+        if isinstance(benchmark_dir, self.__class__):
+            self.__dict__ = benchmark_dir.__dict__
+            return
+
```
tomMoral, line 95: I would rather rely on `__new__` for the no-op, so something like:
```suggestion

    def __new__(cls, benchmark, *args, **kwargs):
        # if already a Benchmark, act as a no-op
        if args and isinstance(benchmark, cls):
            return benchmark
        return super().__new__(cls, benchmark, *args, **kwargs)

    def __init__(
            self, benchmark_dir,
            seed=None,
            no_cache=False,
            allow_meta_from_json=False,
    ):
```

tomMoral (commented):
Great thanks for the PR!
Just a comment, as I am not a fan of messing with `__dict__` attributes.

Can you also add a test in `benchopt/tests/test_benchmark_features.py` to check that it accept all these arguments?
No need for doc on this one.

benchopt/benchmark.py
```diff
@@ -81,12 +82,21 @@ class Benchmark:
         Caching mechanism for the benchmark.
     """
 
+    def __new__(cls, benchmark, *args, **kwargs):
+        # if already a Benchmark, act as a no-op
+        if isinstance(benchmark, cls):
+            return benchmark
```
tomMoral:
```suggestion
    def __new__(cls, benchmark_dir, *args, **kwargs):
        # if already a Benchmark, act as a no-op
        if isinstance(benchmark_dir, cls):
            return benchmark_dir
```

tomMoral (approved):
LGTM! I took the liberty to fix the error and add two tests (overriding `__new__` introduced an error with `pickle` that was a bit complex, sorry for that).

PR #924, opened 2026-05-03
ENH add decorator based interface API to benchopt (mini)

### LINKED ISSUE #605: Implement decorators to allow on-the-fly benchmarks
Idea would be to implement decorators @dataset @solver @objective to be able to generate a benchmark from a single .py

Typical use case: during numerical experiments dev, be able to compare few algorithms on simple dataset, without creating a full benchmark repo.

tomMoral:
Add decorators to create benchmark components based on functional implementation:

- `@dataset` defines a `Dataset` whose `get_data` method is the decorated function.
- `@objective` defines an `Objective` whose `evaluate_result` method is the decorated function.
- `@solver` defines a `Solver` whose `run` method is the decorated function.

These decorator can take `name` to give a name to the object, and its keyword arguments define the parameters of the object.
Using these object, the benchmark can be either runprogramatically or  with the CLI.

Supersedes #641
Fixes #605 

### Checks before merging PR
- [x] added unit test


benchopt/mini.py
```diff
@@ -0,0 +1,500 @@
...
+        class Solver(BaseSolver, metaclass=_CallableFnMeta):
+            name = _name
+            parameters = _params
+            sampling_strategy = "run_once"
+            _fn = staticmethod(fn)
+
+            def set_objective(self, **objective_dict):
+                self._objective_dict = objective_dict
+
+            def run(self, n_iter):
```
etyl: Why is there `n_iter` if you use the strategy `run_once`? I think it would be clearer to remove this arg in the decorated function and leave only the parameters of the solver as args.

benchopt/mini.py
```diff
@@ -0,0 +1,500 @@
...
+>>>
+>>> @dataset(size=100, random_state=0)
+... def simulated(size, random_state):
+...     X = list(range(size))
+...     return dict(X=X)
+>>>
+>>> @solver(name="Solver 1", lr=[1e-2, 1e-3])
+... def solver1(n_iter, X, lr):
+...     return dict(beta=sum(X) * lr)
+>>>
```
etyl, line 20: I would be nice to have to way to log metrics when the solver runs, like a callback to the objective evaluation. Also, as I mentioned in another comment, it would maybe be clearer to keep as args of the solver only the parameters (remove `n_iter`).

PR #930, opened 2026-05-06
MNT use base image for circleci

etyl:
Not sure why we were using a python 3.10 image, since we install python with mamba afterwards.


tomMoral (approved):
Amazing!!!  Thanks a lot @etyl

PR #940, opened 2026-05-29
ENH add install backends (uv/requirements)

tomMoral:
### Checks before merging PR
- [ ] added unit test


LTMeyer (commented):
# What I have tested

I checked it was working with uv doing the following, in the cloned repository.
```bash
uv venv  # install the environment
uv pip install .  # install benchopt
git clone https://github.com/benchopt/template_benchmark_ml.git  # clone the template benchmark in the current folder
uv run benchopt install template_benchmark_ml
uv run benchopt run template_benchmark_ml
```

While the above still runs, it seems some dependencies are overlooked. 
For instance, I got the following message:
> ModuleNotFoundError: No module named 'sklearn'

# Quick questions

Why do rely on `benchopt` for the installation? 
Would using `uv` only for the installation work?
For instance, replacing the 3 last lines of the command above with:

```bash
git clone https://github.com/benchopt/template_benchmark_ml.git  # clone the template benchmark in the current folder
cd template_benchmark_ml
uv pip install .  # Delegate the installation to uv
uv run benchopt run .  # Run the benchmark
```

Relying on uv to handle the dependencies and be sure `scikit-learn` is not omitted in the example above.

LTMeyer:
Just checked with the latest changes.

The following commands still prints missing `sklearn`package warnings, although the package is correctly installed.
```
uv run benchopt install template_benchmark_ml
```

Doing the following works without any warnings:
```
uv run benchopt install template_benchmark_ml --backend requirements
uv pip install -r template_benchmark_ml/requirements.txt
```

PR #953, opened 2026-06-11
ENH table ordering

### LINKED ISSUE #943: VIZ Table doesn't allow ordering per metric
Ordering should be possible according to any of the metric columns.

Image

etyl:
### Checks before merging PR
- [X] added documentation for any new feature
- [X] added unit test
- [X] edited the [what's new](../../whatsnew.rst) (if applicable)

Fixes #943

Tables are now much more customizable (search, ordering, column selection). Uses gridjs.

image


doc/whats_new.rst
```diff
@@ -10,6 +10,15 @@ What's new
 Version 1.10.0 -- In development
 --------------------------------
 
+API
+~~~
```
tomMoral, line 21:
```suggestion
PLOT
~~~~
```

doc/whats_new.rst
```diff
@@ -10,6 +10,15 @@ What's new
...
+
+- Improve the html report table, now rendered with `Grid.js
+  <https://gridjs.io/>`_: rows can be ordered by any column by clicking on its
+  header (with a toggleable arrow for increasing/decreasing order), a search
+  bar filters the rows, and each column can be shown or hidden.
+  By `Hippolyte Verninas`_ (:gh:`953`)
```
```suggestion
- Improve the html report table, now rendered with `Grid.js
  <https://gridjs.io/>`_: sortable by any column, filter by solver name,
  hide/show any columns. By `Hippolyte Verninas`_ (:gh:`953`)
```

tomMoral: Is it possible to store the config (which row to sort on/filters/hidden columns) in a view?

benchopt/plotting/html/static/result.js
```diff
@@ -1040,72 +1040,158 @@ const renderImages = () => {
...
+ * Sort the rows according to the `default_order_column` (a column name or
+ * index) and `default_order_ascending` metadata keys. When absent, the rows
+ * are sorted on the first column in increasing order. Grid.js then handles
+ * the interactive sorting from this initial order.
+ */
+const applyDefaultOrder = (plotData) => {
+  let column = 0;
+  const orderColumn = plotData.default_order_column;
+  if (typeof orderColumn === 'string') {
+    const idx = plotData.columns.indexOf(orderColumn);
```
tomMoral, line 1124: Maybe protect this if orderColum is not in plotdata.columns so it does not break the plot?

tomMoral (commented):
A few comments but otherwise, LGTM

benchopt/plotting/html/static/result.js
etyl: in this case indexOf should return -1, which sets the idx to 0

benchopt/plotting/html/static/result.js
```diff
@@ -403,7 +407,14 @@ const setConfig = (config_item) => {
     if ("plot_kind" in config) {
       kind = config["plot_kind"];
     }
+    // Table settings are applied by renderTable, not through the state.
+    tablePendingView = ('table_order' in config || 'table_hidden_columns' in config)
```
tomMoral, line 411: Why not use `state` here? that would avoid having two pathes for the code, which make it more complex to read no?

doc/whats_new.rst
```diff
@@ -10,14 +10,19 @@ What's new
...
 ~~~
 
-- Improve the html report table, now rendered with `Grid.js
-  <https://gridjs.io/>`_: rows can be ordered by any column by clicking on its
-  header (with a toggleable arrow for increasing/decreasing order), a search
-  bar filters the rows, and each column can be shown or hidden.
-  By `Hippolyte Verninas`_ (:gh:`953`)
+- Custom plot ``options`` values can now be a callable taking the results
+  DataFrame as input and returning the list of possible values for the option.
+  By `Hippolyte Verninas`_ (:gh:`952`)
```
tomMoral, line 25: This was a missing what's new entry?

tomMoral (commented):
Is it not possible to use the state instead of having a completely diffrent mechanism?

other LGTM

doc/whats_new.rst
etyl: this was maybe a bug with github, I merged main now and it doesn't display as an add

benchopt/plotting/html/static/result.js
etyl: Since the table is handled by Grid.js, it's more efficient for it to handle table changes directly instead of running the entire `setState` on each table change. Not sure if it would be useful to move it to the state

tomMoral: Let's not over complicate. I think the right thing to do would be to have views for all plots instead of this global state That I find not so easy to navigate. Probably a good PR for later.

tomMoral (approved):

PR #965, opened 2026-06-22
ENH resize plotly figure

etyl:
### Checks before merging PR
- [x] ~~added unit test~~

### DIFF
--- a/benchopt/plotting/html/static/hover_index.css
+++ b/benchopt/plotting/html/static/hover_index.css
@@ -330,6 +330,15 @@ input:checked+.slider:before {
 /***** Custom legend *****/
 /*----------------------------------------*/
 
+/* Grow the legend to fill the space below the figure so its background runs to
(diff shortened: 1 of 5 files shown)


etyl:
Fixes #964

tomMoral:
Not sure I am a big fan of the design.
I don't like the figure size box that is on top (taking too much attention) and I would prefer that by default it gets the right size, with an option to resize by clicking somewhere?

etyl:
> Not sure I am a big fan of the design. I don't like the figure size box that is on top (taking too much attention) and I would prefer that by default it gets the right size, with an option to resize by clicking somewhere?

Yes I agree the box to resize by value isnt very pretty, I will try to improve this. Did you notice you can drag the plot on the edges as well to resize?

etyl:
This feature is especially useful when you want to export a figure.

PR #974, opened 2026-07-02
FIX Windows shell handling under bash shell

jolars:
`benchopt install`/`test` broke on Windows when the CI step ran under a bash shell (e.g. `shell: bash`), where `$SHELL` is a git-bash path. benchopt then emitted a bash script containing `CALL conda` and invoked it with an unquoted, space-bearing path, and swallowed the real error.

Here are the changes:

- Invoke the shell as an argument list (`_split_shell`), handling spaces in the shell path and dropping the redundant `shell=True` layer.
- Gate the conda `CALL` prefix on cmd detection (`_conda_cmd`) instead of `sys.platform`, so it matches the shell benchopt generates scripts for.
- On Windows, ignore a bash-like `$SHELL` and default to `cmd /c`; override with `BENCHOPT_SHELL`.
- Surface the underlying error when `capture_stdout=False` instead of raising an empty `RuntimeError`.
- Add a `test_windows_bash` CI job that reproduces the scenario, plus unit tests for the new helpers.

I used Claude to out with the PR as well as the PR summary above.

### Checks before merging PR
- [x] added unit test

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -127,12 +127,8 @@ jobs:
         run: benchopt run examples/minimal_benchmark -r 1 -n 10 --no-plot
 
   test_windows_bash:
-    # Regression test for shell handling on Windows when the CI step runs under
-    # a bash shell (the common `shell: bash` setup in downstream benchmarks).
-    # In that case $SHELL is a git-bash path, which used to make benchopt emit a
-    # bash script containing `CALL conda` and break conda env management. See
-    # https://github.com/benchopt/benchopt for the fix. Unlike the `cmd`-based
-    # `test_benchopt` Windows leg, this exercises the bash-shell code path.
+    # Regression test for shell handling on Windows when using a bash shell
+    # instead of cmd.
     name: Test Windows with bash shell
     runs-on: windows-latest
     defaults:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -127,8 +127,9 @@ jobs:
         run: benchopt run examples/minimal_benchmark -r 1 -n 10 --no-plot
 
   test_windows_bash:
-    # Regression test for shell handling on Windows when using a bash shell
-    # instead of cmd.
+    # Regression test: when the CI step runs under bash, $SHELL is a git-bash
+    # path. benchopt must ignore it and drive conda through `cmd /c` rather than
+    # emitting a broken bash script, so `install`/`test` keep working.
     name: Test Windows with bash shell
     runs-on: windows-latest
     defaults:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -133,7 +133,7 @@ jobs:
     # bash script containing `CALL conda` and break conda env management. See
     # https://github.com/benchopt/benchopt for the fix. Unlike the `cmd`-based
     # `test_benchopt` Windows leg, this exercises the bash-shell code path.
-    name: Test Windows under bash shell
+    name: Test Windows with bash shell
     runs-on: windows-latest
     defaults:
       run:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -126,6 +126,46 @@ jobs:
       - name: Run benchopt run without conda
         run: benchopt run examples/minimal_benchmark -r 1 -n 10 --no-plot
 
+  test_windows_bash:
+    # Regression test for shell handling on Windows when the CI step runs under
+    # a bash shell (the common `shell: bash` setup in downstream benchmarks).
+    # In that case $SHELL is a git-bash path, which used to make benchopt emit a
+    # bash script containing `CALL conda` and break conda env management. See
+    # https://github.com/benchopt/benchopt for the fix. Unlike the `cmd`-based
+    # `test_benchopt` Windows leg, this exercises the bash-shell code path.
+    name: Test Windows under bash shell
+    runs-on: windows-latest
+    defaults:
+      run:
+        shell: bash -el {0}
+
+    steps:
+      - uses: actions/checkout@v6
+      - name: Setup Conda
+        uses: conda-incubator/setup-miniconda@v4
+        with:
+          activate-environment: 'testcondaenv'
+          python-version: "3.12"
+          # Use miniforge to only get conda-forge as default channel.
+          miniforge-version: latest
+
+      - name: Show the shell benchopt sees
+        run: echo "SHELL=$SHELL"
+
+      - name: Install benchopt and its dependencies
+        run: pip install -e ".[test]"
+
+      - name: Create a conda env for the benchmark under bash
+        # Exercises create_conda_env / `CONDA_CMD env create` on Windows while
+        # $SHELL is a bash path.
+        run: benchopt install examples/minimal_benchmark --env-name bench_win_bash
+
+      - name: Run the benchmark in that env under bash
+        # Exercises `CALL conda activate` via _run_shell_in_conda_env.
+        run: >
+          benchopt run examples/minimal_benchmark
+          --env-name bench_win_bash -r 1 -n 10 --no-plot
+
   test_no_pytest:
     name: Test benchopt test without pytest installed
     runs-on: ubuntu-latest
@@ -176,16 +216,18 @@ jobs:
 
   report_test:
     if: ${{ always() }}
-    needs: [test_benchopt, test_no_conda]
+    needs: [test_benchopt, test_no_conda, test_windows_bash]
     runs-on: ubuntu-latest
     env:
       TEST_BENCHOPT: ${{ needs.test_benchopt.result }}
       TEST_NO_CONDA: ${{ needs.test_no_conda.result }}
+      TEST_WINDOWS_BASH: ${{ needs.test_windows_bash.result }}
 
     steps:
       - name: "Gather test results."
         run: |
-          if [[ $TEST_BENCHOPT == 'success' && $TEST_NO_CONDA == 'success' ]]; then
+          if [[ $TEST_BENCHOPT == 'success' && $TEST_NO_CONDA == 'success' \
+                && $TEST_WINDOWS_BASH == 'success' ]]; then
             exit 0;
           else
             exit 1;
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -126,6 +126,46 @@ jobs:
       - name: Run benchopt run without conda
         run: benchopt run examples/minimal_benchmark -r 1 -n 10 --no-plot
 
+  test_windows_bash:
+    # Regression test for shell handling on Windows when the CI step runs under
+    # a bash shell (the common `shell: bash` setup in downstream benchmarks).
+    # In that case $SHELL is a git-bash path, which used to make benchopt emit a
+    # bash script containing `CALL conda` and break conda env management. See
+    # https://github.com/benchopt/benchopt for the fix. Unlike the `cmd`-based
+    # `test_benchopt` Windows leg, this exercises the bash-shell code path.
```
tomMoral: Agent tends to repeat there goal everywhere, shorten to just the useful info here
```suggestion
    # Regression test for shell handling on Windows when using a bash shell
    # instead of cmd.
```

.github/workflows/test.yml
```diff
@@ -126,6 +126,46 @@ jobs:
...
+    name: Test Windows under bash shell
```
tomMoral:
```suggestion
    name: Test Windows with bash shell
```

benchopt/utils/conda_env_cmd.py
```diff
@@ -14,12 +14,25 @@
...
 
 SHELL = get_setting('shell')
-CONDA_CMD = get_setting('conda_cmd')
 
-# On windows, calling conda without call exit the cmd script:
-# https://github.com/conda/conda/issues/12418
-if sys.platform == 'win32' and not CONDA_CMD.lower().startswith('call'):
-    CONDA_CMD = f"CALL {CONDA_CMD}"
+
+def _conda_cmd(conda_cmd, is_cmd):
```
tomMoral: Why make a function here? This complexity the code for no gain IMO.
Just change the if statement, replacing `if sys.platform ...` with `IS_CMD` and change the comment from `On windows` to `In CMD`

benchopt/utils/shell_cmd.py
```diff
@@ -12,6 +14,23 @@
...
 
 
+def _split_shell(shell):
+    """Split a shell setting into its program and arguments.
+
+    The `shell` setting carries arguments by design (e.g. ``'cmd /c'`` or
+    ``'bash --norc --noprofile'``), so it cannot be quoted as a whole. Split it
+    into a list suitable for `subprocess.run` without an extra `shell=True`
+    layer, which also handles a shell *path* that contains spaces (e.g. Git
+    Bash on Windows).
```
tomMoral:
```suggestion
    ``'bash --norc --noprofile'``), so it cannot be quoted as a whole.
```

benchopt/utils/shell_cmd.py
```diff
@@ -72,13 +91,30 @@ def _run_shell(script, raise_on_error=None, capture_stdout=True,
     if raise_on_error is True:
         raise_on_error = "{output}"
 
-    command = f'{SHELL} "{tmp.name}"'
+    # Invoke the script directly with the configured shell, without going
+    # through an extra `shell=True` layer that would re-parse the command line.
```
tomMoral: I don't understand this comment, there is no shell=True at all here? Feels like this is an iterative agent comment (it tends to document paste failures from the PR tests).

benchopt/config.py
```diff
@@ -49,8 +52,12 @@
...
   not directly installed on ``$PATH``. This can also be used to use ``mamba``
   to install benchmarks instead of conda. See :ref:`config_mamba`.
-* ``shell``, *str*: can be used to specify the shell to use. Default to
-  `SHELL` from env if it exists and ``'bash'`` otherwise.
+* ``shell``, *str*: can be used to specify the shell to use. On POSIX systems
+  it defaults to the ``SHELL`` env var if it exists and
+  ``'bash --norc --noprofile'`` otherwise. On Windows it defaults to
+  ``'cmd /c'`` and the ``SHELL`` env var is ignored (a POSIX ``SHELL`` set by
+  Git Bash or MSYS would otherwise break conda handling); override it
+  explicitly with the ``BENCHOPT_SHELL`` env var or this setting.
```
tomMoral:
```suggestion
  ``'cmd /c'`` and the ``SHELL`` env var is ignored as it is not reliable.
```

benchopt/utils/shell_cmd.py
```diff
@@ -72,13 +91,30 @@ def _run_shell(script, raise_on_error=None, capture_stdout=True,
...
+    command = [*_split_shell(SHELL), tmp.name]
 
     if capture_stdout:
-        exit_code, output = subprocess.getstatusoutput(command)
+        # Merge stderr into stdout (preserving order, as getstatusoutput did)
+        # so callers parsing the last output line as JSON keep working.
```
tomMoral:
```suggestion
        # Merge stderr into stdout, and strip empty line
        # so callers parsing the last output line as JSON keep working.
```

doc/whats_new.rst
```diff
@@ -68,6 +68,14 @@ TST
 FIX
 ~~~
 
+- Fix shell handling on Windows so ``benchopt install``/``test`` work with a
+  bash-like ``SHELL`` (e.g. GitHub ``shell: bash`` runners). The shell is now
+  invoked as an argument list (handling spaces in the shell path), the conda
+  ``CALL`` prefix is kept consistent with the shell benchopt generates scripts
+  for, on Windows the ``SHELL`` env var no longer overrides the ``cmd /c``
+  default (use ``BENCHOPT_SHELL`` to override), and the underlying error is now
+  surfaced instead of swallowed. By `Johan Larsson`_
```
tomMoral: A bit too technical for a change log, just shortened it a bit.
```suggestion
- Fix shell handling on Windows so ``benchopt install``/``test`` work with a
  bash-like shell (e.g. GitHub ``shell: bash`` runners).
  On Windows the ``SHELL`` env var is now ignored, use ``BENCHOPT_SHELL``
  to override it. Underlying shell error are also better reported,
  instead of being swallowed. By `Johan Larsson`_
```

.github/workflows/test.yml
```diff
@@ -126,6 +126,46 @@ jobs:
...
+    runs-on: windows-latest
+    defaults:
+      run:
+        shell: bash -el {0}
+
+    steps:
```
tomMoral, line 139: What I don't understand here is that it feels like benchopt will revert to using the default cmd here instead of the current bash no?
As we ignore SHELL and don't set BENCHOPT_SHELL, we default to the `cmd /c`?

tomMoral (commented):
Thanks for the PR.
This looks mostly good to me.

My main question is about the need to ingore SHELL and the fact that the test seems to run in cmd instead of trying to run on the bash script benchopt is invoked in?

benchopt/utils/conda_env_cmd.py
jolars: You're right, I've trimmed it now.

doc/whats_new.rst
jolars: Shortened it and also added a PR ref.

.github/workflows/test.yml
jolars: Yes, that is the intention now. The problem with the current state is that benchopt honors `SHELL` and tries to run conda in bash on windows, which is typically not a great idea.

benchopt/utils/shell_cmd.py
jolars: You're right, sorry, fixed it now.

.github/workflows/test.yml
tomMoral: Ok got it! We force using CMD then.
~~This should be made explicit.~~
This is explicit above. 😁 Thx

benchopt/config.py
```diff
@@ -49,8 +52,10 @@
 * ``conda_cmd``, *str*: can be used to give the path to ``conda`` if it is
   not directly installed on ``$PATH``. This can also be used to use ``mamba``
   to install benchmarks instead of conda. See :ref:`config_mamba`.
-* ``shell``, *str*: can be used to specify the shell to use. Default to
-  `SHELL` from env if it exists and ``'bash'`` otherwise.
+* ``shell``, *str*: can be used to specify the shell to use. On POSIX systems
+  it defaults to the ``SHELL`` env var if it exists and
+  ``'bash --norc --noprofile'`` otherwise. On Windows it defaults to
+  ``'cmd /c'`` and the ``SHELL`` env var is ignored as it is not reliable.
```
tomMoral:
```suggestion
  ``'cmd /c'``. The ``SHELL`` env var is ignored as it is a less reliable default.
```

tomMoral (approved):

benchopt/config.py
```diff
@@ -49,8 +52,10 @@
 * ``conda_cmd``, *str*: can be used to give the path to ``conda`` if it is
   not directly installed on ``$PATH``. This can also be used to use ``mamba``
   to install benchmarks instead of conda. See :ref:`config_mamba`.
-* ``shell``, *str*: can be used to specify the shell to use. Default to
-  `SHELL` from env if it exists and ``'bash'`` otherwise.
+* ``shell``, *str*: can be used to specify the shell to use. On POSIX systems
+  it defaults to the ``SHELL`` env var if it exists and
+  ``'bash --norc --noprofile'`` otherwise. On Windows it defaults to
+  ``'cmd /c'``. The ``SHELL`` env var is ignored as it is a less reliable default.
```
tomMoral:
```suggestion
  ``'cmd /c'``. The ``SHELL`` env var is ignored as it is an unreliable default.
```

tomMoral (approved):

benchopt/config.py
```diff
@@ -49,8 +52,10 @@
 * ``conda_cmd``, *str*: can be used to give the path to ``conda`` if it is
   not directly installed on ``$PATH``. This can also be used to use ``mamba``
   to install benchmarks instead of conda. See :ref:`config_mamba`.
-* ``shell``, *str*: can be used to specify the shell to use. Default to
-  `SHELL` from env if it exists and ``'bash'`` otherwise.
+* ``shell``, *str*: can be used to specify the shell to use. On POSIX systems
+  it defaults to the ``SHELL`` env var if it exists and
+  ``'bash --norc --noprofile'`` otherwise. On Windows it defaults to
+  ``'cmd /c'``. The ``SHELL`` env var is ignored as it is an unreliable default.
```
tomMoral:
```suggestion
  ``'cmd /c'``. The ``SHELL`` env var is ignored as it is an less reliable
  default (running conda with it often fails).
```

tomMoral (approved):

PR #975, opened 2026-07-06
FIX table column size and flickering

etyl:
### Checks before merging PR
- [x] ~~added unit test~~
- [x] ~~edited the [what's new](../../whatsnew.rst) (if applicable)~~

### DIFF
--- a/benchopt/plotting/html/static/result.js
+++ b/benchopt/plotting/html/static/result.js
@@ -1306,6 +1306,12 @@ function renderTable() {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


tomMoral:
What does this PR fix? Add some info so it can be reviewed, otherwise it is hard to assess the code validity.

etyl:
> What does this PR fix? Add some info so it can be reviewed, otherwise it is hard to assess the code validity.

When hiding / unhiding columns in the table, the gridjs table was not regenerated, so the columns had the wrong width. The dropdown options were also not correctly plugged to the table.

PR #978, opened 2026-07-10
FIX Detect current-env installs by reloading the classes

felixdivo:
## Problem

`benchopt install` into the **current** env (no `--env-name`) does not detect the packages it just
installed, and exits with code 1 even though the install succeeded. Running the same command a
second time works (`benchopt install -y || benchopt install -y`).

Collecting the benchmark imports every module once and caches the outcome, either on the class
(`import_ctx.failed_import`) or as a `FailedImport` class. The post-install check in
`install_all_requirements` calls `is_installed()` again, but for `env_name is None` this only
re-reads that cached failure and never re-imports. Classes whose requirements come from the global
`Objective.requirements` then stay in `missings`, and `check_missing()` fails the command. The
`env_name is not None` path is unaffected, as it already re-imports the class in a fresh process.

## Fix

When installing into the **current** env, a class whose requirements were missing at collection
time is a `FailedImport`. `is_installed()` gains a `reload` argument, used for the post-install
checks: `FailedImport.is_installed(reload=True)` invalidates the import caches, re-imports the
class, and delegates to the freshly imported one. Delegating is necessary, as the stale class
closes over the globals of the first, failed import. A class that imported fine needs no check —
`is_installed(env_name=None)` returns `True`.

Modules whose import failed (including under a `safe_import_context`) are dropped from
`sys.modules` at load time in `_load_class_from_module`, so re-importing picks up the packages
installed in the meantime.

One related fix: `JuliaSolver.is_installed` forwards `**kwargs`, as its fixed signature would
otherwise raise `TypeError` on the new argument.

## Note

An earlier version of this PR ran the check in a `benchopt check-install` subprocess; it was
replaced by the in-process reload, which is much faster (~80s → ~0.1s for the test above).

### Checks before merging PR
- [x] added unit test


felixdivo:
This surfaced when I had to add `benchopt install -y || benchopt install -y` in my Docker environment as part of https://github.com/benchopt/benchmark_tsfm/pull/53.

I'm a bit confused that this is only a problem now, so maybe I'm also using it wrong.

tomMoral:
I would rather reload the solver class, adding a reload function, that remove the dynamic module and retry loading it instead of spinning a full new subprocess (this tends to be super slow).

felixdivo:
@tomMoral Sure, better now? :)

tomMoral (commented):
I feel this PR is a bit too complex for the feature and could be simplified:

- The only way the class is not loaded should be when we replace it with the FailedImport klass. So the reload mechanism should probably be in this class instead of being in the dependencymixin. For the class where the install was ok, it should not be a check at all.
- when a module fails to import, there is no module put in sys.module, so I think we could avoid this (to check with safe_import_context, but this could be handled by replacing the class at load time by the failed import class too).
- ignore_cache is a bit misleading, I would call it reload.

Make sense?

felixdivo:
Fair point! I updated the commit, PR title, and description.

benchopt/cli/tests/test_cmd_install.py
```diff
@@ -303,6 +304,48 @@ class Dataset(TempDataset):
...
+    def test_install_in_current_env_detects_fresh_package(
+            self, test_env_name, uninstall_dummy_package
+    ):
+        # Regression test: installing in the *current* environment must
+        # detect a package installed during the very same `benchopt install`
+        # call. The post-install check used to reuse the import outcome cached
+        # at collection time (before the install), so freshly installed
+        # packages looked missing and the command exited with an error.
+        # We reproduce the current-env path (``env_name=None``) in isolation by
+        # driving `benchopt install` from *inside* the throwaway test env,
```
tomMoral:
```suggestion
        # Check that, when installing in the *current* environment,
        # the install detection mechanism correctly detects the post-install status.
```

benchopt/cli/tests/test_cmd_install.py
```diff
@@ -303,6 +304,48 @@ class Dataset(TempDataset):
...
+        # where ``env_name=None`` resolves to that env.
```
tomMoral:
```suggestion
```

benchopt/cli/tests/test_cmd_install.py
```diff
@@ -303,6 +303,44 @@ class Dataset(TempDataset):
...
+    def test_install_in_current_env_detects_fresh_package(
+            self, uninstall_dummy_package_current_env
+    ):
+        # Regression test: installing in the *current* environment must
+        # detect a package installed during the very same `benchopt install`
+        # call. The post-install check used to reuse the import outcome cached
+        # at collection time (before the install), so freshly installed
+        # packages looked missing and the command exited with an error.
+        # Run `install` directly (no --env-name), so it targets the current
+        # (this pytest process's) env, reproducing the bug in-process.
```
tomMoral:
```suggestion
        # Check that, when installing in the *current* environment, the
        # install detection mechanism correctly detects the post-install status.
```

benchopt/tests/fixtures.py
```diff
@@ -340,6 +340,22 @@ def uninstall_dummy_package(test_env_name):
     )
 
 
+@pytest.fixture(scope='function')
+def uninstall_dummy_package_current_env():
+    # Like `uninstall_dummy_package`, but targets the current (pytest) env
+    # instead of the throwaway `test_env_name` conda env, and also drops the
+    # module from sys.modules. Does not depend on `test_env_name`/`use_env`,
+    # so it isn't skipped by --skip-env: tests using it exercise the
+    # current-env install path (env_name=None) and need no conda env.
```
tomMoral:
```suggestion
    # Like `uninstall_dummy_package`, but targets the current env.
    # Also drops the module from sys.modules. 
```

benchopt/utils/dynamic_modules.py
```diff
@@ -116,8 +117,67 @@ def _load_class_from_module(benchmark_dir, module_filename, class_name):
...
+        if klass._import_ctx.failed_import:
+            # The module executed fully: the failure was caught by a
+            # safe_import_context, so klass already carries its real
+            # attributes, including ones computed dynamically -- which is
+            # exactly why safe_import_context is used in the first place,
+            # as those cannot be recovered by the AST-based fallback below.
+            # Evict the module so the next load re-executes it and can pick
+            # up packages installed in the meantime, and build the
+            # FailedImport class directly from klass to keep those real
+            # attributes instead of falling back to AST parsing.
```
tomMoral:
```suggestion
            # The module failed to import, but was protected with safe_import_context.
            # It can have dynamic attributes not parsable with ast, so build
            # the FailedImport class directly from klass to keep them.
            # Also evict the module so the next load re-executes it.
```

benchopt/utils/dynamic_modules.py
```diff
@@ -140,13 +200,28 @@ class klass(base_cls, ParametrizedNameMixin, DependenciesMixin,
...
                         env_name=env_name,
                         raise_on_not_installed=raise_on_not_installed,
                         **kwargs
                     )
+                if reload:
+                    # The module was never cached in the first place (it
+                    # failed before being fully executed), so loading it
+                    # again re-attempts the import. Invalidate the import
+                    # caches first, to make packages installed since the
+                    # interpreter started visible.
```
tomMoral:
```suggestion
                    # Invalidate the import caches first, to make packages installed
                    # since the interpreter started visible.
```

benchopt/utils/dynamic_modules.py
```diff
@@ -59,6 +59,54 @@ def __repr__(cls):
...
+            return super().is_installed(
+                env_name=env_name,
+                raise_on_not_installed=raise_on_not_installed,
+                **kwargs
+            )
+        if reload:
+            # The module was evicted from the cache (or never cached) when
+            # this class was built, so loading it again re-attempts the
+            # import. Invalidate the import caches first, to make packages
+            # installed since the interpreter started visible.
```
tomMoral:
```suggestion
            # Checking after an install step. Invalidate the import caches first,
            # to make package installed since the interpreter started visible.
```

benchopt/utils/dynamic_modules.py
```diff
@@ -116,50 +164,51 @@ def _load_class_from_module(benchmark_dir, module_filename, class_name):
...
+            exc_type, value, tb = klass._import_ctx.import_error
+            tb_to_print = ''.join(
+                traceback.format_exception(exc_type, value, tb, chain=False)
+            )
+
+            class klass(_FailedImportMixin, klass, metaclass=FailedImport):
+                "Object for the class list that raises error if used."
+
+                _exc = value.with_traceback(tb)
+                _error_displayed = False
```
tomMoral:
```suggestion
```

benchopt/utils/dynamic_modules.py
```diff
@@ -116,50 +164,51 @@ def _load_class_from_module(benchmark_dir, module_filename, class_name):
...
 
-        class klass(base_cls, ParametrizedNameMixin, DependenciesMixin,
-                    metaclass=FailedImport):
+        # base_cls already inherits ParametrizedNameMixin and
+        # DependenciesMixin, no need to list them again.
+        class klass(_FailedImportMixin, base_cls, metaclass=FailedImport):
             "Object for the class list that raises error if used."
 
             _exc = e
             _error_displayed = False
```
tomMoral:
```suggestion
```

benchopt/utils/dynamic_modules.py
```diff
@@ -59,6 +59,54 @@ def __repr__(cls):
...
+class _FailedImportMixin:
+    """Shared ``is_installed`` for classes representing a failed import.
+
+    Must come first in the MRO of classes using it, so it takes precedence
+    over any ``is_installed`` defined on the wrapped/base class. Relies on
+    ``cls_name``, ``_exc``, ``_error_displayed`` and ``_tb_to_print`` being
+    set by the class using this mixin, and on ``_module_filename``,
+    ``_benchmark_dir`` and ``_base_class_name``, set by
+    ``_load_class_from_module`` once the class is built.
+    """
```
tomMoral, line 71:
```suggestion
    """

    _error_displayed = False

```

benchopt/utils/dynamic_modules.py
```diff
@@ -59,6 +59,54 @@ def __repr__(cls):
         )
 
 
+class _FailedImportMixin:
+    """Shared ``is_installed`` for classes representing a failed import.
+
+    Must come first in the MRO of classes using it, so it takes precedence
+    over any ``is_installed`` defined on the wrapped/base class. Relies on
+    ``cls_name``, ``_exc``, ``_error_displayed`` and ``_tb_to_print`` being
```
tomMoral:
```suggestion
    ``cls_name``, ``_exc``, and ``_tb_to_print`` being
```

benchopt/utils/dynamic_modules.py
```diff
@@ -116,50 +164,51 @@ def _load_class_from_module(benchmark_dir, module_filename, class_name):
...
         from ..plotting.base import BasePlot
         base_cls = dict(
             Solver=BaseSolver, Dataset=BaseDataset,
             Objective=BaseObjective, Plot=BasePlot
         )[class_name]
 
-        class klass(base_cls, ParametrizedNameMixin, DependenciesMixin,
-                    metaclass=FailedImport):
+        # base_cls already inherits ParametrizedNameMixin and
+        # DependenciesMixin, no need to list them again.
```
tomMoral:
```suggestion
```

benchopt/cli/tests/test_cmd_install.py
```diff
@@ -303,6 +303,48 @@ class Dataset(TempDataset):
...
+            install(
+                f'{bench.benchmark_dir} -d test-dataset -y'.split(),
+                'benchopt', standalone_mode=False
+            )
+
+        out.check_output("not importable", repetition=0)
+        # Also check there is no warning about a package left uninstalled
+        # (the objective itself goes through the same post-install reload
+        # check, via `check_installs` rather than `missings`).
+        out.check_output("missing deps", repetition=0)
```
tomMoral:
```suggestion
        # Check that install reports for both direct install (no "missing deps")
        # and indirect ones (no "not importable") have correct status.
        out.check_output("missing deps", repetition=0)
        out.check_output("not importable", repetition=0)
```

tomMoral (approved):
A few notes on the comments.
I did a small pass on the mechanisms but other wise LGTM!!

Thx @felixdivo

PR #984, opened 2026-07-31
FIX get_seed() in Dataset.get_data() for parallel runs

tomMoral:
A `Dataset` that calls `get_seed()` from `get_data()` (the documented reproducible-per-repetition pattern) crashes with `BrokenProcessPool` as soon as `-j` is set above 1. `Objective.__setstate__` was eagerly calling `_set_dataset` -- and thus `get_data()`/`get_seed()` -- during unpickling in the worker, before `run_one_to_cvg` had re-attached the run context.

This PR defers that call to `run_one_to_cvg`, right after the context is re-attached, mirroring how `solver._set_objective` is already called explicitly in the body rather than from `__setstate__`.

### Checks before merging PR
- [x] ~~added documentation for any new feature~~ (not needed)
- [x] added unit test


benchopt/runner.py
```diff
@@ -128,6 +128,12 @@ def run_one_to_cvg(benchmark, objective, solver, meta, timeout, max_runs,
 
     with exception_handler(terminal, pdb=pdb) as ctx:
 
+        # A dataset is always attached by `_generate_runs.py` before this is
+        # called; `_set_dataset` is a no-op if it already ran (see its body).
```
```suggestion
        # Set dataset and objective. Note that _set_dataset is a no-op
        # if it has already been called (sequential runs)
```

benchopt/tests/test_runner.py
```diff
@@ -503,8 +503,32 @@ def test_objective_simple(self, no_debug_log):
                     run(cmd_str.split(),
                         standalone_mode=False)
 
-        seeds = out.check_output(r"(?m)^#SEED-obj=.*", repetition=2)
-        assert seeds[0] == seeds[1], "Seeds should be equal"
+                # Also check that evaluate_result's get_seed() survives real
+                # parallel dispatch (regression test, see run_one_to_cvg).
```
tomMoral:
```suggestion
                # Check that get_seed() works with parallel runs.
```

benchopt/tests/test_runner.py
```diff
@@ -503,8 +503,32 @@ def test_objective_simple(self, no_debug_log):
...
+        ) as bench:
+            with CaptureCmdOutput() as out:
+                cmd_str = f"{bench.benchmark_dir} --no-cache "
+                cmd_str += "--no-plot"
+                for it in range(2):
+                    run(cmd_str.split(),
+                        standalone_mode=False)
+
+                # Also check that set_objective's get_seed() survives real
+                # parallel dispatch (regression test, see run_one_to_cvg).
```
tomMoral:
```suggestion
                # Check that get_seed() works with parallel runs.
```

benchopt/tests/test_runner.py
```diff
@@ -519,8 +543,16 @@ def test_dataset_simple(self, no_debug_log):
                     run(cmd_str.split(),
                         standalone_mode=False)
 
-        seeds = out.check_output(r"(?m)^#SEED-data=.*", repetition=2)
-        assert seeds[0] == seeds[1], "Seeds are not equal"
+                # Also check that the dataset's get_seed() survives real
+                # parallel dispatch, where it is reconstructed in the worker
+                # process (regression test, see run_one_to_cvg).
```
tomMoral:
```suggestion
                # Check that get_seed() works with parallel runs.
```

benchopt/tests/test_runner.py
```diff
@@ -519,8 +543,16 @@ def test_dataset_simple(self, no_debug_log):
...
+                run((cmd_str + " -j 2").split(), standalone_mode=False)
+
+        # get_data() runs once per -j1 call, and twice for the -j2 call (once
+        # in the main process, once more in the worker) -- see
+        # test_no_get_seed_no_extra_reload for the same pattern.
```
tomMoral:
```suggestion
        # get_data() runs once per -j1 call, and twice for the -j2 call (once
        # during dispatch, once more in the worker).
```

doc/whats_new.rst
```diff
@@ -78,6 +78,11 @@ TST
 FIX
 ~~~
 
+- Fix ``get_seed`` raising in a ``Dataset``'s ``get_data`` when running with
+  ``-j`` greater than 1: the run context was re-attached too late, after the
+  dataset had already been reconstructed and its ``get_data`` called in the
+  worker process. By `Thomas Moreau`_ (:gh:`984`)
```
```suggestion
  ``-j`` greater than 1. By `Thomas Moreau`_ (:gh:`984`)
```

PR #990, opened 2026-08-10
ENH add `benchopt info -f` to summarize result files

tomMoral:
## Summary
- `benchopt info -f <result_file>` summarizes a result file (rows, configs,
  repetitions, objective/solver/dataset names, `objective_<name>` columns,
  run date) instead of listing the benchmark's solvers/datasets. `-f all`
  summarizes every result file in the output folder, mirroring `plot -f all`.
- Plain `benchopt info` now also lists the result files available in the
  benchmark's output folder.
- Extracted the summary logic into `describe_results`, moved alongside
  `merge_results` into `benchopt/results/result_processing.py` (renamed from
  `process.py`), so it's usable programmatically and not just from the CLI.

### Checks before merging PR
- [x] added unit test
- [x] edited the [what's new](../../whatsnew.rst)


benchopt/cli/helpers.py
```diff
@@ -321,12 +323,30 @@ def print_info(cls_name_list, cls_list, env_name=None, verbose=False):
               is_flag=True,
               help="If used, list solver/dataset "
               "parameters, dependencies and availability.")
+@click.option('--filename', '-f', 'result_filename', type=str, default=None,
```
tomMoral: It should be like the dataset and silver: multiple, and when not used, just a summary of all results avail
Also should be before envname

benchopt/cli/helpers.py
```diff
@@ -384,6 +404,57 @@ def info(benchmark, solver_names, dataset_names, env_name='False',
         print_info(solver_names, all_solvers, env_name, verbose)
 
 
+def _print_available_result_files(benchmark):
```
tomMoral: Would move all print helpers together. Also would rename print_info to match.

benchopt/cli/helpers.py
```diff
@@ -384,6 +404,57 @@ def info(benchmark, solver_names, dataset_names, env_name='False',
...
+    """Print the result files available in the benchmark's output folder."""
+    output_folder = benchmark.get_output_folder()
+    result_files = sorted(
```
tomMoral: Isn't this parser factorize with other result management tools? For instance `benchmark.get_result_file('all')`?

benchopt/cli/helpers.py
```diff
@@ -279,6 +282,53 @@ def print_info(cls_name_list, cls_list, env_name=None, verbose=False):
...
+    print(f"Configs (objective x solver x dataset): {summary['n_configs']}")
+
+    if summary['n_repetitions'] is not None:
+        print(f"Repetitions: {summary['n_repetitions']}")
+
+    for label, key in [
+        ("Objectives", 'objectives'),
+        ("Solvers", 'solvers'),
+        ("Datasets", 'datasets'),
+    ]:
```
tomMoral: Label can be obtained through capitalize

benchopt/cli/helpers.py
```diff
@@ -321,12 +382,25 @@ def print_info(cls_name_list, cls_list, env_name=None, verbose=False):
...
+        )
+        result_files = benchmark.get_result_files(filenames)
+        for i, result_file in enumerate(result_files):
+            if i > 0:
+                print()
+            _print_result_file_summary(result_file)
+        return
+
     print(f"Info regarding the benchmark '{benchmark.name}'")
+    _print_available_result_files(benchmark)
```
tomMoral: Should be printed at the end of the summary

PR #993, opened 2026-08-11
VIZ format table cells with markers

etyl:
Table cells can now be formatted with some simple markers (bold, italic and underlines), works with text and numbers. This is useful when we want to highlight the best results in a table for example.

### Checks before merging PR
- [X] added documentation for any new feature
- [X] ~~added unit test~~
- [X] edited the [what's new](../../whatsnew.rst) (if applicable)

### DIFF
--- a/benchopt/plotting/html/static/result.js
+++ b/benchopt/plotting/html/static/result.js
@@ -1211,6 +1211,74 @@ const valueToFixed = (value) => {
   return value;
 }
 
+/*
+ * Inline markup for table cells: `**bold**`, `*italic*` and `__underlined__`.
+ * They are rendered as HTML in the report and converted to the matching LaTeX
+ * commands on export. A cell whose text is a single number, e.g. `**1.234**`,
+ * is still handled as a number by the float precision control and the sorting.
+ */
+const CELL_MARKUP = [
+  [/\*\*([^*]+)\*\*/g, 'b'],
+  [/__([^_]+)__/g, 'u'],
+  [/\*([^*]+)\*/g, 'i'],
+];
+const LATEX_COMMANDS = {B: 'textbf', I: 'textit', U: 'underline'};
+
+const escapeHtml = (value) => value.replace(
+  /[&<>]/g, c => ({'&': '&amp;', '<': '&lt;', '>': '&gt;'}[c])
+);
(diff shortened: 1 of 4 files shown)


etyl:
An example:
image

Generated with:
```python
from benchopt import BasePlot


class Plot(BasePlot):
    name = "Custom table"
    type = "table"
    options = {
        "dataset": ...,
    }

    def plot(self, df, dataset):
        df = df[df['dataset_name'] == dataset]
        # Last value reached by each solver, median over the repetitions
        results = (
            df.select_dtypes(include=['number'])
            .join(df['solver_name'])
            .groupby(['solver_name', 'stop_val'])
            .median()
            .groupby('solver_name')
            .last()
        )
        best = results['objective_value'].idxmin()

        return [
            [
                f"**{solver}**" if solver == best else solver,
                f"**{res['objective_value']}**" if solver == best else res['objective_value'],
                res['time'],
                "__best objective__" if solver == best else r"*runner-up\**",
            ]
            for solver, res in results.iterrows()
        ]

    def get_metadata(self, df, dataset):
        return {
            "title": f"Custom table\nData: {dataset}",
            "columns": ["solver", "objective_value", "time (s)", "comment"],
            "default_order_column": "objective_value",
            "default_order_ascending": True,
        }
```

etyl:
What do you think @tomMoral ?

tomMoral (commented):
Over all, a cool feature to add.

I am a bit scared of having to develop our own parser, which will have to cope with various edge case (for instance right now, you can't have a * in the middle of the bold/italic text, no _ in the underline), instead of relying on a markdown parser? The advantage would ba that we can simply say we support markdown formater instead of having our own list of supported ones.

But there is a tradeoff with the complexity while we relying only on few features, so not sure it is worth it. Maybe just take a look at markdown parser to see if there is any very light one, and else, keep this one?

etyl:
The idea was to have a very simple parser to change the style of a cell, and not make complicated styling in a single cell (so only bold, italic or underlined exclusively). Also it is currently possible to use * and _ inside cells using \ (as shown in the example). I have the impression that using a parser for styling the cells would maybe be a bit overkill for the current task but I may be wrong.

Issue #571, opened 2023-07-03
Add a "tag" member to solvers (allow filtering solvers)


svaiter:
Looking at the benchmark_lasso, there is numerous implementation of some algorithms, such as PGD. The idea would be to be able to filter the solver according to some tag (e.g. "pgd_solver").

In term of user interface, a search field could be added at the top left.

The implementation could be as simple as a list of string for each Solver.
Example 
celer : ["proximal_cd", "working_set", "python"]
julia_pgd : ["proximal_gd", "julia"]

Issue #731, opened 2024-06-04
ENH add a require_gpu parameter to create solver-specific slurm configs


pbarbarant:
When only a subset of solvers require to be ran on GPUs, the user is required to use the `-s` flag along with custom slurm configs. As discussed with @tomMoral, it would be nicer to have a `require_gpu` option per solver to automatically generate solver-specific slurm configs.

PR #994, opened 2026-08-17
FEAT Add solver and dataset tag filters

Jad-yehya:
I think it is a good idea for solvers and datasets to have tags and use them to select components when running a benchmark.

Solvers and datasets declare tags as class-level lists:

```python
class Solver(BaseSolver):
    tags = ["gpu", "transformer"]
```
Components can then be selected with:
```
$ benchopt run . -st gpu -dt easy
```
The long forms versions are `--solver-tag` and `--dataset-tag`.

I see this being used for example in benchmarks of foundation models where you want to evaluate on task-specific datasets, instead of selecting each dataset by hand.

(I'll add the what's new if this is accepted)

### AI use : 

I was assisted by Codex for this PR.

### Checks before merging PR
- [x] added unit test

### DIFF
--- a/benchopt/benchmark.py
+++ b/benchopt/benchmark.py
@@ -77,12 +77,10 @@ def _get_class_tags(cls):
 
 def _filter_classes_by_tags(all_classes, tags, name_type):
     """Return classes matching any requested tag."""
-    if tags is None:
+    if not tags:
         return set(all_classes)
     if isinstance(tags, str):
         tags = [tags]
-    if not tags:
-        return set(all_classes)
 
     class_tags = {cls: _get_class_tags(cls) for cls in all_classes}
     available_tags = set().union(*class_tags.values())
--- a/benchopt/benchmark.py
+++ b/benchopt/benchmark.py
@@ -67,10 +67,10 @@ def _get_class_tags(cls):
         ) from exc
 
     if not isinstance(tags, list) or not all(
-            isinstance(tag, str) for tag in tags):
+            isinstance(tag, str) and ',' not in tag for tag in tags):
         raise ValueError(
             f"Invalid tags for {component} {cls.name!r}: expected a literal "
-            "list of strings."
+            "list of strings without commas."
         )
     return tags.copy()
 
--- a/benchopt/benchmark.py
+++ b/benchopt/benchmark.py
@@ -55,6 +55,55 @@ def get_running_benchmark():
     return _RUNNING_BENCHMARK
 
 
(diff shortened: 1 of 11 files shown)


etyl:
Very useful feature
Fixes #571 and #731

benchopt/cli/main.py
```diff
@@ -109,6 +114,16 @@ def _get_run_args(cli_kwargs, config_file_kwargs):
...
+@click.option('--solver-tag', '-st',
+              metavar="<tag>", multiple=True, type=str,
+              help="Only include solvers with <tag>. Repeat this option to "
+              "match any listed solver tag.",
+              shell_complete=complete_solver_tags)
+@click.option('--dataset-tag', '-dt',
+              metavar="<tag>", multiple=True, type=str,
+              help="Only include datasets with <tag>. Repeat this option to "
+              "match any listed dataset tag.",
+              shell_complete=complete_dataset_tags)
```
etyl, line 132: I could maybe be more practical if it were possible to put multiple tags directly using `,` instead of using multiple flags (however you would need to check that the tags don't contain any `,`)

benchopt/benchmark.py
```diff
@@ -55,6 +55,55 @@ def get_running_benchmark():
...
+        return set(all_classes)
+    if isinstance(tags, str):
+        tags = [tags]
+    if not tags:
+        return set(all_classes)
+
+    class_tags = {cls: _get_class_tags(cls) for cls in all_classes}
+    available_tags = {
+        tag for cls_tags in class_tags.values() for tag in cls_tags
+    }
```
etyl: this seems clearer to me but maybe not
```suggestion
    available_tags = set().union(*class_tags.values())
```

benchopt/benchmark.py
```diff
@@ -55,6 +55,55 @@ def get_running_benchmark():
...
+
+
+def _filter_classes_by_tags(all_classes, tags, name_type):
+    """Return classes matching any requested tag."""
+    if tags is None:
+        return set(all_classes)
+    if isinstance(tags, str):
+        tags = [tags]
+    if not tags:
+        return set(all_classes)
```
etyl:
```suggestion
    if not tags:
        return set(all_classes)
    if isinstance(tags, str):
        tags = [tags]
```

benchopt/cli/main.py
Jad-yehya: yes I think that is a better idea. I don't think restricting `,` in the tag names is too constraining for the user.

PR #1009, opened 2026-09-21
ENH add `benchopt info --version` to check min_benchopt_version compatibility

tomMoral:
## Motivation

The reusable benchmark CI workflow (`template_benchmark`) currently gates whether to run `benchopt test` by comparing versions in bash:

```bash
if [[ "$benchopt_version" < "$min_version" ]]
```

`<` in `[[ ]]` is a **lexicographic** string comparison, so once benchopt reaches **1.10.0** it is (wrongly) considered *older* than `1.2`–`1.9`. Since the `Test` step is guarded by `if: compatible == 'true'`, affected benchmarks would **silently skip their tests** (green, not red).

This PR moves the check into benchopt so it is done with a real version comparison, and the workflow can call a single command instead of re-implementing the logic.

## Change

Add `benchopt info <benchmark> --version`:

- prints the installed benchopt version and the benchmark's `min_benchopt_version`;
- exits `0` if compatible, non-zero (clean `Error:`, no traceback) if the installed benchopt is too old;
- uses the same `packaging.version` comparison as the runtime check in `benchopt run` (dev suffixes stripped via `base_version`).

Plain `benchopt info` is unchanged (still purely informational). The companion `template_benchmark` PR replaces the bash comparison with a call to this command.

## Test

`benchopt/cli/tests/test_cmd_info.py`: compatible (exit 0) and incompatible (raises) cases via `temp_benchmark`.


doc/whats_new.rst
```diff
@@ -17,20 +17,15 @@ Version 1.10.0 -- in development
...
-  By `Thomas Moreau`_ (:gh:`1003`)
-
 - Add ``benchopt info -f <result_file>`` (repeatable, or ``-f all``) to
   summarize result file(s) instead of listing benchmark solvers/datasets;
   plain ``benchopt info`` now also lists available result files.
   By `Thomas Moreau`_ (:gh:`990`)
 
+- Add ``benchopt info <benchmark> --version`` to check that the installed
+  benchopt satisfies the benchmark's ``min_benchopt_version``, exiting with a
+  non-zero status if it does not. By `Thomas Moreau`_ (:gh:`1009`)
```
```suggestion
  benchopt satisfies the benchmark's ``min_benchopt_version``.
  By `Thomas Moreau`_ (:gh:`1009`)
```

benchopt/cli/helpers.py
```diff
@@ -365,12 +378,35 @@ def _print_result_file_summary(result_file):
               is_flag=True,
               help="If used, list solver/dataset "
               "parameters, dependencies and availability.")
+@click.option('--version', 'check_version', is_flag=True,
+              help="Check that the installed benchopt version satisfies the "
+              "benchmark's `min_benchopt_version`. Print the versions and "
+              "exit with a non-zero status if it does not, instead of "
+              "listing solvers/datasets.")
```
tomMoral:
```suggestion
              help="Check that the installed benchopt version satisfies the "
              "benchmark's `min_benchopt_version`.")
```

